PR #3, opened 2015-10-23
Stop propagating php* reqs, add random_compat to php70

nicolas-grekas:


stof:
:+1: for that. Someone wanting to use a PHP 5.6 API should not be forced to install the 7.0 polyfill

xabbuh:
:+1:

PR #4, opened 2015-10-25
Mark classes as final

keradus:
Hi!
First of all, it is going to be a great lib!

Then, about PR - shouldn't internal classes be final where it is possible?

### DIFF
--- a/src/Iconv/Iconv.php
+++ b/src/Iconv/Iconv.php
@@ -34,7 +34,7 @@ namespace Symfony\Polyfill\Iconv;
  *
  * @internal
  */
-class Iconv
+final class Iconv
 {
     const ERROR_ILLEGAL_CHARACTER = 'iconv(): Detected an illegal character in input string';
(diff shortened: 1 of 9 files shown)


tests/Intl/Grapheme/GraphemeTest.php
```diff
@@ -18,7 +18,7 @@
  *
  * @covers Symfony\Polyfill\Intl\Grapheme\Grapheme::<!public>
  */
-class GraphemeTest extends \PHPUnit_Framework_TestCase
+final class GraphemeTest extends \PHPUnit_Framework_TestCase
```
stof: marking tests as final is not really useful

nicolas-grekas: can you remove the final keywork on tests @keradus please?

keradus: OK, I wasn't sure so I mark them in second commit on purpose ;)

done

nicolas-grekas:
Thank you @keradus.

PR #9, opened 2015-10-30
Always use mbstring if it is loaded, since other extensions can potentially overload core string fucntions

ircmaxell:
There is little to no cost, since you're already proxying these functions. So why couple to only one potential overload rather than support a more general case.


nicolas-grekas:
Thank you @ircmaxell.

PR #14, opened 2015-12-04
ldap_escape does not encode leading/trailing spaces.

ChadSikorra:
The `ldap_escape` function in this library is actually more RFC compliant at the moment than the official function. This removes the encoding for a leading/trailing space for escaping a DN. Also, a carriage return should be encoded when escaping a DN but the official function doesn't do that either (neither does the one here, but I put it in the test anyhow). I guess a bug report should probably be opened for the PHP function on their bug tracker.

The escape function in the LDAP component should probably be updated to account for the additional encoding mentioned above, but it doesn't appear to have any tests at all at the moment?

### DIFF
--- a/tests/Php56/Php56Test.php
+++ b/tests/Php56/Php56Test.php
@@ -80,7 +80,7 @@ class Php56Test extends \PHPUnit_Framework_TestCase
     public function provideLdapEscapeValues()
     {
         return array(
-            array('foo=bar(baz)*', null, p::LDAP_ESCAPE_DN, 'foo\3dbar(baz)*'),
+            array(" foo=bar\r(baz)* ", null, p::LDAP_ESCAPE_DN, ' foo\3dbar'."\r".'(baz)* '),
             array('foo=bar(baz)*', null, null, '\66\6f\6f\3d\62\61\72\28\62\61\7a\29\2a'),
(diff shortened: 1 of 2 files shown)


tests/Php56/Php56Test.php
```diff
@@ -80,7 +80,7 @@ public function testLdapEscape($subject, $ignore, $flags, $expected)
     public function provideLdapEscapeValues()
     {
         return array(
-            array('foo=bar(baz)*', null, p::LDAP_ESCAPE_DN, 'foo\3dbar(baz)*'),
```
csarrazi: Why do you remove this test? It is in no way related to this PR.

ChadSikorra: I added to it to include the leading/trailing spaces and a carriage return, as that was the test that seemed to be dealing strictly with escaping a DN. However, if you want, I can leave that test alone and include a completely separate line for testing a DN with leading/trailing spaces and a carriage return.

csarrazi: Actually, the escape tests were backported from the tests from the official PHP interpreter.
See https://github.com/php/php-src/tree/master/ext/ldap/tests

For maintenance's sake, you should create a new test case line.

Cheers! :)

ChadSikorra:
The test has been separated along with a little explanation. Let me know if there's anything else to do.

csarrazi:
LGTM :+1:

nicolas-grekas:
Thank you @ChadSikorra.

PR #15, opened 2015-12-06
add missing method

remicollet:
This allow to run unit test with PHPUnit 5.1

### DIFF
--- a/src/Util/TestListener.php
+++ b/src/Util/TestListener.php
@@ -148,4 +148,8 @@ EOPHP
     public function endTest(\PHPUnit_Framework_Test $test, $time)
     {
     }
+
+    public function addWarning(\PHPUnit_Framework_Test $test, \PHPUnit_Framework_Warning $e, $time)
+    {
+    }
 }


xabbuh:
:+1: (too bad the interface changed in a minor version though)

stof:
Yeah, too bad that PHPUnit does not follow semver...

keradus:
maybe raise a bug for PHPUnit?

keradus:
ref https://github.com/sebastianbergmann/phpunit/pull/1840#issuecomment-162552054

keradus:
https://github.com/sebastianbergmann/phpunit/commit/073c4de6d013353df368ccb25f9de37f13f61d2d

xabbuh:
So technically we do not need this change anymore, but maybe we should extend the `BaseTestListener` class so we (hopefully) do not have to change the code when PHPUnit 6 is released.

nicolas-grekas:
we can't extend the BaseTestListener: we already extend TestSuite

nicolas-grekas:
I'm :+1: for merging, planning for 6.0 compat

xabbuh:
Oh indeed, then let's merge it as is. :+1:

remicollet:
I confirm, this PR is no more required with PHPUnit 5.1.2.

nicolas-grekas:
Closing until 6.0 is released, thank you @remicollet

PR #17, opened 2015-12-07
Fix #16 - gzopen() / gzopen64() - 32 bit builds of Ubuntu 14.04

### LINKED ISSUE #16: gzopen()  / gzopen64()
On 32bit builds of Ubuntu 14.04LTS (PHP 5.5.9), three functions are created with the wrong names.
[See one of many bug reports.](https://bugs.launchpad.net/ubuntu/+source/php5/+bug/1315888)
There seems to be no plans to fix this in Ubuntu, so users of this LTS release cannot use the GZIP library.

Is there a place in this package for the necessary polyfills?

```php
if (!function_exists('gzopen') && function_exists('gzopen64')) {
	function gzopen($filename , $mode = 'r', $use_include_path = 0 ) {
		return gzopen64($filename, $mode, $use_include_path);
	}
}
if (!function_exists('gzseek') && function_exists('gzseek64')) {
	function gzseek($zp , $offset, $whence = SEEK_SET) {
		return gzseek64($zp, $offset, $whence);
	}
}
if (!function_exists('gztell') && function_exists('gztell64')) {
	function gztell($zp) {
		return gztell64($zp);
	}
}
```

fisharebest:
Fix for #16

### DIFF
--- a/src/Php56/bootstrap.php
+++ b/src/Php56/bootstrap.php
@@ -21,4 +21,18 @@ if (PHP_VERSION_ID < 50600) {
 
         function ldap_escape($subject, $ignore = '', $flags = 0) { return p\Php56::ldap_escape($subject, $ignore, $flags); }
     }
+
+    if (50509 === PHP_VERSION_ID && 4 === PHP_INT_SIZE) {
+        // Missing functions in PHP 5.5.9 - affects 32 bit builds of Ubuntu 14.04LTS
+        // See https://bugs.launchpad.net/ubuntu/+source/php5/+bug/1315888
+        if (!function_exists('gzopen') && function_exists('gzopen64')) {
+            function gzopen($filename, $mode, $use_include_path = 0) { return gzopen64($filename, $mode, $use_include_path); }
+        }
+        if (!function_exists('gzseek') && function_exists('gzseek64')) {
+            function gzseek($zp, $offset, $whence = SEEK_SET) { return gzseek64($zp, $offset, $whence); }
+        }
+        if (!function_exists('gztell') && function_exists('gztell64')) {
+            function gztell($zp) { return gztell64($zp); }
+        }
+    }
 }


src/Php56/bootstrap.php
```diff
@@ -21,4 +21,16 @@ function hash_equals($knownString, $userInput) { return p\Php56::hash_equals($kn
...
+    // Missing functions in PHP 5.5.9 - affects 32 bit builds of Ubuntu 14.04LTS
+    // See https://bugs.launchpad.net/ubuntu/+source/php5/+bug/1315888
+    if (!function_exists('gzopen') && function_exists('gzopen64')) {
+        function gzopen($filename , $mode = 'r', $use_include_path = 0 ) { return gzopen64($filename, $mode, $use_include_path); }
+    }
+    if (!function_exists('gzseek') && function_exists('gzseek64')) {
+        function gzseek($zp , $offset, $whence = SEEK_SET) { return gzseek64($zp, $offset, $whence); }
+    }
+    if (!function_exists('gztell') && function_exists('gztell64')) {
+        function gztell($zp) { return gztell64($zp); }
```
keradus: `return p\Phy56::gztell64($zp);` ?

keradus:
adding some tests would be awesome!

fisharebest:
The travis failure is not in my code ;-)

It looks like php-7 is using phpunit-5, but perhaps the tests were written for phpunit-4

keradus:
ref https://github.com/symfony/polyfill/pull/15 ( and https://github.com/symfony/polyfill/pull/15#issuecomment-162552595 )

src/Php56/bootstrap.php
nicolas-grekas: i'm ok with not duplicating the function as static method in the Php56 class (reduces boiler plates)

src/Php56/bootstrap.php
```diff
@@ -21,4 +21,16 @@ function hash_equals($knownString, $userInput) { return p\Php56::hash_equals($kn
 
         function ldap_escape($subject, $ignore = '', $flags = 0) { return p\Php56::ldap_escape($subject, $ignore, $flags); }
     }
+
+    // Missing functions in PHP 5.5.9 - affects 32 bit builds of Ubuntu 14.04LTS
+    // See https://bugs.launchpad.net/ubuntu/+source/php5/+bug/1315888
+    if (!function_exists('gzopen') && function_exists('gzopen64')) {
+        function gzopen($filename , $mode = 'r', $use_include_path = 0 ) { return gzopen64($filename, $mode, $use_include_path); }
```
nicolas-grekas: $mode doesn't default to 'r' per the doc (http://php.net/gzopen)
trailing space before `)`

src/Php56/bootstrap.php
stof: why that ?

stof:
The Travis failure is because of a BC break in PHPUnit 5.1

src/Php56/bootstrap.php
stof: and shouldn't ``$use_include_path`` be a boolean instead ?

stof: ah no, they use a crappy int API instead of a proper boolean

src/Php56/bootstrap.php
keradus: more consistency and easier to test then moved into `Php56`

nicolas-grekas: I wonder @keradus how would you test that? We'd need the broken ubuntu isn't it?

fisharebest:
> adding some tests would be awesome!

Since we are working around a platform-specific problem, I'm not sure how we create a meaningful test?

To trigger the code, we'd need to create a situation where the function `gzopen64()` does not exist.  Since we cannot "unload" PHP libraries at runtime, I don't see how to do this.

Besides, none of the other "thin wrappers" have unit tests.

FYI, I've tested the code on a VM...

fisharebest:
> $mode doesn't default to 'r'

Oops.  My bad.  Fixed

src/Php56/bootstrap.php
```diff
@@ -21,4 +21,16 @@ function hash_equals($knownString, $userInput) { return p\Php56::hash_equals($kn
...
         function ldap_escape($subject, $ignore = '', $flags = 0) { return p\Php56::ldap_escape($subject, $ignore, $flags); }
     }
+
+    // Missing functions in PHP 5.5.9 - affects 32 bit builds of Ubuntu 14.04LTS
+    // See https://bugs.launchpad.net/ubuntu/+source/php5/+bug/1315888
+    if (!function_exists('gzopen') && function_exists('gzopen64')) {
+        function gzopen($filename , $mode, $use_include_path = 0) { return gzopen64($filename, $mode, $use_include_path); }
+    }
+    if (!function_exists('gzseek') && function_exists('gzseek64')) {
+        function gzseek($zp , $offset, $whence = SEEK_SET) { return gzseek64($zp, $offset, $whence); }
```
nicolas-grekas: extra space before first comma (same above)

src/Php56/bootstrap.php
```diff
@@ -21,4 +21,16 @@ function hash_equals($knownString, $userInput) { return p\Php56::hash_equals($kn
 
         function ldap_escape($subject, $ignore = '', $flags = 0) { return p\Php56::ldap_escape($subject, $ignore, $flags); }
     }
+
+    // Missing functions in PHP 5.5.9 - affects 32 bit builds of Ubuntu 14.04LTS
+    // See https://bugs.launchpad.net/ubuntu/+source/php5/+bug/1315888
+    if (!function_exists('gzopen') && function_exists('gzopen64')) {
+        function gzopen($filename , $mode, $use_include_path = 0) { return gzopen64($filename, $mode, $use_include_path); }
```
nicolas-grekas: extra space before first comma

fisharebest: > extra space before first comma

Removed.

src/Php56/bootstrap.php
```diff
@@ -21,4 +21,16 @@ function hash_equals($knownString, $userInput) { return p\Php56::hash_equals($kn
 
         function ldap_escape($subject, $ignore = '', $flags = 0) { return p\Php56::ldap_escape($subject, $ignore, $flags); }
     }
+
+    // Missing functions in PHP 5.5.9 - affects 32 bit builds of Ubuntu 14.04LTS
```
nicolas-grekas: I propose to enclose these checks in a `if (8 > PHP_INT_SIZE) {` (and removing it from this `PHP_VERSION_ID < 50600` block) to get a few more CPU cycles on 64b setups (which are the most common)

fisharebest: Just powered up my 14.04 / 32bit virtual machine...

`if (PHP_VERSION_ID === 50509 && PHP_INT_SIZE === 4) { ... }` works nicely for me.

nicolas-grekas:
:+1:

src/Php56/bootstrap.php
stof: Could it affect a few other versions of PHP shipped in other distributions than Ubuntu 14.04 ?

fisharebest: @stof - not to my knowledge.  All google searches and bug reports point to just the 32bit build of 14.04

nicolas-grekas:
Thank you @fisharebest.

PR #18, opened 2015-12-11
Add polyfill for PHP5.4 http_response_code()

fisharebest:


fisharebest:
The test failures (which I also see in the master branch) can be eliminated by using conditional code in `src/Php56/bootstrap.php`.  e.g.

````
define('LDAP_ESACPE_FILTER', 1);
define('LDAP_ESACPE_DN', 2);
````
becomes
````
defined('LDAP_ESACPE_FILTER') || define('LDAP_ESACPE_FILTER', 1);
defined('LDAP_ESACPE_DN') || define('LDAP_ESACPE_DN', 2);
````

But presumably the correct answer is to fix the process isolation in PHPUnit?

FWIW, the new tests (`phpunit tests/Php54`) work OK.

File: src/Php54/Php54.php [hunk 1]
```diff
@@ -39,4 +39,63 @@ public static function hex2bin($data)
...
+          415 => 'Unsupported Media Type',
+          500 => 'Internal Server Error',
+          501 => 'Not Implemented',
+          502 => 'Bad Gateway',
+          503 => 'Service Unavailable',
+          504 => 'Gateway Time-out',
+          505 => 'HTTP Version not supported',
+        );
+
+        $previous_code = $current_code;
```
stof: please use camelCased variable names

fisharebest: @stof - should function names also be camelCase?  I notice that all the existing ones are snake_case.

src/Php54/Php54.php
```diff
@@ -39,4 +39,63 @@ public static function hex2bin($data)
 
         return $data;
     }
+
+    public static function http_response_code($response_code)
+    {
+        static $current_code = 200;
+
+        $messages = array(
```
keradus: static?

src/Php54/Php54.php
```diff
@@ -39,4 +39,63 @@ public static function hex2bin($data)
...
+          203 => 'Non-Authoritative Information',
+          204 => 'No Content',
+          205 => 'Reset Content',
+          206 => 'Partial Content',
+          300 => 'Multiple Choices',
+          301 => 'Moved Permanently',
+          302 => 'Moved Temporarily',
+          303 => 'See Other',
+          304 => 'Not Modified',
+          305 => 'Use Proxy',
```
keradus: ` 307 => "Temporary Redirect",` ?

src/Php54/Php54.php
```diff
@@ -39,4 +39,63 @@ public static function hex2bin($data)
...
+          406 => 'Not Acceptable',
+          407 => 'Proxy Authentication Required',
+          408 => 'Request Time-out',
+          409 => 'Conflict',
+          410 => 'Gone',
+          411 => 'Length Required',
+          412 => 'Precondition Failed',
+          413 => 'Request Entity Too Large',
+          414 => 'Request-URI Too Large',
+          415 => 'Unsupported Media Type',
```
keradus:
```
            416 => "Requested Range Not Satisfiable",
            417 => "Expectation Failed",
```
?

src/Php54/Php54.php
fisharebest: The function needs to return its previous value, so we need a static variable for this.
Are you saying that you would prefer a class-level static, rather than a function-level static?

src/Php54/Php54.php
stof: function names should match the name of the function they poyfill

src/Php54/Php54.php
keradus: I put it under `$messages`, not `$current_code`. `$messages` could be static

src/Php54/Php54.php
fisharebest: @keradus - 307/308 added.  FWIW, most servers ignore this value and use their own.  I've only been able to see these values using `php -S`

src/Php54/Php54.php
fisharebest: Added

nicolas-grekas:
IMHO, this function can't be polyfilled correctly, and thus shouldn't be polyfilled at all: the case with no argument for getting the current status is just impossible to implement in plain PHP: if you set the status code using `header()`, no polyfill can return the correct result for `http_response_code()`...
Sadly, I'm :-1: here...

fisharebest:
> IMHO, this function can't be polyfilled correctly, and thus shouldn't be polyfilled at all

OK - we have different objectives.

I just want my applications to run on old servers.

I can create another package for this function, and similar.

keradus:
(outside the discussion, @fisharebest - ref [this](https://github.com/keradus/IdeaWebServer_like_Apache/blob/master/IdeaWebServer_like_Apache.php))

src/Php54/Php54.php
```diff
@@ -39,4 +39,67 @@ public static function hex2bin($data)
 
         return $data;
     }
+
+    public static function http_response_code($responseCode)
+    {
+        static $currentCode = 200;
+
+        $messages = array(
```
nicolas-grekas: This array us not necessary: php source already embed the list for you. All you have to do is send a `header('Status: XXX')` and let it do the job

tests/Php54/Php54Test.php
```diff
@@ -0,0 +1,58 @@
...
+
+use Symfony\Polyfill\Php54\Php54 as p;
+
+class Php54Test extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @runInSeparateProcess
+     */
+    public function testHttpResponseCodeDefaultValueIs200() {
+        $code = p::http_response_code(null);
```
nicolas-grekas: tests is this polyfill framework should be against the native function name (no `p::` prefix):
there is a phpunit test listener that will do the job of duplicating the test case, one testing the polyfill, the other testing the native implem if it's available.

stof:
@fisharebest the issue with your proposal is that it polyfills only half of the behavior, and breaks the other one. So this will make *some* projects run, and break others (even if they might be able to support 5.3 by detecting the availibility of the feature)

tests/Php54/Php54Test.php
fisharebest: > should be against the native function 

That won't work.  This function behaves differently under CLI.  We'd want to compare the CGI/FPM versions, which we obviously cannot do using PHPUnit.

src/Php54/Php54.php
fisharebest: > php source already embeds the list for you

No it does.  Apache/Nginx ignores your message and inserts its own.  To see tha native PHP behaviour, runr PHP as a server (`php -S ... -t ...`) and run these scripts:

`header('HTTP/1.0 200 Whatever');`
`header('HTTP/1.0 200');`

In your brower, you'll see these exact headers.  PHP doesn't add "OK".

fisharebest:
> So this will make some projects run

That is the idea ;-)

PR #19, opened 2015-12-15
[Intl/Normalizer] Add Unicode conformance tests

nicolas-grekas:


tests/Intl/Normalizer/NormalizerTest.php
```diff
@@ -78,4 +78,66 @@ public function testNormalize()
...
+                $this->assertSame($c[3], pn::normalize($c[1], pn::NFKC));
+                $this->assertSame($c[3], pn::normalize($c[2], pn::NFKC));
+                $this->assertSame($c[3], pn::normalize($c[3], pn::NFKC));
+                $this->assertSame($c[3], pn::normalize($c[4], pn::NFKC));
+
+                $this->assertSame($c[4], pn::normalize($c[0], pn::NFKD));
+                $this->assertSame($c[4], pn::normalize($c[1], pn::NFKD));
+                $this->assertSame($c[4], pn::normalize($c[2], pn::NFKD));
+                $this->assertSame($c[4], pn::normalize($c[3], pn::NFKD));
+                $this->assertSame($c[4], pn::normalize($c[4], pn::NFKD));
```
stof: should we really call the stub directly here ? Shouldn't it call the polyfilled API instead, relying on the custom test listener ?

nicolas-grekas: I'd say that testing the native implantation again the unicode spec is out of scope. We have a standardize reference here so no need for the native php reference. At least that's the reasoning I've had for now.

nicolas-grekas: OK, done

nicolas-grekas:
So, now that unicode conformance tests run against the native implementations, we have to find a common subset that works for all the various ICU versions into the wild.

nicolas-grekas:
Green

tests/Intl/Normalizer/NormalizerTest.php
```diff
@@ -44,14 +44,14 @@ public function testIsNormalized()
         $c = 'déjà';
         $d = in::normalize($c, pn::NFD);
 
-        $this->assertTrue(pn::isNormalized(''));
-        $this->assertTrue(pn::isNormalized('abc'));
-        $this->assertTrue(pn::isNormalized($c));
-        $this->assertTrue(pn::isNormalized($c, pn::NFC));
+        $this->assertTrue(!!normalizer_is_normalized(''));
```
stof: why ``!!`` ? is it returning something else than a boolean ?

nicolas-grekas: Yes on hhvm, I reported it btw

stof: I suggest adding a comment referencing the HHVM issue, to remember why this casting is there.

And I think ``(bool)`` would be more readable for PHP devs (the ``!!`` is common in JS to save bytes, but you are the only PHP dev I know who try to make the PHP code look like minified code)

PR #22, opened 2015-12-29
[APCu] A new polyfill for the legacy APC users

nicolas-grekas:
Now that we know that `apc_*` functions are legacy and `apcu_*` is the only way in PHP7, I propose to move forward with this polyfill. Let's replace all usages of `apc_*` in Symfony and elsewhere by their `apcu_*` equivalent and add a deps on this new polyfill to cope with legacy `apc` users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,16 @@ matrix:
 
 before_install:
     - composer self-update;
-    - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then phpenv config-rm xdebug.ini; fi;
-    - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then echo "extension = ldap.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi;
-    - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then php -i; fi;
+    - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then INI_FILE=~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; else INI_FILE=/etc/hhvm/php.ini; fi;
+    - echo memory_limit = -1 >> $INI_FILE
+    - echo session.gc_probability = 0 >> $INI_FILE
+    - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then phpenv config-rm xdebug.ini; fi;
+    - if [[ $TRAVIS_PHP_VERSION = 5.* ]]; then (echo yes | pecl install -f apcu-4.0.10 && echo apc.enable_cli = 1 >> $INI_FILE) || echo "Let's continue without apcu extension"; fi;
+    - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then echo extension = ldap.so >> $INI_FILE; fi;
+    - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then php -i; fi;
 
 install:
-    - if [ "$TRAVIS_BRANCH" = "master" ]; then export COMPOSER_ROOT_VERSION=dev-master; else export COMPOSER_ROOT_VERSION="$TRAVIS_BRANCH".x-dev; fi;
+    - if [[ $TRAVIS_BRANCH = master ]]; then export COMPOSER_ROOT_VERSION=dev-master; else export COMPOSER_ROOT_VERSION=$TRAVIS_BRANCH.x-dev; fi;
     - composer --prefer-source install;
 
 script:
--- a/src/Apcu/LICENSE
+++ b/src/Apcu/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2016 Fabien Potencier
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is furnished
+to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
--- a/src/Apcu/composer.json
+++ b/src/Apcu/composer.json
@@ -1,0 +1,31 @@
+{
+    "name": "symfony/polyfill-apcu",
+    "type": "library",
+    "description": "Symfony polyfill backporting apcu_* functions to lower PHP versions",
+    "keywords": ["polyfill", "shim", "compatibility", "portable", "apcu"],
+    "homepage": "https://symfony.com",
+    "license": "MIT",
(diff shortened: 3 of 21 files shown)


nicolas-grekas:
ping @dunglas @wouterj @ogizanagi

src/Apcu/Apcu.php
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        $entry = apcu_fetch($key, $success);
+
+        if ($success) {
+            return $entry;
+        }
+
+        $h = fopen(__FILE__, 'r');
+        flock($h, LOCK_EX);
```
dunglas: Is it sufficient to avoid race conditions?
What if the file is already being executed by another thread before the call to `flock`? What's the behavior when the opcache caching is on?

Maybe can you just throw a `NotSupportedException` for this specific function?

nicolas-grekas: > Is it sufficient to avoid race conditions?
I think so
> What if the file is already being executed by another thread
locking happens at the kernel level and is share among threads/processes
> throw a NotSupportedException
as last resort  but for now I see no issue here

dunglas:
Btw, APC is not used directly in Symfony except in the `ApcClassLoader` (https://github.com/symfony/symfony/search?utf8=%E2%9C%93&q=apc_fetch&type=Code).

Most APC calls are done by the Doctrine Cache library.

src/Apcu/Apcu.php
dunglas: Nervermind it should be fine. At least it's ok for the case I described before.

src/Apcu/composer.json
```diff
@@ -0,0 +1,31 @@
...
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "require": {
+        "php": ">=5.3.3"
+    },
+    "autoload": {
+        "files": [ "bootstrap.php" ],
+        "classmap": [ "Resources/stubs" ]
```
stof: missing autoloading for the ``Symfony\Polyfill\Apcu\Apcu`` class

src/Apcu/Apcu.php
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @internal
+ */
+final class Apcu
+{
+    public static function apcu_entry($key, $generator, $ttl = 0)
+    {
+        if (!is_scalar($key)) {
+            throw new \InvalidArgumentException(sprintf('apcu_entry() expects parameter 1 to be string, %s given', __FUNCTION__, gettype($value)));
```
stof: there is no placeholder for ``__FUNCTION__``

nicolas-grekas: fixed

src/Apcu/composer.json
stof: this needs to be fixed

nicolas-grekas: good catch, fixed

nicolas-grekas:
Ping @symfony/deciders

dunglas:
:+1:

src/Apcu/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2014-2015 Fabien Potencier
```
Tobion: 2016 only

.travis.yml
```diff
@@ -14,12 +14,16 @@ matrix:
...
 before_install:
     - composer self-update;
-    - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then phpenv config-rm xdebug.ini; fi;
-    - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then echo "extension = ldap.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi;
-    - if [[ "$TRAVIS_PHP_VERSION" != "hhvm" ]]; then php -i; fi;
+    - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then INI_FILE=~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; else INI_FILE=/etc/hhvm/php.ini; fi;
+    - echo memory_limit = -1 >> $INI_FILE
+    - echo session.gc_probability = 0 >> $INI_FILE
+    - if [[ $TRAVIS_PHP_VERSION != hhvm ]]; then phpenv config-rm xdebug.ini; fi;
+    - if [[ $TRAVIS_PHP_VERSION = 5.* ]]; then (echo yes | pecl install -f apcu-4.0.10 && echo apc.enable_cli = 1 >> $INI_FILE) || echo "Let's continue without apcu extension"; fi;
```
Tobion: why `apc.enable_cli` here and in `phpunit.xml.dist`

nicolas-grekas: I removed it from phpunit.xml.dist since it can't be "ini_setted"

src/Apcu/LICENSE
nicolas-grekas: fixed

tests/Apcu/ApcuTest.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->assertSame(array(__CLASS__), array_keys($entries));
+        $this->assertSame(__CLASS__, $entries[__CLASS__]['key']);
+        $this->assertSame(456, $entries[__CLASS__]['value']);
+    }
+
+    public function testApcuEntry()
+    {
+        apcu_delete(__CLASS__);
+
+        $gen = function ($key) use (&$ok) {
```
Tobion: shoulnd't `$ok` be intitialized first?

nicolas-grekas: It could but that would be exactly the same, references are automatically assigned to null

src/Apcu/Apcu.php
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @internal
+ */
+final class Apcu
+{
+    public static function apcu_entry($key, $generator, $ttl = 0)
+    {
+        if (!is_scalar($key)) {
+            throw new \InvalidArgumentException(sprintf('apcu_entry() expects parameter 1 to be string, %s given', gettype($value)));
```
Tobion: this is not the same behavior as the real apcu_entry` method. it does not throw an exception and it also does not seem to generate a warning either from what I tested. It just return null.

Tobion:
The doc about `apcu_entry http://php.net/manual/en/function.apcu-entry.php says

> In addition, it prohibits the concurrent execution of any other APCu functions, since they will acquire the same lock.

This does not happen in the current polyfill implementation. This means you can still have race conditions if another process modifies the data while you are working with it inside the generator callable.

tests/Apcu/ApcuTest.php
```diff
@@ -0,0 +1,63 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Tests\Apcu;
+
+/**
+ * @requires extension apc
```
Tobion: shouldn't these tests run against both apc and apcu to ensure they work the same. now it's only executed against apc.

src/Apcu/Apcu.php
nicolas-grekas: Thanks for testing this, I will add a test then

tests/Apcu/ApcuTest.php
nicolas-grekas: Apcu provides apc, it returns true to extension_loaded('apc')

nicolas-grekas:
@tobion then we should acquire the lock in every apc function call, I'll add it tomorrow.

nicolas-grekas:
@Tobion I removed the polyfilling of `apcu_entry`: too many edge cases with locking, and not required for moving forward.

fabpot:
:+1:

tests/Apcu/ApcuTest.php
Tobion: Hm not for me using APCu 5.1.2 on Windows

```
$ php --version
PHP 7.0.1 (cli) (built: Dec 16 2015 13:32:58) ( NTS )
$ php -d "apc.enable_cli=1" -r "var_dump(extension_loaded('apc'));"
bool(false)
$ php -d "apc.enable_cli=1" -r "var_dump(extension_loaded('apcu'));"
bool(true)
```

nicolas-grekas: so, it looks like you don't have the apcu bc layer enabled, but that doesn't change my point: this is tested properly on travis, against apcu + apc-br-from-apcu, which means there is nothing to fix, the case is covered

nicolas-grekas: OH, and btw, this polyfill makes no sense on PHP 7 (where apc has never been released and apcu is the only avail thing)

PR #24, opened 2015-12-30
Add .gitattributes

nicolas-grekas:
After reading https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production

PR #26, opened 2016-01-03
Exclude dev/test files from releases

fisharebest:
Users of the package (as opposed to developers) do not need the test/dev files.

Exclude these from releases.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+.gitattributes   export-ignore
+.gitignore       export-ignore
+.travis.yml      export-ignore
+appveyor.yml     export-ignore
+phpunit.xml.dist export-ignore
+tests            export-ignore


xabbuh:
see #24 and symfony/symfony#6605

fisharebest:
> see #24

I did not see this.

I write software that is deployed on thousands of servers.
By people with slow internet connections.
Keeping distributions small is important.

By definition, this package will be used in similar circumstances - software that is widely distributed.  Anyone writing bespoke software for their own server will be able to upgrade PHP to a suitable version.

@fabpot says
 > the benefits are really small and the drawbacks quite huge

IMHO wrong.  The benefits are large (for thousands of people).  The drawbacks are tiny (the small number of developers can use `composer --prefer-source` or `git clone`).

This package says (in README.md)

> This package is designed for low overhead

I guess this depends on how you define "overhead" ;-)

nicolas-grekas:
You should require individual polyfills, not the bigger one

nicolas-grekas:
Individual polyfills are naturally stripped down from tests. We should mention that in the README file instead I believe.

keradus:
:-1: for PR
:+1: for @nicolas-grekas proposal

nicolas-grekas:
Thanks for proposing

PR #27, opened 2016-01-06
Add the Throwable interface

derrabus:
This PR adds the `Throwable` interface to the PHP7 polyfill. This way, the interface can be used in type hints and `catch` statements.

Limitations in PHP 5.x:
* Regular exceptions won't implement the `Throwable` interface, only children of the polyfilled `Error` class.
* While a `catch (Throwable $t)` statement won't produce an error, it will not catch anything either.

The interface becomes interesting for writing `try`/`catch` blocks that should catch everything.

```php
try {
    do_something();
} catch (Throwable $t) {
    // Global exception/error handling for PHP >= 7
} catch (Exception $e) {
    // Legacy exception/error handling block for PHP 5.x
    // Can be deleted when PHP 5.x support is dropped.
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,9 @@ Compatibility notes
 
 To write portable code between PHP5 and PHP7, some care must be taken:
 - `\*Error` exceptions must by caught before `\Exception`;
+- The `\Throwable` interface is not implemented by `\Exception` in PHP5.
+- Both `\Error` and `\Exception` will fall through a `try {} catch (\Throwable $e) {}`
+  block in PHP5.
 - after calling `error_clear_last()`, the result of `$e = error_get_last()` must be
   verified using `isset($e['message'][0])` instead of `null === $e`.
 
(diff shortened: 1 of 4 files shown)


nicolas-grekas:
We already rejected this polyfill, and even removed it from paragonie/random_compat (https://github.com/paragonie/random_compat/pull/66) with this argument:
> By creating a `Throwable` interface that is not implemented by `Exception`, `$exception instanceof Throwable` returns `false` in PHP5 and `true` on PHP7. This is a serious issue.

stof:
@derrabus you can already write such try/catch without a polyfill (you can catch any class name in PHP, even if it does not exist. Same than ``instanceof`` and typehints).

Stubbing the interface makes things worse by breaking feature detection and still not polyfilling the behavior of the interface

nicolas-grekas:
Thanks for proposing

derrabus:
Sorry for not responding. You are right, the only advantage that this polyfill would give us is is working `Throwable` typehints on `Error` instances. This is probably not enough to legitimate the change.

PR #28, opened 2016-01-13
[Php70] Workaround https://bugs.php.net/63206

nicolas-grekas:
This ensures we never create the conditions for https://bugs.php.net/63206
I.e. never use set_error_handler's second argument.

### DIFF
--- a/src/Php70/Php70.php
+++ b/src/Php70/Php70.php
@@ -51,7 +51,11 @@ final class Php70
 
     public static function error_clear_last()
     {
-        set_error_handler('var_dump', 0);
+        static $handler;
+        if (!$handler) {
+             $handler = function() { return false; };
+        }
+        set_error_handler($handler);
         @trigger_error('');
         restore_error_handler();
     }


src/Php70/Php70.php
```diff
@@ -51,7 +51,11 @@ public static function preg_replace_callback_array(array $patterns, $subject, $l
 
     public static function error_clear_last()
     {
-        set_error_handler('var_dump', 0);
+        static $handler;
+        if (!$handler) {
+             $handler = function() {return false;};
```
fabpot: CS wrong (whitespace around return false)

nicolas-grekas: better now?

fabpot: looks great :)

stof:
Do we also have some places in Symfony affected by this bug ?

nicolas-grekas:
@stof as you may have already spotted, this has been fixed in Symfony now

PR #30, opened 2016-01-20
Update the branch alias to 1.1 as there is a new component

stof:


nicolas-grekas:
Is it a new feature? There is no newly created interface here, but only a polyfill for an existing one.
This looks like a case where semver doesn't apply very well to me...

stof:
Well, there is a new polyfilled API. This is a new feature in the package (the public API of the package is the list of polyfilled APIs).
Adding more polyfills is not a bug fix.

keradus:
We have new features (/polyfill) at main symfony/polyfill repo, raise MAJOR semver version, great. Why do we raise version of each separated polyfill, even if it wasn't touched?
I know that this is that way at whole symfony repos, but could you ref some documentation about bumping version of modules?

stof:
@keradus why major version for new features ? New features are semver minor.

And this is bumped for all sub-components due to the way our development process work. If we want to decouple the versionning of each polyfill component, we cannot have them in the same repo in the first place (as they would share the same tags). They would have to be developed separately (which is what the ZF team is doing now btw)

nicolas-grekas:
And which something which would be way more difficult to scale. I heard someone asking how the Symfony project managed to keep adding new component without imploding. This process is really a key element to the answer. Makes everyone's life easier, both for maintainers, releasers and even users. One number to remember for a bunch of components managed in bulk. Thank you @fabpot!

keradus:
@stof yeah minor, sorry ;) 

I know what ZF is doing and indeed it could be painful, for end-users it is for sure easier to contribute/use it with Symfony way, but I can't find any refs with background of that decision.

Thanks ;)

fabpot:
Thank you @stof.

PR #31, opened 2016-01-25
Fix class_uses polyfill

WouterJ:
`class_uses` accepts both the class name and object instance as first argument, `class_exists` only accepts a class name. If an object instance is giving, the class is already autoloaded, so the other logic can be skipped.

### DIFF
--- a/src/Php54/bootstrap.php
+++ b/src/Php54/bootstrap.php
@@ -16,7 +16,14 @@ if (PHP_VERSION_ID < 50400) {
         function trait_exists($class, $autoload = true) { return $autoload && class_exists($class, $autoload) && false; }
     }
     if (!function_exists('class_uses')) {
-        function class_uses($class, $autoload = true) { return $autoload && class_exists($class, $autoload) && false; }
+        function class_uses($class, $autoload = true)
+        {
+            if (is_object($class) || class_exists($class, $autoload) || interface_exists($class, false)) {
+                return array();
+            }
+
+            return false;
+        }
     }
     if (!function_exists('hex2bin')) {
         function hex2bin($data) { return p\Php54::hex2bin($data); }
--- a/tests/Php54/Php54Test.php
+++ b/tests/Php54/Php54Test.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Tests\Php54;
+
+class Php54Test extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @dataProvider provideClassUsesValid
+     */
+    public function testClassUsesValid($classOrObject)
+    {
+        $this->assertSame(array(), class_uses($classOrObject));
+    }
+
+    public function provideClassUsesValid()
+    {
+        return array(
+            array('stdClass'),
+            array(new \stdClass()),
+            array('Iterator'),
+        );
+    }
+
+    public function testClassUsesInvalid()
+    {
+        $this->assertFalse(@class_uses('NotDefined'));
+    }
+}


nicolas-grekas:
Could you add a test please ?

WouterJ:
@nicolas-grekas done

WouterJ:
I've completely changed the function now btw, as it had to return an array and not a boolean as it did before.

nicolas-grekas:
Great! Somme comments:
- class_uses with a non-existent class name  returns false,
-  we miss a test with a class instance,
- I'm not sure the spl_autoload_register provides more behavior coverage, does it? If yes, then I'd suggest unregistering the autoloader to leave the global state clean after the test has run.

WouterJ:
I've fixed the first 2 comments.

The test with `spl_autoload_register` was there to test the second argument of class_uses: `$autoload`. However, I couldn't get things to work probably (unregistering didn't work for me for some reason), so I've removed the tests for now.

src/Php54/bootstrap.php
```diff
@@ -16,7 +16,14 @@
         function trait_exists($class, $autoload = true) { return $autoload && class_exists($class, $autoload) && false; }
     }
     if (!function_exists('class_uses')) {
-        function class_uses($class, $autoload = true) { return $autoload && class_exists($class, $autoload) && false; }
+        function class_uses($class, $autoload = true)
+        {
+            if (!is_object($class) && $autoload && !class_exists($class, true)) {
```
nicolas-grekas: when autoload === false, we should still return false when the class does not exist. But when $class is an interface, we shoudld return array()... Could be done with:
```php
if (is_object($class) || class_exists($class, $autoload) || interface_exists($class, false)) {
    return array();
}

return false;
```

tests/Php54/Php54Test.php
```diff
@@ -0,0 +1,32 @@
...
+class Php54Test extends \PHPUnit_Framework_TestCase
+{
+    public function testClassUsesAutoloadedClass()
+    {
+        $this->assertEquals(array(), class_uses(__CLASS__.'_Fixture'));
+    }
+
+    public function testClassUsesClassInstance()
+    {
+        $this->assertEquals(array(), class_uses(new Php54Test_Fixture()));
```
nicolas-grekas: could be just `new stdClass()`

tests/Php54/Php54Test.php
```diff
@@ -0,0 +1,32 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Tests\Php54;
+
+class Php54Test extends \PHPUnit_Framework_TestCase
+{
+    public function testClassUsesAutoloadedClass()
+    {
+        $this->assertEquals(array(), class_uses(__CLASS__.'_Fixture'));
```
nicolas-grekas: assertSame

WouterJ:
Updated again

src/Php54/bootstrap.php
```diff
@@ -16,7 +16,14 @@
         function trait_exists($class, $autoload = true) { return $autoload && class_exists($class, $autoload) && false; }
     }
     if (!function_exists('class_uses')) {
-        function class_uses($class, $autoload = true) { return $autoload && class_exists($class, $autoload) && false; }
+        function class_uses($class, $autoload = true)
+        {
+            if (is_object($class) || class_exists($class, $autoload) || interface_exists($class, $autoload)) {
```
nicolas-grekas: interface_exists($class, **false**) because the autoloader has already been triggered, if needed, by class_exists

tests/Php54/Php54Test.php
```diff
@@ -0,0 +1,37 @@
...
+        return array(
+            array('stdClass'),
+            array(new \stdClass()),
+            array('Iterator'),
+        );
+    }
+
+    public function testClassUsesInvalid()
+    {
+        $this->assertFalse(@class_uses('NotDefined'));
```
stof: do we need the silencing here ?

nicolas-grekas: yes: php triggers a warning

stof: then shouldn't we try to polyfill it properly (and test that) ?

WouterJ: How can you polyfill a warning trigger? And I was unsure how to test triggered errors in PHPunit, so I skipped that test.

src/Php54/bootstrap.php
WouterJ: fixed

nicolas-grekas:
:+1: (the missing warning is a bonus to me, nobody should rely on it)

tests/Php54/Php54Test.php
nicolas-grekas: @WouterJ by trigger an empty error, then using a custom error handler that aways returns false, the call `@class_uses`, then check the result of error_get_last.

xabbuh:
:+1:

WouterJ:
Can we please merge this?

fabpot:
Thank you @WouterJ.

PR #34, opened 2016-02-14
tfix

DavidPrevot:


src/Intl/Icu/README.md
```diff
@@ -2,7 +2,7 @@ Symfony Polyfill / Intl: ICU
 ============================
 
 This component maps a collection of functions/classes using the
-[`symfony/intl`](https://github.com/symfony/intl) package in the abscense of the
+[`symfony/intl`](https://github.com/symfony/intl) package in the absence of the
```
fabpot: I think `absence` is not the right word here. I would rephrase the sentence to something along the lines of:

`This component maps a collection of functions/classes using the intl package when the intl extension is not installed, including:`

ping @weaverryan

weaverryan: I agree! And unless I'm missing some important meaning, I would also change `maps` => `provides`

nicolas-grekas:
Thank you @DavidPrevot.

PR #35, opened 2016-02-15
Fix composer.json synstax

DavidPrevot:


fabpot:
Thank you @DavidPrevot.

PR #38, opened 2016-03-02
bail if class exist

collizo4sky:
WordPress plugin svn repo has a pre-commit hook that run a check on php codes on various php version (think CI builds)

commit was failing due to existence of `CallbackFilterIterator` in php 5.4 already. add a class_exists check solved it.

### DIFF
--- a/src/Php54/Resources/stubs/CallbackFilterIterator.php
+++ b/src/Php54/Resources/stubs/CallbackFilterIterator.php
@@ -9,20 +9,22 @@
  * file that was distributed with this source code.
  */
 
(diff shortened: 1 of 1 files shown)


stof:
Well, this does not make sense. the autoloader won't load the file if the class already exists, as the autoloader won't be triggered.
If the Wordpress CI system requires to be able to load any file itself without carign about how the file is actually loaded for real, there CI is totally broken.

nicolas-grekas:
I agree with @stof, there is nothing wrong here...

collizo4sky:
I agree too. This isn't a symfony bug.

PR #40, opened 2016-03-03
[Apcu] Load APCUIterator only when APCIterator exists

### LINKED ISSUE #39: APCU polyfill register a class to be autoloadable without taking into account dependency
The \APCUIterator stubs is register to be autoloadable but it extend a class that may very well not exist.
Therfore calling :
`class_exists('\APCUIterator');` 
will trigger a Fatal Error when \APCIterator isnt available.

For exemple, \Stash\Driver\Apc::isAvailable() rely on such a call.

I think, the stubs should only be know by Composer's autoloader if it's instanciable.

nicolas-grekas:
Fixes #39


stof:
:+1:

fgrandjean:
:+1:

src/Apcu/bootstrap.php
```diff
@@ -26,3 +26,13 @@ function apcu_inc($key, $step = 1, &$success = false) { return apc_inc($key, $st
     function apcu_sma_info($limited = false) { return apc_sma_info($limited); }
     function apcu_store($key, $var = null, $ttl = 0) { return apc_store($key, $var, $ttl); }
 }
+
+if (!class_exists('APCUIterator', false)) {
```
stof: Actually, the condition should also check for APCIterator

nicolas-grekas: I though that the existing check in https://github.com/symfony/polyfill/pull/40/files#diff-af8f2afac0a67b224b25386b2a75cc16R12 would be enough.
I don't get why, but it looks like so... hhvm?

keradus: also, `APCIterator` is available from `PECL apc >= 3.1.1`, so it doesn't exist on 3.0 for example

PR #48, opened 2016-03-25
php54 polyfill hex2bin(): don't return false if nul bytes found

### LINKED ISSUE #47: hex2bin() returns false if NUL bytes are in decoded string
hex2bin() returns false if NUL bytes are in decoded string. No error or exception is thrown which breaks code written for PHP 5.4+ that does not explicitly test for this condition, or which uses another test such as `preg_match('/^([a-f0-9]{2})*$/', $str)`.

This breaks binary-safety and is inconsistent with PHP's behavior.

fuhry:
Fixes #47

### DIFF
--- a/src/Php54/Php54.php
+++ b/src/Php54/Php54.php
@@ -33,10 +33,6 @@ final class Php54
 
         $data = pack('H*', $data);
 
-        if (false !== strpos($data, "\0")) {
-            return false;
-        }
-
         return $data;
     }
 }
--- a/tests/Php54/Php54Test.php
+++ b/tests/Php54/Php54Test.php
@@ -34,4 +34,14 @@ class Php54Test extends \PHPUnit_Framework_TestCase
     {
         $this->assertFalse(@class_uses('NotDefined'));
     }
+
+    public function testHexDecode()
+    {
+        $php54 = new \Symfony\Polyfill\Php54\Php54;
+        // issue #48: hex2bin() returns false if nul byte present
+        $this->assertEquals(
+            "\x61\x62\x00\x63\x64",
+            $php54->hex2bin("6162006364")
+		);
+    }
 }


binwiederhier:
:+1:

xabbuh:
If this is the expected behaviour, we will at least need a test to prevent regressions.

fuhry:
@xabbuh Test added.

binwiederhier:
I suspect that this is a rather big issue on older systems using Symfony. The current implementation apparently does not correctly convert hex strings with null bytes. I'd really love to see this in the next release. Thanks!

src/Php54/Php54.php
```diff
@@ -33,10 +33,6 @@ public static function hex2bin($data)
 
         $data = pack('H*', $data);
```
nicolas-grekas: No need for the $data intermediate step. Can you `return` directly please?

tests/Php54/Php54Test.php
```diff
@@ -34,4 +34,14 @@ public function testClassUsesInvalid()
...
         $this->assertFalse(@class_uses('NotDefined'));
     }
+
+    public function testHexDecode()
+    {
+        $php54 = new \Symfony\Polyfill\Php54\Php54;
+        // issue #48: hex2bin() returns false if nul byte present
+        $this->assertEquals(
+            "\x61\x62\x00\x63\x64",
+            $php54->hex2bin("6162006364")
```
nicolas-grekas: Please test against the `hex2bin` function directly. The test suite automatically tests against the polyfill and the native implementations.

binwiederhier:
In the interest of time, I took the liberty of making the requested changes in https://github.com/symfony/polyfill/issues/49 -- I hope that's alright, @fuhry.

@nicolas-grekas Looks alright now?

nicolas-grekas:
Thank you @fuhry

PR #49, opened 2016-03-29
Fix hex2bin return null

binwiederhier:
Fixes the hex2bin implementation as per #47/#48.
This supersedes https://github.com/symfony/polyfill/pull/48.

### DIFF
--- a/tests/Php54/Php54Test.php
+++ b/tests/Php54/Php54Test.php
@@ -35,13 +35,15 @@ class Php54Test extends \PHPUnit_Framework_TestCase
         $this->assertFalse(@class_uses('NotDefined'));
     }
 
-    public function testHexDecode()
+    public function testHexToBinValid()
     {
-        $php54 = new \Symfony\Polyfill\Php54\Php54;
-        // issue #48: hex2bin() returns false if nul byte present
-        $this->assertEquals(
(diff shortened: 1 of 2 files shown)


tests/Php54/Php54Test.php
```diff
@@ -34,4 +34,16 @@ public function testClassUsesInvalid()
     {
         $this->assertFalse(@class_uses('NotDefined'));
     }
+
+    public function testHexToBinValid()
+    {
+        $this->assertEquals("\x61\x62\x00\x63\x64", hex2bin("6162006364")); // With null byte, see #48
```
keradus: :-1: for referencing a PR in test.
If this case is somehow special it's enough to describe it here.

binwiederhier: Removed reference to pull/issue

binwiederhier: See https://github.com/symfony/polyfill/pull/49/commits/7d2358d11386b4fdb82dabeced8119ce9a221294

nicolas-grekas:
Thank you @binwiederhier.

binwiederhier:
@nicolas-grekas My pleasure.

When will this change be tagged/released? I am thinking about whether or not to apply a workaround or wait for your update.

PR #53, opened 2016-04-30
Allow paragonie/random_compat 2 in the php 7 subpackage as well

derrabus:
Greetings from the Symfony Live Cologne Hackday. :-)

This is a followup to the merged PR #44. I have added `"paragonie/random_compat": "~1.0|~2.0"` to the composer.json file of the `Php70` subpackage. Currently, only the full polyfill package allows version 2 of that library while the subpackage sill indicated compatibility with version 1 only.

### DIFF
--- a/src/Php70/composer.json
+++ b/src/Php70/composer.json
@@ -17,7 +17,7 @@
     ],
     "require": {
         "php": ">=5.3.3",
-        "paragonie/random_compat": "~1.0"
+        "paragonie/random_compat": "~1.0|~2.0"
     },
     "autoload": {
         "psr-4": { "Symfony\\Polyfill\\Php70\\": "" },


xabbuh:
👍 good catch

fabpot:
Thank you @derrabus.

AndrewCarterUK:
Note: The reason that a version 2 has been tagged is because of a security issue with version 1.

In the absence of alternatives, version 1 will use the OpenSSL API for generating random data. This isn't "fork-safe" and, in certain environments, different PHP processes could get the same output from this API. Obviously, this makes it predictable - and not suitable for use as a CSPRNG.

I'd personally have removed the version 1 dependency, but doing this would case an error under environments without access to `ext/libsodium`, `/dev/urandom` (windows or openbasedir restriction), `ext/mcrypt` or `ext/com_dotnet`. However, this error would be because there isn't a suitable CSPRNG available (which is important to know about).

PR #54, opened 2016-05-10
Add proper provides rules for mbstring and iconv polyfills

stof:
Once https://github.com/composer/composer/issues/5030 is resolved, this will allow to install `symfony/polyfill-mbstring` in a project and having it fulfill a ext-mbstring requirement set by another library not aware of the polyfill.
Without the bug fix, things work already when installing the polyfill from a VCS repo rather than from Packagist or when the polyfill was already installed locally (the bug is on Packagist side when loading metadata).
This would avoid having to do things like https://github.com/beberlei/assert/pull/149 which forces any project using the library to install the polyfill even if they expect to always run with ext-mbstring available (the library could just continue to require ext-mbstring and projects may then choose to require the polyfill to fulfill the dependency)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,10 @@
         "paragonie/random_compat": "~1.0|~2.0",
         "symfony/intl": "~2.3|~3.0"
     },
+    "provide": {
+        "ext-iconv": "7",
+        "ext-mbstring": "7"
+    },
     "replace": {
         "symfony/polyfill-apcu": "self.version",
         "symfony/polyfill-php54": "self.version",
--- a/src/Iconv/composer.json
+++ b/src/Iconv/composer.json
@@ -18,6 +18,9 @@
     "require": {
         "php": ">=5.3.3"
     },
+    "provide": {
+        "ext-iconv": "7"
+    },
     "autoload": {
         "psr-4": { "Symfony\\Polyfill\\Iconv\\": "" },
         "files": [ "bootstrap.php" ]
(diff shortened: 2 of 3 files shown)


stof:
note that I used ``7`` as version for the provided extension, because polyfill are compatible with the PHP 7 version of extensions. A better provide rule would be to give the version number of the extension itself, but core PHP extensions don't have a proper version number (which is why composer recommends to require ``*`` for them as the actual version being detected for them is not useful and may even depend on the way the extension was compiled AFAIK) and so I simply needed to specify a version number here.

stof:
@nicolas-grekas the test failure here does not look related to my change

gerryvdm:
I believe a package promising to provide some functionality, should really provide all expected functionality. The mbstring polyfill for example does not provide an expected level of compatibility in my opinion, given only UTF-8 is supported.

keradus:
@gerryvdm , native mbstring could provide different encodings, you should always check it via `mb_list_encodings`. for that, the polyfill provide only one of them (utf-8), which is fine.

nicolas-grekas:
I'm closing here as nothing changed on Composer's side in more than one year. Let's reopen when it'll be possible.

DvdGiessen:
With the recent release of Composer 2.0, this is now possible! See also https://github.com/composer/composer/issues/5030#issuecomment-580812337.

stof:
That's a good news, but we'll probably wait a bit for it to be adopted over Composer 1 to avoid issues.

Issue #58, opened 2016-05-18
Normalizer::isNormalized() should fail if given the form Normalizer::NONE.


gitlost:
For compatibility with the intl extension. See https://github.com/gitlost/tl-normalizer/blob/master/Symfony/Normalizer.php#L52

Issue #57, opened 2016-05-18
Normalizer::isNormalized() and Normalizer::normalize() should check for multibyte string function overload. 


gitlost:
Me again. The functions assume strlen() and substr() haven't been overloaded by mbstring. See https://github.com/gitlost/tl-normalizer/blob/master/Symfony/Normalizer.php#L57 and https://github.com/gitlost/tl-normalizer/blob/master/Symfony/Normalizer.php#L115

PR #59, opened 2016-05-18
Normalizer fixes from @gitlost 

### LINKED ISSUE #55: Normalizer::recompose() should reset the last combining class on ASCII.
This can be seen using the string "\xcc\x83\xc3\x92\xd5\x9b", which gets decomposed into "\xcc\x83\x4f\xcc\x80\xd5\x9b". On recompose, $lastUcls isn't reset on the ASCII "\x4f" so the string gets left in this decomposed form instead of the expected NFC normalized "\xcc\x83\xc3\x92\xd5\x9b".

The fix is reset the $lastUcls variable to zero on ASCII, see for instance the version of "Normalizer.php" I'm using in a fork of the WordPress plugin "tl-normalizer" https://github.com/gitlost/tl-normalizer/blob/master/Symfony/Normalizer.php#L184

nicolas-grekas:
Fixes #55, #57 and #58.

### DIFF
--- a/src/Intl/Normalizer/Normalizer.php
+++ b/src/Intl/Normalizer/Normalizer.php
@@ -122,6 +122,7 @@ class Normalizer
 
                 $result .= $lastUchr;
                 $lastUchr = $s[$i];
+                $lastUcls = 0;
                 ++$i;
                 continue;
             }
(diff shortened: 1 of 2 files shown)


stof:
Looks good to me. :+1: 

Btw, it would be great to check why the Grapheme tests are failing on PHP 7

nicolas-grekas:
I guess tests are failing until https://github.com/tchwork/utf8/pull/59 is ported here

fabpot:
Thank you @nicolas-grekas.

PR #60, opened 2016-05-18
bump version to 1.2

fabpot:


stof:
:+1:

stof:
We should include a patch fixing tests on PHP 7 before the release though IMO (see the discussion in #59)

nicolas-grekas:
:+1: after rebase and changelog update

PR #61, opened 2016-05-18
Normalizer::decompose() should reorder "recursive" combining chars

### LINKED ISSUE #56: Normalizer::decompose() should check that reordered sequences aren't themselves reordered.
This can be seen using the string "\xe0\xbd\xb6\xe0\xbe\x81", which currently gets decomposed into "\xe0\xbe\xb2\xe0\xbe\x80\xe0\xbd\xb1\xe0\xbe\x80", but the reordered sequences "\xe0\xbe\x80" and "\xe0\xbd\xb1" should themselves be reordered, to give the expected NFD form "\xe0\xbe\xb2\xe0\xbd\xb1\xe0\xbe\x80\xe0\xbe\x80".

I've put in a hacky fix for this on the version of "Normalizer.php" I'm using in the "tl-normalizer" WP plugin fork, which recurses when a reordered sequence has a reordering combining class, see https://github.com/gitlost/tl-normalizer/blob/master/Symfony/Normalizer.php#L237 passim, but I'm sure there's a better way of doing it.

nicolas-grekas:
Fix #56

### DIFF
--- a/src/Intl/Normalizer/Normalizer.php
+++ b/src/Intl/Normalizer/Normalizer.php
@@ -223,26 +223,10 @@ class Normalizer
             $uchr = substr($s, $i, $ulen);
             $i += $ulen;
 
-            if (isset($combClass[$uchr])) {
-                // Combining chars, for sorting
-
-                if (!isset($c[$combClass[$uchr]])) {
-                    $c[$combClass[$uchr]] = '';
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @nicolas-grekas.

PR #62, opened 2016-05-18
Fix tests for php 5.5.35/5.6.21/7.0.6

nicolas-grekas:
See https://github.com/tchwork/utf8/pull/59

### DIFF
--- a/src/Intl/Grapheme/Grapheme.php
+++ b/src/Intl/Grapheme/Grapheme.php
@@ -177,7 +177,11 @@ final class Grapheme
         if ($offset > 0) {
             $s = self::grapheme_substr($s, $offset);
         } elseif ($offset < 0) {
-            $offset = 0;
+            if (defined('HHVM_VERSION_ID') || PHP_VERSION_ID < 50535 || (50600 <= PHP_VERSION_ID && PHP_VERSION_ID < 50621) || (70000 <= PHP_VERSION_ID && PHP_VERSION_ID < 70006)) {
+                $offset = 0;
+            } else {
+                return false;
+            }
         }
 
         switch ($mode) {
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
See also https://3v4l.org/Svk6W

stof:
We should keep a test covering negative indexes, to ensure they produce the same behavior than PHP for each version (being accepted silently as 0, being rejected, or being taken from the end in PHP 7.1)

nicolas-grekas:
@stof fixed

src/Intl/Grapheme/Grapheme.php
```diff
@@ -177,7 +177,11 @@ private static function grapheme_position($s, $needle, $offset, $mode)
         if ($offset > 0) {
             $s = self::grapheme_substr($s, $offset);
         } elseif ($offset < 0) {
-            $offset = 0;
+            if (PHP_VERSION_ID < 50535 || PHP_VERSION_ID < 50621 || PHP_VERSION_ID < 70006) {
```
stof: this is wrong, as it will also match PHP 5.6.22 (as 50622 is smaller than 70006). It should be this:

```php
if (PHP_VERSION_ID < 50535 || (PHP_VERSION_ID >= 50600 && PHP_VERSION_ID < 50621) || (PHP_VERSION_ID >= 70000 && PHP_VERSION_ID < 70006)) {
    $offset = 0;
} else {
    return false;
}
```

nicolas-grekas: Pff indeed, fixed

stof:
:+1:

fabpot:
Thank you @nicolas-grekas.

PR #65, opened 2016-06-07
Bypass iconv for some charsets in mb_strlen

### LINKED ISSUE #63: mb_strlen: iconv_strlen triggering notice in certain situations
As suggested by @stof in https://github.com/php-amqplib/php-amqplib/issues/72#issuecomment-220060121, I'm opening this ticket.

I can't really tell how to replicate this - I ran into it when installing RabbitMQ and not having `mbstring` extension installed.

Seems like theres being passed an "illegal" character in the string and its not being considered. This triggers a Notice, and in RabbitMQ case, it throws an exception.

```
PHP Notice:  iconv_strlen(): Detected an illegal character in input string in …/vendor/symfony/polyfill-mbstring/Mbstring.php on line 338
```

```php
    public static function mb_strlen($s, $encoding = null)
    {
        $encoding = self::getEncoding($encoding);
        var_dump($s);

        return iconv_strlen($s, $encoding);
    }
```

```
string(14) "\000\000\000\000�\000
\000
\000    \000"
```

Installing the `mbstring` module prevents RabbitMQ from throwing the Exception, which I guess means this *polyfill* method is not really replacing the module (?)

nicolas-grekas:
Fix #63

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -333,7 +333,11 @@ final class Mbstring
 
     public static function mb_strlen($s, $encoding = null)
     {
-        $encoding = self::getEncoding($encoding);
+        switch ($encoding = self::getEncoding($encoding)) {
+            case 'ASCII':
+            case 'CP850':
+                return strlen($s);
+        }
 
         return iconv_strlen($s, $encoding);
     }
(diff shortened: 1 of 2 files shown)


stof:
:+1:

fabpot:
Thank you @nicolas-grekas.

PR #68, opened 2016-07-03
use phpseclib/mcrypt_compat

terrafrost:
The fact that mcrypt, as an extension, is becoming less and less common on hosts, there is also an accepted RFC to deprecate mcrypt on PHP 7.1 and to remove it all together in the next major release: https://wiki.php.net/rfc/mcrypt-viking-funeral .

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
         "php": ">=5.3.3",
         "ircmaxell/password-compat": "~1.0",
         "paragonie/random_compat": "~1.0|~2.0",
+        "phpseclib/mcrypt_compat": "~1.0",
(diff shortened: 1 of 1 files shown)


keradus:
From readme:
> This project backports features found in the latest PHP versions (...)

This project is to backport features from new PHP to older one.
Your proposal is the other way around. There was some feature, it's removed in new PHP, so let bring it back to life. I'm :-1: for that.

stof:
I agree with @keradus. People needing to keep using deprecated APIs in versions in which they are removed should use the paragonie library themselves, but symfony/polyfill is meant to work the other way (using the new API in your code), so I'm also :-1: here

PR #69, opened 2016-08-17
Test with PHP 7.1 + minor fixes

nicolas-grekas:


nicolas-grekas:
@ChadSikorra @csarrazi FYI, PHP 7.1 fixed encoding leading/trailing space and a carriage returns for DN. I think we should only polyfill the correct behavior. This PR does just that, with an adaptive test case.

ChadSikorra:
Good catch. I forgot to open an issue for polyfill after I sent the PR to fix that for PHP :/

Issue #72, opened 2016-08-24
iconv_substr(): Detected an illegal character in input string


gectus:
Hello

I have the same error https://github.com/symfony/polyfill/issues/63

```
[2016-08-24 15:03:51] lumen.ERROR: ErrorException: iconv_substr(): Detected an illegal character in input string in /var/www/app-storage/vendor/symfony/polyfill/src/Mbstring/Mbstring.php:417
Stack trace:
#0 [internal function]: Laravel\Lumen\Application->Laravel\Lumen\Concerns\{closure}(8, 'iconv_substr():...', '/var/www/heaven...', 417, Array)
#1 /var/www/app-storage/vendor/symfony/polyfill/src/Mbstring/Mbstring.php(417): iconv_substr('\x01\x00\x00\x00\x00\x01\xE5', 0, 1, 'ASCII')
#2 /var/www/app-storage/vendor/symfony/polyfill/src/Mbstring/bootstrap.php(36): Symfony\Polyfill\Mbstring\Mbstring::mb_substr('\x01\x00\x00\x00\x00\x01\xE5', 0, 1, 'ASCII')
#3 /var/www/app-storage/vendor/php-amqplib/php-amqplib/PhpAmqpLib/Wire/AMQPReader.php(163): mb_substr('\x01\x00\x00\x00\x00\x01\xE5', 0, 1, 'ASCII')
```

nicolas-grekas:
which version of symfony/polyfill are you using?

gectus:
```
"repositories": [
        {
            "type": "git",
            "url": "https://github.com/symfony/polyfill.git"
        }
    ],
    "require": {
        "symfony/polyfill": "dev-master"
    },
```

stof:
please run ``composer show | grep symfony/polyfill``, which will also display the exact commit being used, because dev-master is a moving target, and so what you use depends on when you ran a composer update.

Btw, you don't need to register a git repository to use this package. It is available on Packagist.

gectus:
`symfony/polyfill                          dev-master 2e13b98 Symfony polyfills backporting features to lower PHP versions`

stof:
OK, this is similar to #63, but this is actually happening in another polyfilled method

bensquire:
Same issue detected when using the previously referenced, php-amqplib/php-amqplib.

woshiwenquan:
PHP installation `mbstring ` extension.

bensquire:
[...]

PR #74, opened 2016-10-06
[Mbstring] Fix iconv_substr(): Detected an illegal character in input string

vitorsdcs:
#72


keradus:
Please, could we add new tests to check introduced changes and ensure it won't be broken in the future (thanks, for example, to some refactoring) ?

vitorsdcs:
Hey @keradus. I added the tests and they worked on my machine. But for some reason they're failing on AppVeyor and Travis. Any idea on what might be happening?

nicolas-grekas:
Thank you @vitorsdcs.

PR #76, opened 2016-11-05
Polyfill for is_iterable

keradus:


src/Php71/composer.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "symfony/polyfill-php71",
```
stof: you are missing the replace rule for this package in the main package

src/Php71/composer.json
```diff
@@ -0,0 +1,32 @@
...
+            "email": "p@tchwork.com"
+        },
+        {
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "require": {
+        "php": ">=5.3.3",
+        "paragonie/random_compat": "~1.0|~2.0"
```
stof: this looks wrong to me

src/Php71/Php71.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ */
+final class Php71
+{
+    public static function is_iterable($var)
+    {
+        return true === is_array($var) || $var instanceof \Traversable;
```
stof: ``true ===`` is useless

src/Php71/composer.json
keradus: my bad, copy paste :|

src/Php71/composer.json
keradus: somehow I missed that, nice catch ;)

keradus:
Thanks @stof !

src/Php71/composer.json
```diff
@@ -0,0 +1,31 @@
...
+        "php": ">=5.3.3"
+    },
+    "autoload": {
+        "psr-4": { "Symfony\\Polyfill\\Php71\\": "" },
+        "files": [ "bootstrap.php" ]
+    },
+    "minimum-stability": "dev",
+    "extra": {
+        "branch-alias": {
+            "dev-master": "1.2-dev"
```
stof: Adding polyfills for new APIs means that we should bump the version to 1.3 for the library

keradus: That's why I hate aliasing. I didn't modified it as I don't manage releasing cycle of polyfill.
Do you want me to update this in PR? Then it would make sense to update this in other `composer.json` files, would you agree ?

keradus: So, let me update it only in this one file and let the PR wait for `1.3@dev` happen

nicolas-grekas:
:+1:
Note that before releasing 1.3, we maybe should look at new functions in 7.1 that could be polyfilled:
https://github.com/php/php-src/blob/php-7.1.0RC5/UPGRADING#L237

fabpot:
Thank you @keradus.

PR #82, opened 2017-01-19
Added PHP_INT_MIN constant

Fleshgrinder:
The `PHP_INT_MIN` constant is available in PHP 7.0.0 onwards but not in any previous release. Seems like an oversight since that value is constantly being created in the test cases. However, I did not change those since we should not rely on code that is being tested.

### DIFF
--- a/tests/Php70/Php70Test.php
+++ b/tests/Php70/Php70Test.php
@@ -16,6 +16,7 @@ class Php70Test extends \PHPUnit_Framework_TestCase
     public function testPhpIntMin()
     {
         $this->assertTrue(defined('PHP_INT_MIN'));
+        $this->assertSame(~PHP_INT_MAX, PHP_INT_MIN);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
can you please add a "defined()" check around?

Fleshgrinder:
I guess I did this in a hurry, done. 😸

tests/Php70/Php70Test.php
```diff
@@ -13,6 +13,11 @@
 
 class Php70Test extends \PHPUnit_Framework_TestCase
 {
+    public function testPhpIntMin()
+    {
+        $this->assertTrue(defined('PHP_INT_MIN'));
```
nicolas-grekas: missing: assertSame(~PHP_INT_MAX, PHP_INT_MIN);

fabpot:
Thank you @Fleshgrinder.

PR #84, opened 2017-04-30
Feature/if exists

korobochkin:
Hi.
Thank you for the great libraries and code first :) My pool request is a little tweak for this code.

I developing a WordPress plugin for wp.org which uses a lot of Symfony components. And I have a trouble when deploying my code into wp.org plugins directory. The problem appears every time when I commiting to wp-org. It simply not allows me to make a commit actually 😀 This happens because wp-org and their SVN repo run `php -l` command for each file in commit while you commiting. And since vendor folder contains files and classes like `ArithmeticError.php` — this triggering error. Just PHP 7 already have `ArithmeticError` class and there is no `if class exists` check for all of this classes and methods.

Please allow me deploying :) Right now I make this changes manually every time when I'm deploying.

And by the way. Here is plugin that I'm talking about: https://wordpress.org/plugins/setka-editor/
The source code with fixed stubs: https://plugins.trac.wordpress.org/browser/setka-editor/tags/1.7.9/vendor/symfony/polyfill-php70/Resources/stubs/

### DIFF
--- a/src/Intl/Icu/bootstrap.php
+++ b/src/Intl/Icu/bootstrap.php
@@ -12,8 +12,17 @@
 use Symfony\Component\Intl\Globals\IntlGlobals;
 
 if (!function_exists('intl_is_failure')) {
-    function intl_is_failure($errorCode) { return IntlGlobals::isFailure($errorCode); }
-    function intl_get_error_code() { return IntlGlobals::getErrorCode(); }
-    function intl_get_error_message() { return IntlGlobals::getErrorMessage(); }
-    function intl_error_name($errorCode) { return IntlGlobals::getErrorName($errorCode); }
+	function intl_is_failure($errorCode) { return IntlGlobals::isFailure($errorCode); }
+}
+
+if(!function_exists('intl_get_error_code')) {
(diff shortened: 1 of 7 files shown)


korobochkin:
Looks like travis build is failed. Interesting. Need some time to investigate that happened.

xabbuh:
This change is a 👎 for me. You should simply not run static analyses on third-party code.

korobochkin:
But why? Unfortunately I can't change this behaviour in wp.org.
And by the way most of other code in this repository, for example https://github.com/symfony/polyfill/blob/master/src/Php54/bootstrap.php have `if function exists` statements.

keradus:
Because there is no autoloading mechanism for function.
So to polyfill a function, bootstrap file is always executed via composer autoloader, and then conditionally function is created.
For classes, autoloading works, so file is executed only when there is no native class with that name.

robfrawley:
@korobochkin Word Press needs to fix their static analysis tools. This code works as expected when *run* due to the nature of the PHP autoloader. I'd suggest opening this issue in the respective Word Press issue tracker.

nicolas-grekas:
Closing as explained, this should be fixed on Wordpress' side.

PR #85, opened 2017-05-29
Fix CI

nicolas-grekas:


stof:
CI is still red

nicolas-grekas:
I know, just wait a bit :)

.travis.yml
```diff
@@ -2,6 +2,11 @@ language: php
 
 sudo: false
 
+cache:
+    directories:
+        - vendor
```
stof: this can lead to weird things in composer. If the goal is to cache the PHPUnit download, put it outside `vendor` IMO

nicolas-grekas: the component is small enough to not trigger strange things IMHO. We'll see :)

PR #87, opened 2017-06-08
Add Php72 polyfill, abandon Xml polyfill

nicolas-grekas:


src/Php72/README.md
```diff
@@ -0,0 +1,17 @@
+Symfony Polyfill / Php71
```
keradus: 72

src/Php72/Php72.php
```diff
@@ -9,15 +9,25 @@
...
+    {
+        return function_exists('posix_isatty') && @posix_isatty($stream);
+    }
+
+    public static function sapi_windows_vt100_support()
+    {
+        return false;
+    }
+
     public static function utf8_encode($s)
```
stof: this looks weird here. This function is not a function added in PHP 7.2, is it ?

src/Php72/README.md
```diff
@@ -0,0 +1,17 @@
+Symfony Polyfill / Php72
+========================
+
+This component provides functions unavailable in releases prior to PHP 7.2:
+
+- [`stream_isatty`](https://php.net/stream_isatty)
+- [`sapi_windows_vt100_support`](https://php.net/sapi_windows_vt100_support)
+- [`utf8_encode`](https://php.net/utf8_encode)
+- [`utf8_decode`](https://php.net/utf8_decode)
```
stof: this sentence looks weird to me

src/Php72/composer.json
```diff
@@ -0,0 +1,34 @@
...
+        {
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "require": {
+        "php": ">=5.3.3"
+    },
+    "replace": {
+        "symfony/polyfill-xml": "self.version"
```
stof: you should not have this replacement here, as the metapackage requires this package, which will forbid resolving deps

src/Php72/Php72.php
nicolas-grekas: yes, see https://github.com/php/php-src/blob/master/NEWS#L227

src/Php72/composer.json
nicolas-grekas: removed

src/Php72/README.md
nicolas-grekas: Tweaked. Any suggestion?

stof: Maybe just add a small note saying `(available in the optional XML extension earlier)` ?

nicolas-grekas: updated

.travis.yml
```diff
@@ -9,7 +9,10 @@ cache:
 
 matrix:
     include:
-        - php: hhvm
+        - php: hhvm-3.18
+          sudo: required
+          dist: trusty
+          group: edge
```
fabpot: What about removing HHVM altogether?

tests/Php70/Php70Test.php
```diff
@@ -34,18 +34,24 @@ public function provideIntdiv()
     }
 
     /**
-     * @expectedException ArithmeticError
+     * @expectedException \ArithmeticError
      */
     public function testIntdivArithmetic()
     {
+        if (defined('HHVM_VERSION')) {
+            $this->markTestSkipped('HHVM fails here.');
```
fabpot: Removing HHVM really makes sense as just skipping tests that fail for it just means we don't support it anyway.

nicolas-grekas: just one of them :)

tests/Php70/Php70Test.php
```diff
@@ -34,18 +34,24 @@ public function provideIntdiv()
...
         intdiv(~PHP_INT_MAX, -1);
     }
 
     /**
-     * @expectedException DivisionByZeroError
+     * @expectedException \DivisionByZeroError
      */
     public function testIntdivByZero()
     {
+        if (defined('HHVM_VERSION')) {
```
fabpot: ok, here is another one :)

.travis.yml
fabpot: And as we are bumping version to 1.4, it looks like this is a great occasion to do so.

fabpot:
Thank you @nicolas-grekas.

keradus:
awesome !

PR #88, opened 2017-06-13
Add PHP_OS_FAMILY polyfill

keradus:


bcremer:
See https://github.com/sebastianbergmann/environment/issues/21 for PHPUnit's implementation.

keradus:
ref https://github.com/sebastianbergmann/phpunit/issues/2697#issuecomment-308098725

src/Php72/Php72.php
```diff
@@ -60,4 +61,23 @@ public static function utf8_decode($s)
...
     }
+
+    public static function php_os_family()
+    {
+        if ('\\' === DIRECTORY_SEPARATOR) {
+            return 'Windows';
+        }
+
+        $map = array(
+            'Darwin' => 'Mac',
```
nicolas-grekas: this changed recently to Darwin, see http://git.php.net/?p=php-src.git;a=commitdiff;h=a251d9b2c11c2ab53b0214d728c40d70cd06567c;hp=ef4b2fc283ddaf9bd692015f1db6dad52171c3ce

src/Php72/bootstrap.php
```diff
@@ -21,3 +21,6 @@ function stream_isatty($stream) { return function_exists('posix_isatty') && @pos
     function utf8_encode($s) { return p\Php72::utf8_encode($s); }
     function utf8_decode($s) { return p\Php72::utf8_decode($s); }
 }
+if (!defined('PHP_OS_FAMILY')) {
```
nicolas-grekas: needs to be move inside the if added by #90

src/Php72/Php72.php
keradus: and before to Mac... http://git.php.net/?p=php-src.git;a=commitdiff;h=362d2e42a02fe018a7d1261a53ff59b73fed91f6

they cannot decide - osx, mac, darwin now... changing

stof:
We need to bump the dev version to 1.5 (best would be to do it before merging to have the right version in the GH merge commit)

stof:
@keradus no. My comment was a reminder for maintainers to bump the version before the merge so that we get the right version (as I may not be the one doing the merge)

nicolas-grekas:
Here we are: #92

keradus:
then also changelog need to be updated, but after merge ;)

fabpot:
Thank you @keradus.

fabpot:
CHANGELOG updated in aae8251

PR #89, opened 2017-06-13
PHP_INT_MIN - add missing docs

keradus:

PR #91, opened 2017-06-14
Fix the readme content

stof:
#89 reordered some lines, but it did not keep both lines about `random_bytes` together


nicolas-grekas:
Good catch, thanks @stof.

keradus:
thanks @stof

PR #95, opened 2017-06-19
Fix apcu function when apc ones are polyfilled with Zend Server's Data Cache

### LINKED ISSUE #94: Zend Server and APC functionality
Zend Server tries to implement compatibility with apc by registering compatibility functions with the same name as the APC functions.

http://files.zend.com/help/Zend-Server-6/content/zend_data_cache_-_configuration_directives.htm

Specifically, the option is "zend_datacache.apc_compatibility".  It looks like it gets aliased to zend_shm_cache_store, which mostly works.  Unfortunately, it doesn't support the array form of apc_store(), so every call results in errors:

"apc_store() expects parameter 1 to be string, array given"

It looks like this is a problem that other [projects](https://github.com/UnionOfRAD/lithium/issues/1119) have dealt with, and it looks like it's one that's been going on for 8 years or so, so I'm not expecting Zend to fix it anytime soon.

My inclination would be to add a check to see if the parameter to apcu_add() or apcu_store() is an array, and if so to call the apc function individually for each item in the array.  It would maintain compatibility with APC, and fix the issues with Zend Server.  It's a little bit of a hack, though, and I figured I'd ask before submitting a PR.

Alternatively, I could add support for Zend Data Cache to the Symfony Cache Component.

kategray:
Fixes #94 
Detect if Zend Server's Data Cache is enabled, and work around it's lack of support for the array syntax.

### DIFF
--- a/src/Apcu/bootstrap.php
+++ b/src/Apcu/bootstrap.php
@@ -9,22 +9,32 @@
  * file that was distributed with this source code.
  */
 
+use Symfony\Polyfill\Apcu as p;
+
 if (!extension_loaded('apc') && !extension_loaded('apcu')) {
     return;
 }
 
 if (!function_exists('apcu_add')) {
-    function apcu_add($key, $var = null, $ttl = 0) { return apc_add($key, $var, $ttl); }
+    if (extension_loaded('Zend Data Cache')) {
+        function apcu_add($key, $var = null, $ttl = 0) { return p\Apcu::apcu_add($key, $var, $ttl); }
+        function apcu_delete($key) { return p\Apcu::apcu_delete($key); }
+        function apcu_exists($keys) { return p\Apcu::apcu_exists($keys); }
+        function apcu_fetch($key, &$success = null) { return p\Apcu::apcu_fetch($key, $success); }
+        function apcu_store($key, $var = null, $ttl = 0) { return p\Apcu::apcu_store($key, $var, $ttl); }
+    } else {
+        function apcu_add($key, $var = null, $ttl = 0) { return apc_add($key, $var, $ttl); }
+        function apcu_delete($key) { return apc_delete($key); }
+        function apcu_exists($keys) { return apc_exists($keys); }
+        function apcu_fetch($key, &$success = null) { return apc_fetch($key, $success); }
+        function apcu_store($key, $var = null, $ttl = 0) { return apc_store($key, $var, $ttl); }
+    }
     function apcu_cache_info($limited = false) { return apc_cache_info('user', $limited); }
     function apcu_cas($key, $old, $new) { return apc_cas($key, $old, $new); }
     function apcu_clear_cache() { return apc_clear_cache('user'); }
     function apcu_dec($key, $step = 1, &$success = false) { return apc_dec($key, $step, $success); }
-    function apcu_delete($key) { return apc_delete($key); }
-    function apcu_exists($keys) { return apc_exists($keys); }
-    function apcu_fetch($key, &$success = false) { return apc_fetch($key, $success); }
     function apcu_inc($key, $step = 1, &$success = false) { return apc_inc($key, $step, $success); }
     function apcu_sma_info($limited = false) { return apc_sma_info($limited); }
-    function apcu_store($key, $var = null, $ttl = 0) { return apc_store($key, $var, $ttl); }
 }
 
 if (!class_exists('APCUIterator', false) && class_exists('APCIterator', false)) {
--- a/tests/Apcu/ApcuTest.php
+++ b/tests/Apcu/ApcuTest.php
@@ -37,6 +37,33 @@ class ApcuTest extends \PHPUnit_Framework_TestCase
         $this->assertArrayHasKey('cache_list', apcu_cache_info());
     }
 
+    public function testArrayCompatibility()
+    {
+        $data = array (
+            'key1' => 'value1',
+            'key2' => 'value2',
+        );
+        apcu_delete(array_keys($data));
+        apcu_add($data);
+
+        foreach ($data as $key => $value) {
+            $this->assertEquals($value, apcu_fetch($key));
+        }
+
+        $data = array (
+            'key1' => 'value2',
+            'key2' => 'value3',
+        );
+        apcu_store($data);
+
+        $this->assertEquals($data, apcu_fetch(array_keys($data)));
+        $this->assertSame(array('key1' => true, 'key2' => true), apcu_exists(array('key1', 'key2', 'key3')));
+
+        apcu_delete(array_keys($data));
+        $this->assertSame(array(), apcu_exists(array_keys($data)));
+
+    }
+
     public function testAPCUIterator()
     {
         $key = __CLASS__;
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,18 +18,20 @@ matrix:
     fast_finish: true
 
 before_install:
-    - INI_FILE=~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
-    - echo memory_limit = -1 >> $INI_FILE
-    - echo session.gc_probability = 0 >> $INI_FILE
-    - echo extension = ldap.so >> $INI_FILE
+    - INI=~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+    - echo memory_limit = -1 >> $INI
+    - echo session.gc_probability = 0 >> $INI
+    - echo extension = ldap.so >> $INI
+    - echo apc.enable_cli = 1 >> $INI
     - phpenv config-rm xdebug.ini || echo "xdebug not available"
-    - if [[ $TRAVIS_PHP_VERSION = 5.* ]]; then (echo yes | pecl install -f apcu-4.0.11 && echo apc.enable_cli = 1 >> $INI_FILE) || echo "Let's continue without apcu extension"; fi;
-    - if [[ $TRAVIS_PHP_VERSION = 7.0 ]]; then (echo yes | pecl install -f apcu-5.1.5 && echo apc.enable_cli = 1 >> $INI_FILE) || echo "Let's continue without apcu extension"; fi;
+    - if [[ $TRAVIS_PHP_VERSION = 5.* ]]; then echo yes | pecl install -f apcu-4.0.11; fi
+    - if [[ $TRAVIS_PHP_VERSION = 7.0 ]]; then echo yes | pecl install -f apcu-5.1.5; fi
+    - if [[ $TRAVIS_PHP_VERSION = 7.0 ]]; then echo yes | pecl install channel://pecl.php.net/apcu_bc-1.0.3 && echo extension = apc.so >> $INI; fi
     - php -i
 
 install:
-    - if [[ $TRAVIS_BRANCH = master ]]; then export COMPOSER_ROOT_VERSION=dev-master; else export COMPOSER_ROOT_VERSION=$TRAVIS_BRANCH.x-dev; fi;
-    - composer --prefer-source install;
+    - if [[ $TRAVIS_BRANCH = master ]]; then export COMPOSER_ROOT_VERSION=dev-master; else export COMPOSER_ROOT_VERSION=$TRAVIS_BRANCH.x-dev; fi
+    - composer --prefer-source install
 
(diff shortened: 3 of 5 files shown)


src/Apcu/bootstrap.php
```diff
@@ -11,10 +11,14 @@
 
 if (!extension_loaded('apc') && !extension_loaded('apcu')) {
     return;
-}
-
-if (!function_exists('apcu_add')) {
+} else if (extension_loaded ('Zend Data Cache')) {
+    // Zend Data Cache refuses to load in the CLI, regardless of the configuration
+    if ('cli' != PHP_SAPI && false !== ini_get('zend_datacache.apc_compatibility')) {
+        require_once __DIR__ . DIRECTORY_SEPARATOR . 'bootstrap_zend.php';
```
keradus: do we really need to load 2 files ?

src/Apcu/bootstrap.php
```diff
@@ -11,10 +11,14 @@
 
 if (!extension_loaded('apc') && !extension_loaded('apcu')) {
     return;
-}
-
-if (!function_exists('apcu_add')) {
+} else if (extension_loaded ('Zend Data Cache')) {
+    // Zend Data Cache refuses to load in the CLI, regardless of the configuration
+    if ('cli' != PHP_SAPI && false !== ini_get('zend_datacache.apc_compatibility')) {
```
keradus:
```
ker@dus:~$ cat x.php 
<?php echo PHP_SAPI;
ker@dus:~$ phpdbg -qrr x.php 
phpdbg
```

src/Apcu/bootstrap.php
kategray: I'm open to suggestions.  I was trying to keep the APCUIterator, as well as trying to avoid the overhead associated with the Zend path on non-zend servers.  I could have moved it to composer.json and loaded both files that way, or I could have combined the functionality into a single file, with different function definitions depending on which path was taken.

I also didn't want to force APC to have to run the compatibility code, either.

src/Apcu/bootstrap_zend.php
```diff
@@ -0,0 +1,85 @@
...
+    function apcu_add($key, $var = null, $ttl = 0) {
+        if (is_array($key)) {
+            $errors = array ();
+            foreach ($key as $cache_key => $cache_value) {
+                if (!apc_add ($cache_key, $cache_value, $ttl)) {
+                    $errors[] = $cache_key;
+                }
+            }
+            return $errors;
+        } else {
```
keradus: no need for else after return

src/Apcu/bootstrap_zend.php
```diff
@@ -0,0 +1,85 @@
...
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Bootstrap for Zend Server Data Cache
+ */
+if (!function_exists('apc_add')) {
+    function apcu_add($key, $var = null, $ttl = 0) {
+        if (is_array($key)) {
+            $errors = array ();
```
keradus: weird spacing

src/Apcu/bootstrap_zend.php
```diff
@@ -0,0 +1,85 @@
...
+        return apc_fetch($key, $success);
+    }
+
+    function apcu_cache_info($limited = false) { return apc_cache_info('user', $limited); }
+    function apcu_cas($key, $old, $new) { return apc_cas($key, $old, $new); }
+    function apcu_clear_cache() { return apc_clear_cache('user'); }
+    function apcu_dec($key, $step = 1, &$success = false) { return apc_dec($key, $step, $success); }
+    function apcu_delete($key) { return apc_delete($key); }
+
+
```
keradus: not needed

src/Apcu/bootstrap_zend.php
```diff
@@ -0,0 +1,85 @@
...
+    function apcu_inc($key, $step = 1, &$success = false) { return apc_inc($key, $step, $success); }
+    function apcu_sma_info($limited = false) { return apc_sma_info($limited); }
+}
```
keradus: new line is missing

src/Apcu/bootstrap.php
kategray: I'm aware of that limitation.  Unfortunately, Zend Data Cache does the checking internally, and does not have the option that APC does to force enabling in the cli.

The point was to avoid adding the compatibility to the cli, since it wouldn't work anyway.  It will fail gracefully if that check fails - the apc functions will simply return false, the same as any other failure.

tests/Apcu/ApcuTest.php
```diff
@@ -37,6 +37,30 @@ public function testApcu()
...
+            $this->assertEquals($value, apcu_fetch($key));
+        }
+
+        // Test add and array fetch.  As key is a duplicate, it should not be overwritten.  key3 should be created.
+        apcu_add($test2);
+        $correct = array ('key' => 'value1', 'key2' => 'value2', 'key3' => 'value4');
+        $this->assertEquals($correct, apcu_fetch(array_keys($correct)));
+    }
+
+
```
keradus: double new line is wrong

tests/Apcu/ApcuTest.php
```diff
@@ -37,6 +37,30 @@ public function testApcu()
         $this->assertArrayHasKey('cache_list', apcu_cache_info());
     }
 
+    public function testArrayCompatibility() {
+        // Test that arrays are properly written and read
```
keradus: would be nice to test both, zend and non-zend logic

src/Apcu/bootstrap_zend.php
kategray: You're right - it's not required.  I put it there for clarity, and included the brackets on the if statement (which are also not required) for clarity.

src/Apcu/bootstrap.php
keradus: why not add `phpdbg !== PHP_SAPI` ?

src/Apcu/bootstrap_zend.php
keradus: brackets on if are standards
extra else is not needed, producing larger bytecode and so on

tests/Apcu/ApcuTest.php
kategray: It would. phpUnit makes it hard to override built in functions, and I was trying to do a minor contribution rather than an overhaul.  Ideally, I'd like it to be in an object that can be broken down better for unit testing, but that's more than just a minor fix.

There's also the issue of Zend refusing to run CLI, requiring a web-based unit test.

src/Apcu/bootstrap.php
kategray: Exclude that one, as well as cli?

src/Apcu/bootstrap_zend.php
kategray: What's not needed?

src/Apcu/bootstrap.php
```diff
@@ -11,20 +11,56 @@
...
-    function apcu_dec($key, $step = 1, &$success = false) { return apc_dec($key, $step, $success); }
-    function apcu_delete($key) { return apc_delete($key); }
-    function apcu_exists($keys) { return apc_exists($keys); }
-    function apcu_fetch($key, &$success = false) { return apc_fetch($key, $success); }
-    function apcu_inc($key, $step = 1, &$success = false) { return apc_inc($key, $step, $success); }
-    function apcu_sma_info($limited = false) { return apc_sma_info($limited); }
-    function apcu_store($key, $var = null, $ttl = 0) { return apc_store($key, $var, $ttl); }
+} elseif (extension_loaded('Zend Data Cache')) {
+    // Zend Data Cache refuses to load in the CLI, regardless of the configuration
+    if ('phpdbg' != PHP_SAPI && 'cli' != PHP_SAPI && false !== ini_get('zend_datacache.apc_compatibility')) {
```
stof: please use struct comparison

src/Apcu/bootstrap.php
```diff
@@ -11,20 +11,56 @@
...
-    function apcu_cas($key, $old, $new) { return apc_cas($key, $old, $new); }
-    function apcu_clear_cache() { return apc_clear_cache('user'); }
-    function apcu_dec($key, $step = 1, &$success = false) { return apc_dec($key, $step, $success); }
-    function apcu_delete($key) { return apc_delete($key); }
-    function apcu_exists($keys) { return apc_exists($keys); }
-    function apcu_fetch($key, &$success = false) { return apc_fetch($key, $success); }
-    function apcu_inc($key, $step = 1, &$success = false) { return apc_inc($key, $step, $success); }
-    function apcu_sma_info($limited = false) { return apc_sma_info($limited); }
-    function apcu_store($key, $var = null, $ttl = 0) { return apc_store($key, $var, $ttl); }
+} elseif (extension_loaded('Zend Data Cache')) {
```
stof: `if` is enough as there is an early return

src/Apcu/bootstrap_zend.php
keradus: there were too much empty lines

src/Apcu/bootstrap_zend.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    function apcu_inc($key, $step = 1, &$success = false)
+    {
+        return apc_inc($key, $step, $success);
+    }
+
+    function apcu_sma_info($limited = false)
+    {
+        return apc_sma_info($limited);
```
stof: We should avoid defining all these functions twice, as they are the same when using Zend Server or no

src/Apcu/bootstrap.php
kategray: `'phpdbg' !== PHP_SAPI` ?

stof:
The APCUIterator polyfill should not be defined in case you skip the loading of other parts of the polyfill due to being in the CLI or phpdbg on Zend Server.

src/Apcu/bootstrap.php
stof: yes. Sorry, there was a typo in my comment. It should have said `strict`, not `struct`

src/Apcu/bootstrap_zend.php
kategray: Will do.

kategray:
Zend Server doesn't define the APCIterator class, so that code doesn't run.

src/Apcu/bootstrap.php
```diff
@@ -13,18 +13,61 @@
     return;
 }
 
+if (extension_loaded('Zend Data Cache')) {
+    // Zend Data Cache refuses to load in the CLI, regardless of the configuration
+    if ('phpdbg' !== PHP_SAPI && 'cli' !== PHP_SAPI && false !== ini_get('zend_datacache.apc_compatibility')) {
+        require_once __DIR__.DIRECTORY_SEPARATOR.'bootstrap_zend.php';
```
nicolas-grekas: instead of a separate bootstrap file, we should use the same strategy as other polyfills (see e.g. https://github.com/symfony/polyfill/blob/master/src/Iconv/bootstrap.php): a single bootstrap file that delegates "complex" implementations to static methods.

src/Apcu/bootstrap.php
```diff
@@ -13,18 +13,61 @@
...
+    }
+}
+
 if (!function_exists('apcu_add')) {
-    function apcu_add($key, $var = null, $ttl = 0) { return apc_add($key, $var, $ttl); }
```
nicolas-grekas: please keep these on a single line as they were, that'll help review, and as a change is unrelated to what you want to achieve here

tests/Apcu/ApcuTest.php
nicolas-grekas: the test suite already automatically does it, when using the static methods strategy + single line style in the bootstrap file

nicolas-grekas:
@kategray any chance you might finish this PR? please advise, I can take over.

kategray:
I'm swamped for a couple weeks.  I can do it after that, or you're welcome to.  Whichever you prefer.

nicolas-grekas:
@kategray no pb, I took over :)
I removed the SAPI-related checks because I don't get them.
Should be green and ready.

nicolas-grekas:
@kategray PS: http://sanfrancisco2017.live.symfony.com ;)

fabpot:
Thank you @kategray.

PR #97, opened 2017-07-11
[7.2] Add polyfill for spl_object_id()

nicolas-grekas:
As [discussed on php-internals](https://externals.io/message/99751) and implemented in https://github.com/php/php-src/pull/2611.
Implementation [borrowed from VarDumper](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/VarDumper/Cloner/VarCloner.php#L300).

### DIFF
--- a/src/Php72/Php72.php
+++ b/src/Php72/Php72.php
@@ -19,6 +19,8 @@ namespace Symfony\Polyfill\Php72;
  */
 final class Php72
 {
+    private static $hashMask;
+
     public static function utf8_encode($s)
     {
         $s .= $s;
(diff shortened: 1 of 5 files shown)


src/Php72/Php72.php
```diff
@@ -80,4 +83,31 @@ public static function php_os_family()
 
         return isset($map[PHP_OS]) ? $map[PHP_OS] : 'Unknown';
     }
+
+    public static function spl_object_id($object)
+    {
+        if (null === self::$hashMask) {
```
lyrixx: You can use an early return here.

nicolas-grekas: not sure it would help, L108 and L111 are not the same, so that would require duplicating L111, is that what you mean?

keradus: actually hashMask is calculated without usage of input param here`$object`. maybe extract it to separated function and then 

```php
public static function spl_object_id($object)
{
    return self::getHashMask() ^ hexdec(substr(spl_object_hash($object), self::$hashOffset, PHP_INT_SIZE));
}
```
?

nicolas-grekas: I prefer raw perf, thus prevent an extra call here

nicolas-grekas: here we are, init logic split in separate function, called only once

tests/Php72/Php72Test.php
```diff
@@ -48,4 +48,18 @@ public function testPhpOsFamily()
...
           $this->assertSame(PHP_OS_FAMILY, p::php_os_family());
     }
+
+    /**
+     * @covers Symfony\Polyfill\Php72\Php72::spl_object_id
+     */
+    public function testSplObjectId()
+    {
+        $obj = new \stdClass();
+        $id = spl_object_id($obj);
```
keradus: please test that calling `spl_object_id` for non-object works the same as native function

stof:
@nicolas-grekas I suppose that VarDumper can then be migrated to use the new `spl_object_id` function, right ?

stof:
note: merging this PR must wait until the PR is merged in PHP itself. We don't want to merge the polyfill if they reject the PR.

src/Php72/Php72.php
```diff
@@ -80,4 +83,36 @@ public static function php_os_family()
...
+            self::initHashMask();
+        }
+
+        return self::$hashMask ^ hexdec(substr(spl_object_hash($object), self::$hashOffset, PHP_INT_SIZE));
+    }
+
+    private static function initHashMask()
+    {
+        $obj = (object) array();
+        self::$hashOffset = 16 - PHP_INT_SIZE;
```
stof: if you want to improve perf even further, you could even fully qualify these constants (as done for `\PHP_VERSION_ID`). Not sure it is worth it though (this code is called only once)

src/Php72/Php72.php
lyrixx: :+1:

tests/Php72/Php72Test.php
nicolas-grekas: done

src/Php72/Php72.php
nicolas-grekas: done by removing $hashOffset

nicolas-grekas:
@stof VarDumper can use this polyfill only in master, because of HHVM compat...

robfrawley:
The `spl_object_id()` function now merged upstream in https://github.com/php/php-src/commit/5097e2ee13de12b4445b4123e1554c0733c6853c (also see https://github.com/php/php-src/pull/2611#issuecomment-319717081) so this can now be safely merged here, as well.

fabpot:
Thank you @nicolas-grekas.

robfrawley:
@fabpot Sorry to comment on a closed PR, but I don't think this is worth a new issue: it would be great to get a release tagged that includes this PR (I hate having to include `dev-master` in my composer files). :-)

fabpot:
@robfrawley Done, 1.5 is out now.

PR #99, opened 2017-08-29
enhance readme

gharlan:


nicolas-grekas:
Good catch, thanks @gharlan.

Issue #101, opened 2017-09-11
Throwable interface in polyfill-php70


hlecorche:
Hello,

There is the `Error` class in polyfill-php70. But why isn't there any `Throwable` interface ?

http://php.net/manual/fr/language.errors.php7.php

Thanks

keradus:
as you cannot mock behaviour of `Throwable`.
even if you would create interface, this:
`new Exception() instanceof Throwable` will not work

for that, if you cannot implement the behaviour, adding implement would be wrong

nicolas-grekas:
Closing as explained.

PR #102, opened 2017-10-11
Add SessionUpdateTimestampHandlerInterface polyfill

nicolas-grekas:
This interface is mostly undocumented, but is required to implement lazy writes and session fixation prevention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-* v1.5.1
+* v1.6.0
 
+ * add `SessionUpdateTimestampHandlerInterface` in PHP 7.0 polyfill
  * fixed loading of Apcu polyfill when Zend Server's Data Cache is used
 
 * v1.5.0
--- a/src/Php70/README.md
+++ b/src/Php70/README.md
@@ -1,16 +1,15 @@
 Symfony Polyfill / Php70
 ========================
 
-This component provides functions unavailable in releases prior to PHP 7.0:
+This component provides features unavailable in releases prior to PHP 7.0:
 
 - [`intdiv`](http://php.net/intdiv)
 - [`preg_replace_callback_array`](http://php.net/preg_replace_callback_array)
 - [`error_clear_last`](http://php.net/error_clear_last)
 - `random_bytes` and `random_int` (from [paragonie/random_compat](https://github.com/paragonie/random_compat))
 - [`*Error` throwable classes](http://php.net/Error)
-
-Also, it provides a constant added to PHP 7.0:
 - [`PHP_INT_MIN`](http://php.net/manual/en/reserved.constants.php#constant.php-int-min)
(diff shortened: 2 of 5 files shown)


src/Php70/README.md
```diff
@@ -1,16 +1,15 @@
...
 
 - [`intdiv`](http://php.net/intdiv)
 - [`preg_replace_callback_array`](http://php.net/preg_replace_callback_array)
 - [`error_clear_last`](http://php.net/error_clear_last)
 - `random_bytes` and `random_int` (from [paragonie/random_compat](https://github.com/paragonie/random_compat))
 - [`*Error` throwable classes](http://php.net/Error)
-
-Also, it provides a constant added to PHP 7.0:
 - [`PHP_INT_MIN`](http://php.net/manual/en/reserved.constants.php#constant.php-int-min)
+- `SessionUpdateTimestampHandlerInterface`
```
keradus: everything in the list except this bullet is linkable

src/Php70/Resources/stubs/SessionUpdateTimestampHandlerInterface.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+interface SessionUpdateTimestampHandlerInterface
```
keradus: I see it somehow similar: https://github.com/symfony/polyfill/issues/101#issuecomment-328663214
but probably only things that were not available in 5.x are using this interface, so it shall be safe.
correct?

nicolas-grekas: it's safe, this interface is for userland, it's not implemented by core classes.

src/Php70/README.md
nicolas-grekas: I don't have any target for this, so I cannot add a link.

keradus: damn. then it would be nice to document types of it. or it's `mixed` x3 ?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+* v1.6.0
```
nicolas-grekas: Can replace the 1.5.1 item if we decide to release only 1.6.0, but not 1.5.1 before. Please advise.

src/Php70/README.md
nicolas-grekas: docblocks added

CHANGELOG.md
keradus: I suggest to do bugfixes in patches and features in minors ;)

stof: well, if a single release does both bugfixes and features, it is fine. Forcing to split into 2 releases just to have only features in 1.6.0 does not make sense.

keradus: ah, only one line is actively supported here

fabpot: 1.6.0 looks good to me.

PR #103, opened 2017-10-11
[Apcu] Add missing PSR-4 entry

### LINKED ISSUE #98: "Class not found" exception when using polyfill-apcu together with Zend Server
We were happy #94 was closed and 1.5.0 released.
However, the additional file `Apcu.php` is neither in the `autoload.files` property in `composer.json` nor is it required in `bootstrap.php`.

Maybe its best to simply `require` the file in
https://github.com/symfony/polyfill/blob/master/src/Apcu/bootstrap.php#L20
so it is not loaded when Zend Server Data Cache is not enabled?

nicolas-grekas:
Fix #98

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 * v1.5.1
 
- * n/a
+ * fixed loading of Apcu polyfill when Zend Server's Data Cache is used
 
 * v1.5.0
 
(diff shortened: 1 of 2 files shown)


keradus:
please restart the build to confirm 5.3 works ;)

PR #104, opened 2017-10-11
Use the old precise environment for PHP 5.3 jobs

stof:
Travis does not provide PHP 5.3 anymore on Trusty.


stof:
The precise environment will be deleted next year, but this allows us to keep testing the polyfill on PHP 5.3 in the meantime.

nicolas-grekas:
Thank you @stof.

PR #106, opened 2017-12-02
Move to namespaced phpunit versions

DavidPrevot:
These changes should be needed when you decide to test the code with PHPUnit 6.

Bug-Debian: https://bugs.debian.org/882930

### DIFF
--- a/src/Util/TestListener.php
+++ b/src/Util/TestListener.php
@@ -11,15 +11,22 @@
 
 namespace Symfony\Polyfill\Util;
 
+use PHPUnit\Framework\AssertionFailedError;
+use PHPUnit\Framework\Warning;
+use PHPUnit\Framework\SkippedTestError;
+use PHPUnit\Framework\Test;
+use PHPUnit\Framework\TestListener as PHPUnit_Framework_TestListener;
+use PHPUnit\Framework\TestSuite;
+
 /**
  * @author Nicolas Grekas <p@tchwork.com>
  */
-class TestListener extends \PHPUnit_Framework_TestSuite implements \PHPUnit_Framework_TestListener
+class TestListener extends TestSuite implements PHPUnit_Framework_TestListener
 {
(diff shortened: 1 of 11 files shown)


carusogabriel:
Some tests were declared wrong.

keradus:
what about php 5.3.3 ?

src/Util/TestListener.php
```diff
@@ -11,15 +11,22 @@
...
+use PHPUnit\Framework\SkippedTestError;
+use PHPUnit\Framework\Test;
+use PHPUnit\Framework\TestListener as PHPUnit_Framework_TestListener;
+use PHPUnit\Framework\TestSuite;
+
 /**
  * @author Nicolas Grekas <p@tchwork.com>
  */
-class TestListener extends \PHPUnit_Framework_TestSuite implements \PHPUnit_Framework_TestListener
+class TestListener extends TestSuite implements PHPUnit_Framework_TestListener
```
stof: This test listener is now totally incompatible with older versions of PHPUnit. This is not good (Symfony itself relies on support for older versions due to PHP support).

tests/Iconv/IconvTest.php
```diff
@@ -110,7 +111,8 @@ public function testIconvMimeDecode()
     public function testIconvMimeDecodeIllegal()
     {
         iconv_mime_decode('Legal encoded-word: =?utf-8?Q?*?= .');
-        $this->setExpectedException('PHPUnit_Framework_Error_Notice', 'Detected an illegal character in input string');
+        $this->expectException('PHPUnit\Framework\Error\Notice');
```
stof: this is not compatible with PHPUnit 4.8, which is needed to test the package on PHP 5.3 to 5.5 (which must remain supported for us)

nicolas-grekas:
Thanks you @DavidPrevot for raising the topic and proposing a patch. I took over in #112

PR #110, opened 2017-12-17
Add missing function to phpdoc

flack:
This one is implemented, but wasn't present in the list


fabpot:
Good catch, thanks @flack.

PR #111, opened 2017-12-28
Add polyfills for mb_*_numericentity

RGustBardon:


src/Mbstring/Mbstring.php
```diff
@@ -138,6 +138,128 @@ public static function mb_encode_mimeheader($s, $charset = null, $transferEncodi
...
+        if (!is_array($convmap) || !$convmap) {
+            return false;
+        }
+
+        if (null !== $encoding && !is_scalar($encoding)) {
+            trigger_error('mb_decode_numericentity() expects parameter 3 to be string, '.gettype($s).' given', E_USER_WARNING);
+            return '';  // Instead of null (cf. mb_encode_numericentity).
+        }
+
+        if ('' === $s .= '') {
```
keradus: please don't cast to string like that

src/Mbstring/Mbstring.php
```diff
@@ -138,6 +138,128 @@ public static function mb_encode_mimeheader($s, $charset = null, $transferEncodi
...
+
+        $cnt = floor(count($convmap) / 4) * 4;
+
+        for ($i = 0; $i < $cnt; $i += 4) {
+            // collector_decode_htmlnumericentity ignores $convmap[$i + 3]
+            $convmap[$i] += $convmap[$i + 2];
+            $convmap[$i + 1] += $convmap[$i + 2];
+        }
+
+        $s = preg_replace_callback('/&#(?:0*([0-9]+)|x0*([0-9a-fA-F]+))(?!&);?/', function ($m) use ($cnt, $convmap) {
```
keradus: `m` could be typehinted

src/Mbstring/Mbstring.php
RGustBardon: @keradus: I understand that some might not find it readable; however, I have tried to imitate the style of the project (see `mb_convert_case`). Thank you for bringing this up!

src/Mbstring/Mbstring.php
RGustBardon: @keradus: I agree! That being said, the file does not use type declarations in similar situations (cf. `html_encoding_callback`). Thank you for your kind suggestion!

RGustBardon:
@nicolas-grekas: Thank you for the opportunity to contribute to this amazing project!

src/Mbstring/Mbstring.php
keradus: the fact that somebody, somewhere, forgot this, is not an excuse to not to it in new code.

RGustBardon: @keradus: Point taken! I have acted accordingly.

keradus: looks lovely, thanks ;)

nicolas-grekas:
Thank you @RGustBardon.

PR #112, opened 2018-01-31
Add compatibility with namespaced PHPUnit

nicolas-grekas:
Replaces #106, fixes Bug-Debian: https://bugs.debian.org/882930

### DIFF
--- a/tests/Iconv/IconvTest.php
+++ b/tests/Iconv/IconvTest.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Polyfill\Tests\Iconv;
 
+use PHPUnit\Framework\TestCase;
 use Symfony\Polyfill\Iconv\Iconv as p;
 
 /**
@@ -18,7 +19,7 @@ use Symfony\Polyfill\Iconv\Iconv as p;
  *
  * @covers Symfony\Polyfill\Iconv\Iconv::<!public>
  */
-class IconvTest extends \PHPUnit_Framework_TestCase
+class IconvTest extends TestCase
 {
     /**
      * @covers Symfony\Polyfill\Iconv\Iconv::iconv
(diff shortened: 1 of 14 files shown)


tests/Iconv/IconvTest.php
```diff
@@ -106,11 +107,12 @@ public function testIconvMimeDecode()
 
     /**
      * @covers Symfony\Polyfill\Iconv\Iconv::iconv_mime_decode
+     * @expectedException \PHPUnit\Framework\Error\Notice
+     * @expectedExceptionMessage Detected an illegal character in input string
      */
     public function testIconvMimeDecodeIllegal()
     {
         iconv_mime_decode('Legal encoded-word: =?utf-8?Q?*?= .');
-        $this->setExpectedException('PHPUnit_Framework_Error_Notice', 'Detected an illegal character in input string');
```
stof: this test should be split in 2, so that the exception is allowed only in the second call, not the first one.

stof: or it should use a method_exists check to keep configuring it inline

tests/Iconv/IconvTest.php
```diff
@@ -106,11 +107,12 @@ public function testIconvMimeDecode()
 
     /**
      * @covers Symfony\Polyfill\Iconv\Iconv::iconv_mime_decode
+     * @expectedException \PHPUnit\Framework\Error\Notice
```
stof: does this work even on older versions where the class name had underscores ?

tests/Intl/Grapheme/GraphemeTest.php
```diff
@@ -11,22 +11,24 @@
...
+class GraphemeTest extends TestCase
 {
     /**
      * @covers Symfony\Polyfill\Intl\Grapheme\Grapheme::grapheme_extract
+     * @expectedException \PHPUnit\Framework\Error\Warning
+     * @expectedExceptionMessage expects parameter 1 to be string, array given
      */
     public function testGraphemeExtractArrayError()
     {
         grapheme_extract('', 0);
```
stof: why do we have this call here ?

nicolas-grekas:
Now green

PR #115, opened 2018-03-02
Add is_countable under PHP 7.3

carusogabriel:
As requested by @nicolas-grekas :)

Waiting for https://github.com/php/php-src/pull/3026

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "symfony/polyfill-php70": "self.version",
         "symfony/polyfill-php71": "self.version",
         "symfony/polyfill-php72": "self.version",
+        "symfony/polyfill-php73": "self.version",
         "symfony/polyfill-iconv": "self.version",
         "symfony/polyfill-intl-grapheme": "self.version",
         "symfony/polyfill-intl-icu": "self.version",
@@ -50,6 +51,7 @@
             "src/Php70/bootstrap.php",
             "src/Php71/bootstrap.php",
             "src/Php72/bootstrap.php",
+            "src/Php73/bootstrap.php",
             "src/Iconv/bootstrap.php",
             "src/Intl/Grapheme/bootstrap.php",
             "src/Intl/Icu/bootstrap.php",
@@ -65,7 +67,7 @@
     "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
-            "dev-master": "1.7-dev"
+            "dev-master": "1.8-dev"
         }
     }
 }
--- a/src/Php73/LICENSE
+++ b/src/Php73/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2018 Fabien Potencier
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is furnished
+to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
--- a/src/Php73/bootstrap.php
+++ b/src/Php73/bootstrap.php
@@ -1,0 +1,18 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+if (PHP_VERSION_ID < 70300) {
+    if (!function_exists('is_countable')) {
+        function is_countable($var) {
+            return is_array($var) || $var instanceof Countable;
+        }
+    }
+}
--- a/src/Php73/composer.json
+++ b/src/Php73/composer.json
@@ -1,8 +1,8 @@
 {
-    "name": "symfony/polyfill-util",
+    "name": "symfony/polyfill-php73",
     "type": "library",
-    "description": "Symfony utilities for portability of PHP codes",
-    "keywords": ["polyfill", "shim", "compat", "compatibility"],
+    "description": "Symfony polyfill backporting some PHP 7.3+ features to lower PHP versions",
+    "keywords": ["polyfill", "shim", "compatibility", "portable"],
     "homepage": "https://symfony.com",
     "license": "MIT",
     "authors": [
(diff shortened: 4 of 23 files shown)


src/Php73/bootstrap.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Php73 as p;
+
+if (PHP_VERSION_ID < 70300) {
+    if (!function_exists('is_countable')) {
+        function is_countable($s) { return p\Php73::is_countable($s); }
```
nicolas-grekas: for such a simple implementation, I think we can inline the code a remove the static class (as done on other polyfills AFAIK)

tests/Php73/Php72Test.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Symfony\Polyfill\Tests\Php73;
+
+use PHPUnit\Framework\TestCase;
+use Symfony\Polyfill\Php73\Php73 as p;
+
+/**
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ *
+ * @covers Symfony\Polyfill\Php73\Php73::<!public>
```
nicolas-grekas: should be removed in you accept the previous comment (same below)

src/Php73/bootstrap.php
carusogabriel: @nicolas-grekas Sure, gonna do this!

tests/Php73/Php72Test.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * @covers Symfony\Polyfill\Php73\Php73::is_countable
+     */
+    public function testIsCountable()
+    {
+        $this->assertTrue(is_countable([1, 2, 3]));
+        $this->assertTrue(is_countable(new \ArrayIterator(['foo', 'bar', 'baz'])));
+        $this->assertTrue(is_countable(new \ArrayIterator()));
+        $this->assertFalse(is_countable(new \stdClass()));
```
keradus: please add a case with generator

src/Php73/composer.json
```diff
@@ -0,0 +1,31 @@
...
+        "php": ">=5.3.3"
+    },
+    "autoload": {
+        "psr-4": { "Symfony\\Polyfill\\Php73\\": "" },
+        "files": [ "bootstrap.php" ]
+    },
+    "minimum-stability": "dev",
+    "extra": {
+        "branch-alias": {
+            "dev-master": "1.7-dev"
```
stof: should be 1.8 (and we should bump the version in master before merging this), as this is a new feature and so will be a new minor version, not a new patch version

tests/Php73/Php72Test.php
carusogabriel: @keradus Is this possible with 5.x tests?

src/Php73/composer.json
carusogabriel: @stof I'll just wait for the official merge in php-src, I'll do it :)

tests/Php73/Php72Test.php
keradus: even if not, it shall still be tested

carusogabriel:
@nicolas-grekas Should be ready to go :smiley:

tests/Php73/Php73Test.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $this->assertFalse(is_countable($this->genOneToTen()));
+    }
+
+    private function genOneToTen()
+    {
+        for ($i = 1; $i <= 10; ++$i) {
+            yield $i;
+        }
+    }
```
carusogabriel: @keradus Is there a way to add generators tests without breaking the build?

tests/Php73/Php73Test.php
```diff
@@ -0,0 +1,47 @@
...
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ */
+class Php73Test extends TestCase
+{
+    /**
+     * @covers Symfony\Polyfill\Php73\Php73::is_countable
+     */
+    public function testIsCountable()
+    {
+        $this->assertTrue(is_countable(array(1, 2, '3')));
```
carusogabriel: @keradus Fixed the array syntax, thanks.

tests/Php73/Php73Test.php
keradus: the `yield` code would need to be extracted to dedicated file and loaded conditionally

carusogabriel: @keradus Done, thanks.

keradus:
looks better with every change @carusogabriel !
(btw, if you would not squash, it would be way easier to review the changes, now one need to read whole diff again and again, instead of just patch - don't worry, squash can be done during merging!)

one last thing - could you extend `.travis.yml` matrix to contain 7.2 and 7.3 ?

carusogabriel:
@keradus 
> btw, if you would not squash, it would be way easier to review the changes, now one need to read whole diff again and again, instead of just patch - don't worry, squash can be done during merging!

Sure, thanks for the tip, I always amend my commits, but I'll start to commit and then squash them before merging the PR, thanks!

> could you extend .travis.yml matrix to contain 7.2 and 7.3 ?

Sure, pushing this changes ASAP.

carusogabriel:
@keradus There's one remain error in PHP 7.3 that I couldn't figure out how to solve. Maybe we should just add 7.2?

keradus:
the feature you are adding has to be tested under 7.3. if not, then we test only "does our polyfill work like we expect", but don't "does our polyfill expectation are same as original implementation".
yet, I agree 7.3 failing is not related to your PR.
sth nice to have for sure, but maybe indeed out of the scope here. nevertheless, adding it (even temporary) to travis matrix has proven that your changes works same as native implementation.

if you ask me, I would be fine with reverting travis changes and do it as separated PR.

keradus:
@stof , your request has been already fulfilled, could you please revisit your review ?

In general, missing thing on Travis is related to `Mbstring`, thus it's not connected with feature introduced by this PR.
Could we make this one released?

composer.json
```diff
@@ -50,6 +51,7 @@
             "src/Php70/bootstrap.php",
             "src/Php71/bootstrap.php",
             "src/Php72/bootstrap.php",
+            "src/Php73/bootstrap.php",
```
nicolas-grekas: the branch-alias should be updated to 1.8-dev for all composer.json files of the repo

src/Php73/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2015-2018 Fabien Potencier
```
nicolas-grekas: 2018 only (not "2015-")

src/Php73/bootstrap.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+if (PHP_VERSION_ID < 70300) {
+    if (!function_exists('is_countable')) {
+        function is_countable($s) {
```
nicolas-grekas: the function should be on one single line (to make the test suite work as expected)
`$s` should be renamed by `$var` I suppose also.

nicolas-grekas:
Would you mind adding `hrtime()` also in the PHP 7.3 polyfill (based on `microtime()`)?
```
  . Added monotonic timer function hrtime([bool get_as_num]). It returns an
    array of the form [seconds, nanoseconds] with the timestamp starting at
    an unspecified point in the past. If the optional argument is passed as
    true, the return value is an integer on 64-bit systems or float on
    32-bit systems, representing the nanoseconds. The timestamp is not
    adjustable and is not related to wall clock or time of day. The timers
    are available under Linux, FreeBSD, Windows, Mac, SunOS, AIX and their
    derivatives. If no required timers are provided by a corresponding
    platform, the function returns false.
```

keradus:
Could we please add different feature with different PR?
This one is already awaiting for 2.5 months, no need to postpone it even further ;)

nicolas-grekas:
Thank you @carusogabriel.

nicolas-grekas:
FYI, I removed the nightly line because it failed for unrelated reason, to be fixed in a next PR. hrtime() welcome now.

Majkl578:
As noted in the commit here: https://github.com/symfony/polyfill-php73/commit/f7139426b2a7fce4d4440fde55284fe47a1b18ca#r28736995
This implementation is incorrect.

PR #116, opened 2018-04-06
Php70: do not expose intArg

glensc:
`Php70::intArg` is used from `Php70` class only. make it private.

project search result: https://github.com/symfony/polyfill/search?q=intArg&type=Code


keradus:
a little :+1:, but still :+1:
it's internal class, so no big deal here

nicolas-grekas:
Thank you @glensc.

PR #118, opened 2018-04-25
Add the ctype polyfill

BackEndTea:
This PR adds a ctype pollyfill as mentioned in https://github.com/symfony/symfony/issues/24168

A copy of https://github.com/BackEndTea/ctype-compat adapted to the standards of this project.

ping @nicolas-grekas

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,6 +46,7 @@
         "psr-4": { "Symfony\\Polyfill\\": "src/" },
         "files": [
             "src/Apcu/bootstrap.php",
+            "src/Ctype/bootstrap.php",
             "src/Php54/bootstrap.php",
             "src/Php55/bootstrap.php",
             "src/Php56/bootstrap.php",
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
     },
     "replace": {
         "symfony/polyfill-apcu": "self.version",
+        "symfony/polyfill-ctype": "self.version",
         "symfony/polyfill-php54": "self.version",
         "symfony/polyfill-php55": "self.version",
         "symfony/polyfill-php56": "self.version",
--- a/src/Ctype/Ctype.php
+++ b/src/Ctype/Ctype.php
@@ -1,0 +1,246 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Ctype;
+
+/**
+ * Ctype implementation through regex.
+ *
+ * @author Gert de Pagter<BackEndTea@gmail.com
+ */
+final class Ctype
+{
+    /**
+     * Returns TRUE if every character in text is either a letter or a digit, FALSE otherwise.
+     *
+     * @see http://php.net/manual/en/function.ctype-alnum.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_alnum($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^A-Za-z0-9]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in text is a letter, FALSE otherwise.
+     *
+     * @see http://php.net/manual/en/function.ctype-alpha.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_alpha($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^A-Za-z]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in text is a control character from the current locale, FALSE otherwise.
+     *
+     * @see http://php.net/manual/en/function.ctype-cntrl.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_cntrl($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^\x00-\x1f\x7f]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in the string text is a decimal digit, FALSE otherwise.
+     *
+     * @see http://php.net/manual/en/function.ctype-digit.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_digit($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^0-9]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in text is printable and actually creates visible output (no white space), FALSE otherwise.
+     *
+     * @see http://php.net/manual/en/function.ctype-graph.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_graph($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^!-~]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in text is a lowercase letter.
+     *
+     * @see http://php.net/manual/en/function.ctype-lower.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_lower($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^a-z]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in text will actually create output (including blanks). Returns FALSE if text contains control characters or characters that do not have any output or control function at all.
+     *
+     * @see http://php.net/manual/en/function.ctype-print.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_print($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^ -~]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in text is printable, but neither letter, digit or blank, FALSE otherwise.
+     *
+     * @see http://php.net/manual/en/function.ctype-punct.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_punct($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^!-\/\:-@\[-`\{-~]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in text creates some sort of white space, FALSE otherwise. Besides the blank character this also includes tab, vertical tab, line feed, carriage return and form feed characters.
+     *
+     * @see http://php.net/manual/en/function.ctype-space.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_space($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^\s]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in text is an uppercase letter.
+     *
+     * @see http://php.net/manual/en/function.ctype-upper.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     *
+     */
+    public static function ctype_upper($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^A-Z]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Returns TRUE if every character in text is a hexadecimal 'digit', that is a decimal digit or a character from [A-Fa-f] , FALSE otherwise.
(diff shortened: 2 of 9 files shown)


composer.json
```diff
@@ -26,6 +26,7 @@
     },
     "replace": {
         "symfony/polyfill-apcu": "self.version",
+        "symfony/polyfill-ctype": "self.version",
```
nicolas-grekas: missing entry in autoload.files

src/Ctype/composer.json
```diff
@@ -0,0 +1,31 @@
...
+    "php": ">=5.3.3"
+  },
+  "autoload": {
+    "psr-4": { "Symfony\\Polyfill\\Ctype\\": "" },
+    "files": [ "bootstrap.php" ]
+  },
+  "minimum-stability": "dev",
+  "extra": {
+    "branch-alias": {
+      "dev-master": "1.7-dev"
```
nicolas-grekas: 1.8-dev

src/Ctype/README.md
```diff
@@ -0,0 +1,12 @@
+Symfony Polyfill / APCu
```
nicolas-grekas: the root README needs an update also

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,246 @@
...
+     *
+     * @internal
+     *
+     * @param string|int $int
+     *
+     * @return string
+     */
+    private static function convert_int_to_char_for_ctype($int)
+    {
+        if(!is_integer($int)) {
```
nicolas-grekas: missing space, and should be `\is_int(...)`

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,246 @@
...
+            return $int;
+        }
+
+        if ($int < -128 || $int > 255) {
+            return (string) $int;
+        }
+
+        if ($int < 0) {
+            $int += 256;
```
nicolas-grekas: what happens when $int < 256?

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        return chr($int);
+    }
+
```
nicolas-grekas: extra line

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,246 @@
...
+        if ($int < -128 || $int > 255) {
+            return (string) $int;
+        }
+
+        if ($int < 0) {
+            $int += 256;
+        }
+
+        return chr($int);
```
nicolas-grekas: should have a backslash `\chr(...)`
(see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048 for some background on this)

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,246 @@
...
+            !preg_match(
+                '/[^A-Fa-f0-9]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
+    }
+
+    /**
+     * Converts integers to their char versions according to normal ctype behaviour, if needed.
+     *
+     * @internal
```
nicolas-grekas: should be moved at the class level

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,246 @@
...
+     *
+     * @return bool
+     */
+    public static function ctype_alnum($text)
+    {
+        return !empty($text) &&
+            !preg_match(
+                '/[^A-Za-z0-9]/',
+                self::convert_int_to_char_for_ctype($text)
+            );
```
nicolas-grekas: we'd prefer this on a single line (same in all methods)

src/Ctype/composer.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "symfony/polyfill-ctype",
+  "type": "library",
+  "description": "Symfony polyfill for ctype_ functions",
+  "keywords": ["polyfill", "compatibility", "portable", "ctype"],
+  "homepage": "https://symfony.com",
+  "license": "MIT",
+  "authors": [
+    {
+      "name": "Nicolas Grekas",
```
fabpot: Should be you instead :)

src/Ctype/Ctype.php
fabpot: I know this is minor, but for consistency, could you change all `preg_match()` calls to be one-liners instead?

src/Ctype/Ctype.php
fabpot: extra blank line should be removed.

src/Ctype/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2015-2018 Fabien Potencier
```
fabpot: 2018 is enough

composer.json
BackEndTea: Fixed it

BackEndTea:
>Of course this needs test cases (see how they are implemented on other polyfills)

It seems i forgot to add the test to my initial commit, i added them now.

I'm not sure what i causing the travis/appveyor errors. I'm having trouble running tests locally as well.

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,206 @@
...
+     *
+     * @see http://php.net/manual/en/function.ctype-alnum.php
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_alnum($text)
+    {
+        return !'' === $text && !preg_match('/[^A-Za-z0-9]/', self::convert_int_to_char_for_ctype($text));
```
nicolas-grekas: `'' !== $text` (same below)

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,206 @@
...
+     * (negative values have 256 added in order to allow characters in the Extended ASCII range).
+     * Any other integer is interpreted as a string containing the decimal digits of the integer.
+     *
+     * @param string|int $int
+     *
+     * @return string
+     */
+    private static function convert_int_to_char_for_ctype($int)
+    {
+        if(! \is_integer($int)) {
```
nicolas-grekas: `if(!\is_int($int)) {`
actually, I wonder how native PHP behaves when passed with a non compatibe type (an array, a resource, etc.)
could be worth adding a few tests for this and mimic the behavior

src/Ctype/Ctype.php
nicolas-grekas: I'd suggest to add a few tests for these, with eg -127, and -384

src/Ctype/composer.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "symfony/polyfill-ctype",
+  "type": "library",
+  "description": "Symfony polyfill for ctype_ functions",
```
nicolas-grekas: I'd suggest to remove the underscore: ctype

tests/Ctype/CtypeTest.php
```diff
@@ -0,0 +1,521 @@
...
+     */
+    public function testInvalidCtypeAlnum($text)
+    {
+        $this->assertFalse(ctype_alnum($text));
+    }
+
+    public function provideInvalidAlnum()
+    {
+        return array (
+            array (8),
```
nicolas-grekas: extra space the two lines above: `array(` (similar below)

tests/Ctype/CtypeTest.php
```diff
@@ -0,0 +1,521 @@
...
+    {
+        $this->assertTrue(ctype_lower($text));
+    }
+
+    public function provideValidLowers()
+    {
+        return array(
+            array(98),
+            array('asdf'),
+            array('stuff')
```
nicolas-grekas: missing trailing comma (same below)

src/Ctype/Ctype.php
BackEndTea: From what ive found it will simply return false, and not trigger any error, which is not preg_match result

BackEndTea:
Running tests gives the following error:
```
which: no composer.phar in (/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/home/gpagter/bin)
PHP Fatal error:  Uncaught Error: Class 'Symfony\Bridge\PhpUnit\Legacy\TestRunner' not found in /home/gpagter/Projects/polyfill/vendor/symfony/phpunit-bridge/Legacy/CommandForV5.php:26
Stack trace:
#0 /home/gpagter/Projects/polyfill/vendor/bin/.phpunit/phpunit-5.7/src/TextUI/Command.php(129): Symfony\Bridge\PhpUnit\Legacy\CommandForV5->createRunner()
#1 /home/gpagter/Projects/polyfill/vendor/bin/.phpunit/phpunit-5.7/src/TextUI/Command.php(116): PHPUnit_TextUI_Command->run(Array, true)
#2 /home/gpagter/Projects/polyfill/vendor/bin/.phpunit/phpunit-5.7/phpunit(17): PHPUnit_TextUI_Command::main()
#3 /home/gpagter/Projects/polyfill/vendor/symfony/phpunit-bridge/bin/simple-phpunit(251): include('/home/gpagter/P...')
#4 {main}
  thrown in /home/gpagter/Projects/polyfill/vendor/symfony/phpunit-bridge/Legacy/CommandForV5.php on line 26

```

src/Ctype/composer.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "symfony/polyfill-ctype",
```
stof: please use 4 spaces for the indentation, as done in other JSON files

stof:
@BackEndTea do you have composer available globally in your PATH ?

BackEndTea:
@stof Yeah, but its named composer not composer.phar

BackEndTea:
I now added test cases for throwing in things that aren't ints or strings.

src/Ctype/Ctype.php
BackEndTea: Done

src/Ctype/bootstrap.php
```diff
@@ -0,0 +1,26 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Ctype as p;
+
+if (!extension_loaded('ctype')) {
+    function ctype_alnum($text) { return p\Ctype::ctype_alnum($text); }
```
BackEndTea: Do we want individual function_exists/ is_callable checks for each function?
In case they have their own polyfills somewhere?

nicolas-grekas: good question :)
in other polyfills, we don't use extension_loaded but function exists, but we do only one check
so the above if should be eg `if (!function_exists('ctype_alnum')) {`
we'll consider that if ppl have one, they have them all.
If not, issue is on their side

BackEndTea: :+1: I'll go ahead and change that

BackEndTea:
Extracting a few to one comment:

* all strings are evaluated normally
* integers between -128 and 255 are evaluated as their charcter on the ascii table, where negative numbers have 256 added to them
* all other integers are interpeted as strings, so -250 becomes '-250' etc
* anything that isn't a string returns false, even if their value casted to a string would return true
* no error is ever triggered, even when arrays or objects are thrown into the mix

nicolas-grekas:
Can you rebase, then update the CHANGELOG file please? I just merged a few PRs preparing v1.8

BackEndTea:
Rebased and updated the changelog

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,228 @@
...
+ * @internal
+ *
+ * @author Gert de Pagter <BackEndTea@gmail.com>
+ */
+final class Ctype
+{
+    /**
+     * Returns TRUE if every character in text is either a letter or a digit, FALSE otherwise.
+     *
+     * @see http://php.net/manual/en/function.ctype-alnum.php
```
nicolas-grekas: let's use short urls (same below)
http://php.net/ctype_alnum

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,228 @@
...
+     *
+     * @param string|int $text
+     *
+     * @return bool
+     */
+    public static function ctype_alnum($text)
+    {
+        $text = self::convert_int_to_char_for_ctype($text);
+
+        return is_string($text) && '' !== $text && !preg_match('/[^A-Za-z0-9]/', $text);
```
nicolas-grekas: \is_string

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,228 @@
...
+     * (negative values have 256 added in order to allow characters in the Extended ASCII range).
+     * Any other integer is interpreted as a string containing the decimal digits of the integer.
+     *
+     * @param string|int $int
+     *
+     * @return string
+     */
+    private static function convert_int_to_char_for_ctype($int)
+    {
+        if(! \is_int($int)) {
```
nicolas-grekas:
```diff
-if(! \is_int($int)) {
+if (!\is_int($int)) {
```

fabpot:
Thank you @BackEndTea, great work!

BackEndTea:
Thanks everyone, for the feedback during the PR and the merge.

src/Ctype/Ctype.php
```diff
@@ -0,0 +1,227 @@
...
+     * Converts integers to their char versions according to normal ctype behaviour, if needed.
+     *
+     * If an integer between -128 and 255 inclusive is provided,
+     * it is interpreted as the ASCII value of a single character
+     * (negative values have 256 added in order to allow characters in the Extended ASCII range).
+     * Any other integer is interpreted as a string containing the decimal digits of the integer.
+     *
+     * @param string|int $int
+     *
+     * @return string
```
stof: this does not always return a string, as the polyfill silently accepts invalid types (see your tests)

BackEndTea: My bad the return type should be `mixed`.

And the silently accepting invalid types is intended since that is default behavior

PR #119, opened 2018-04-25
Update is_countable to include countable objects that doesn't implement \Countable

pierredup:
Some objects are countable but don't implement `Countable` ([symfony/polyfill#115#issuecomment-384361414](https://github.com/symfony/polyfill/pull/115#issuecomment-384361414))

The objects that I could identify that is countable is `ResourceBundle` and `SimpleXmlElement`


carusogabriel:
@pierredup Is there any way to find all of then?

pierredup:
I used [https://github.com/php/php-src/search?utf8=✓&q=count_elements&type=](https://github.com/php/php-src/search?utf8=✓&q=count_elements&type=) as a simple check, which should cover everything in core. The other items found in the result (SplHeap, DoublyLinkedList, SplFixedArray etc) already implement Countable

tests/Php73/Php73Test.php
```diff
@@ -26,6 +26,8 @@ public function testIsCountable()
         $this->assertTrue(is_countable(array(1, 2, '3')));
         $this->assertTrue(is_countable(new \ArrayIterator(array('foo', 'bar', 'baz'))));
         $this->assertTrue(is_countable(new \ArrayIterator()));
+        $this->assertTrue(is_countable(new \SimpleXMLElement('<foo><bar/><bar/><bar/></foo>')));
+        $this->assertTrue(is_countable(\ResourceBundle::create('en', __DIR__.'/fixtures')));
```
Ayesh: Don't we need `class_exists()` calls for these two to make sure `intl` and `xml` extensions are loaded.

pierredup: In Travis the extensions are loaded so the classes should always exist. In the actual polyfill, the check would just return `false` if the extension isn't loaded and the class doesn't exist

PR #123, opened 2018-04-30
Update Ctype

BackEndTea:
The function would only return a string if provided with a string or an int.


fabpot:
Thank you @BackEndTea.

PR #125, opened 2018-05-06
PHP 7.3: Add hrtime() function

### LINKED ISSUE #122: Polyfill hrtime()
missing in the PHP 7.3 polyfill

@carusogabriel maybe? :)

IonBazan:
Fixes #122 
This PR adds `hrtime()` function introduced in PHP 7.3. This implementation is based on `microtime` and is far from being precise. However, results should remain similar to actual `hrtime()` implementation.

### DIFF
--- a/src/Php73/Php73.php
+++ b/src/Php73/Php73.php
@@ -3,12 +3,20 @@
 namespace Symfony\Polyfill\Php73;
 
 /**
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
  * @internal
  */
 final class Php73
 {
     const NANO_IN_SEC = 1000000000.0;
 
+    /**
+     * @param mixed $var
+     *
+     * @return bool
+     */
     public static function is_countable($var)
     {
         return is_array($var)
@@ -17,6 +25,11 @@ final class Php73
             || $var instanceof \SimpleXmlElement;
     }
 
+    /**
+     * @param bool $as_num
+     *
+     * @return array|float
+     */
     public static function hrtime($as_num = false)
     {
         if ($as_num) {
--- a/src/Php73/Php73.php
+++ b/src/Php73/Php73.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Symfony\Polyfill\Php73;
+
+/**
+ * @internal
+ */
+final class Php73
+{
+    const NANO_IN_SEC = 1000000000.0;
+
+    public static function is_countable($var)
+    {
+        return is_array($var)
+            || $var instanceof \Countable
+            || $var instanceof \ResourceBundle
+            || $var instanceof \SimpleXmlElement;
+    }
+
+    public static function hrtime($as_num = false)
+    {
+        if ($as_num) {
+            return microtime(true) * self::NANO_IN_SEC;
+        }
+
+        $time = explode(' ', microtime());
+
+        return array((int) $time[1], (int) ($time[0] * self::NANO_IN_SEC));
+    }
+}
--- a/src/Php73/Php73.php
+++ b/src/Php73/Php73.php
@@ -26,13 +26,13 @@ final class Php73
     }
 
     /**
-     * @param bool $as_num
+     * @param bool $asNum
      *
      * @return array|float
      */
-    public static function hrtime($as_num = false)
+    public static function hrtime($asNum = false)
     {
-        if ($as_num) {
+        if ($asNum) {
             return microtime(true) * self::NANO_IN_SEC;
         }
 
--- a/src/Php73/bootstrap.php
+++ b/src/Php73/bootstrap.php
@@ -17,6 +17,6 @@ if (PHP_VERSION_ID < 70300) {
     }
 
     if (!function_exists('hrtime')) {
-        function hrtime($as_num = false) { return p\Php73::hrtime($as_num); }
+        function hrtime($asNum = false) { return p\Php73::hrtime($asNum); }
     }
 }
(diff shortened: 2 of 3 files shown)


IonBazan:
Ping @nicolas-grekas @carusogabriel

src/Php73/Php73.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Symfony\Polyfill\Php73;
+
+/**
+ * @internal
+ */
+final class Php73
+{
+    const NANO_IN_SEC = 1000000000.0;
+
+    public static function is_countable($var)
```
carusogabriel: Unnecessary :-1:

tests/Mbstring/MbstringTest.php
```diff
@@ -195,7 +195,7 @@ public function testStrCase()
         }
         if (PHP_VERSION_ID >= 70000) {
             // Native iconv() is buggy before PHP 7
-            $this->assertSame('ab', str_replace('?', '', mb_strtolower(urldecode('a%A1%C0b'))));
+            $this->assertSame('ab', str_replace('?', '', mb_strtolower(urldecode('a%A1%C0b'), 'ASCII')));
```
stof: why is this included in this PR ? It looks totally unrelated

IonBazan: @nicolas-grekas done: https://bugs.php.net/bug.php?id=76319. I've tested that on my local master build and it failed too.

IonBazan: @stof I've removed this commit from PR.

IonBazan:
@carusogabriel @stof @nicolas-grekas I've added `@author` docblocks and removed unnecessary changes. Please review again when possible :smiley:

src/Php73/Php73.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        return is_array($var)
+            || $var instanceof \Countable
+            || $var instanceof \ResourceBundle
+            || $var instanceof \SimpleXmlElement;
+    }
+
+    /**
+     * @author Ion Bazan <ion.bazan@gmail.com>
```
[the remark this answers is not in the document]
IonBazan: @fabpot I've seen it in one of the classes but changed it to meet the requirements :+1:

src/Php73/bootstrap.php
```diff
@@ -9,8 +9,14 @@
...
+use Symfony\Polyfill\Php73 as p;
+
 if (PHP_VERSION_ID < 70300) {
     if (!function_exists('is_countable')) {
-        function is_countable($var) { return is_array($var) || $var instanceof Countable || $var instanceof ResourceBundle || $var instanceof SimpleXmlElement; }
+        function is_countable($var) { return p\Php73::is_countable($var); }
+    }
+
+    if (!function_exists('hrtime')) {
+        function hrtime($as_num = false) { return p\Php73::hrtime($as_num); }
```
fabpot: Should be `$asNum`

IonBazan: I wasn't sure if I should follow PSR or original function's parameter names here. I'll change it to PSR then.

src/Php73/Php73.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * @param bool $asNum
+     *
+     * @return array|float
+     */
+    public static function hrtime($asNum = false)
+    {
+        if ($asNum) {
+            return microtime(true) * self::NANO_IN_SEC;
```
nicolas-grekas: This overflows 100% of the time.
What's the unit of the native function?
Does this really return a number in nanoseconds (ignoring the floating part)?
(same below btw)
This should be double checked because right now I don't get how this is supposed to work.

IonBazan: @nicolas-grekas As of [PHP source](https://github.com/php/php-src/blob/1cdf5b7b4b27f35140d88d629d2977c9833384e2/ext/standard/hrtime.c#L163) it
> Returns an array of integers in form [seconds, nanoseconds] counted from an arbitrary point in time [...]
This example https://3v4l.org/Fne2W shows, that the value changes when called at different time.
Can you post an example output of this function overflowing?

IonBazan: @nicolas-grekas bump :)

nicolas-grekas: I need a few more days before being able to look again at it :)

IonBazan: Sure, take your time :+1:

src/Php73/Php73.php
```diff
@@ -0,0 +1,43 @@
...
+        return is_array($var)
+            || $var instanceof \Countable
+            || $var instanceof \ResourceBundle
+            || $var instanceof \SimpleXmlElement;
+    }
+
+    /**
+     * @param bool $asNum
+     *
+     * @return array|float
```
nicolas-grekas: So, I confirm this does not work yet: this should be `array|int` and the current logic overflows.
Since hrtime is relative to a random moment in the past, we should just subtract before multiplying, with an initialized offset, which could be computed like:
```php
class Php73
{
    public static $timeOffset;
    // [...]
}

Php73::$timeOffset = time();
```

src/Php73/Php73.php
```diff
@@ -0,0 +1,43 @@
...
+final class Php73
+{
+    const NANO_IN_SEC = 1000000000.0;
+
+    /**
+     * @param mixed $var
+     *
+     * @return bool
+     */
+    public static function is_countable($var)
```
nicolas-grekas: please revert this change: it is unrelated thus is noise to the current review

nicolas-grekas:
Closing in favor of #130, thank you @IonBazan

PR #127, opened 2018-05-17
Suggest ctype extension

BackEndTea:
This is in line with other poly fills like mbstring


fabpot:
Good catch, thanks @BackEndTea.

BackEndTea:
Thank you @fabpot, @nicolas-grekas

PR #128, opened 2018-05-30
Fix description typo for mb_strstr

vuongggggg:


fabpot:
Good catch, thanks @vuongggggg.

PR #130, opened 2018-07-04
PHP 7.3: Add hrtime() function

mcfedr:
I've taken https://github.com/symfony/polyfill/pull/125
- removed extra changes to other functions
- corrected scale issue with array version in microsecond to nanosecond conversion

### DIFF
--- a/src/Php73/Php73.php
+++ b/src/Php73/Php73.php
@@ -10,12 +10,13 @@ namespace Symfony\Polyfill\Php73;
  */
 final class Php73
 {
-    const NANO_IN_SEC = 1000000000.0;
+    const NANO_IN_SEC = 1e9;
+    const NANO_IN_MSEC = 1e3;
 
     /**
      * @param bool $asNum
      *
-     * @return array|float
+     * @return array|float|int
      */
     public static function hrtime($asNum = false)
     {
@@ -25,6 +26,6 @@ final class Php73
 
         $time = explode(' ', microtime());
 
-        return array((int) $time[1], (int) ($time[0] * self::NANO_IN_SEC));
+        return array((int) $time[1], (int) ($time[0] * self::NANO_IN_MSEC));
     }
 }
--- a/src/Php73/Php73.php
+++ b/src/Php73/Php73.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Symfony\Polyfill\Php73;
+
+/**
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
+ * @internal
+ */
+final class Php73
+{
+    const NANO_IN_SEC = 1000000000.0;
+
+    /**
+     * @param bool $asNum
+     *
+     * @return array|float
+     */
+    public static function hrtime($asNum = false)
+    {
+        if ($asNum) {
+            return microtime(true) * self::NANO_IN_SEC;
+        }
+
+        $time = explode(' ', microtime());
+
+        return array((int) $time[1], (int) ($time[0] * self::NANO_IN_SEC));
+    }
+}
--- a/tests/Php73/Php73Test.php
+++ b/tests/Php73/Php73Test.php
@@ -40,25 +40,27 @@ class Php73Test extends TestCase
 
     public function testHardwareTimeAsNum()
     {
-        $microtime = microtime(true);
         $hrtime = hrtime(true);
+        $this->assertTrue(is_float($hrtime) || is_int($hrtime));
+
         usleep(1000);
-        $d0 = (microtime(true) - $microtime) * 1000000000;
         $d1 = hrtime(true) - $hrtime;
 
         $this->assertEquals(10000000, $d1, '', 9e6);
-        $this->assertLessThan(0.05, abs($d0 - $d1) / $d1);
     }
 
     public function testHardwareTimeAsArray()
     {
-        $microtime = microtime(true);
         $hrtime = hrtime();
+        $this->assertInternalType('array', $hrtime);
+        $this->assertCount(2, $hrtime);
+        $this->assertInternalType('int', $hrtime[0]);
+        $this->assertInternalType('int', $hrtime[1]);
+
         usleep(1000);
-        $microDelta = (microtime(true) - $microtime) * 1000000000;
         $hrtime2 = hrtime();
 
         $this->assertSame(0, $hrtime2[0] - $hrtime[0]);
-        $this->assertEquals($microDelta, $hrtime2[1] - $hrtime[1], '', 1e5);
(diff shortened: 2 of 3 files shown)


mcfedr:
To address some comments from the other PR, 

when getting a number:
- the documentation says maybe be int or float, http://php.net/manual/en/function.hrtime.php
- this polyfill will always return a float, because the large numbers overflow, not sure if its possible to prevent this.

regarding arbitrary point in time when the counter starts
- it would be possible to create a starting point, but why not just use 1970-01-01 as that arbitrary point in time? and keep the polyfill simple.

In anycase this polyfill isn't completely correct - as far as I can tell there is no way to polyfill 100%, as one of the points of `hrtime` is that it never goes backwards, even with system clock changes.

tests/Php73/Php73Test.php
```diff
@@ -37,4 +37,30 @@ public function testIsCountableForGenerator()
...
+        $hrtime = hrtime();
+        $this->assertInternalType('array', $hrtime);
+        $this->assertCount(2, $hrtime);
+        $this->assertInternalType('int', $hrtime[0]);
+        $this->assertInternalType('int', $hrtime[1]);
+
+        usleep(1000);
+        $hrtime2 = hrtime();
+
+        $this->assertSame(0, $hrtime2[0] - $hrtime[0]);
```
stof: isn't this subject to race conditions ?

mcfedr: Sort of, the `usleep` call make it sleep for 0.001 of a second, if the system causes it to sleep 1000 times longer, then this might fail

mcfedr:
> because microtime(true) * self::NANO_IN_SEC overflows and looses all its precision.

Yes, I see what you mean now, I've changed this now, starts counting from first use of `hrtime`

src/Php73/Php73.php
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public static function hrtime($asNum = false)
+    {
+        if ($asNum) {
+            if (is_null(static::$startAt)) {
+                static::$startAt = microtime(true);
+            }
+            $nanos = (microtime(true) - static::$startAt) * self::NANO_IN_SEC;
+
+            if (PHP_INT_SIZE === 4) {
```
stof: should be fully-qualified to be resolved at compile time.

src/Php73/Php73.php
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * @param bool $asNum
+     *
+     * @return array|float|int
+     */
+    public static function hrtime($asNum = false)
+    {
+        if ($asNum) {
+            if (is_null(static::$startAt)) {
```
stof: use `self::` to access private static properties (here, as it is final, it won't allow breaking stuff by creating a child class, but it still pushes more work at runtime)

stof: Symfony uses `null ===` to compare to null, not `is_null`

src/Php73/Php73.php
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * @param bool $asNum
+     *
+     * @return array|float|int
+     */
+    public static function hrtime($asNum = false)
+    {
+        if (null === self::$startAt) {
+            self::$startAtArr = self::microtimeNumbers();
+            self::$startAt = self::$startAtArr[0] + (float) self::$startAtArr[1];
```
nicolas-grekas: The cast is not needed here

src/Php73/Php73.php
```diff
@@ -0,0 +1,65 @@
...
+            if (\PHP_INT_SIZE !== 4) {
```
nicolas-grekas: These checks look strange, what's the rationale?

src/Php73/Php73.php
```diff
@@ -0,0 +1,65 @@
...
+            $now = (int) (microtime(true) * self::MSEC_IN_SEC);
+
+            return ($now - self::$startAt) * self::NANO_IN_MSEC;
+        }
+
+        $time = self::microtimeNumbers();
+
+        $secs = $time[1] - self::$startAtArr[1];
+        $msecs = $time[0] - self::$startAtArr[0];
+        if ($msecs < 0) {
```
nicolas-grekas: I don't think this works: it doesn't prevent measuring two consecutive values with negative delta.
Let's remove it, don't you think?

mcfedr: because $time[0] is the number of milliseconds since the last second, so it could have been 10.8 seconds in `self::$startAtArr` but now its 15.1, so `$msecs` ends up as -0.7

src/Php73/Php73.php
mcfedr: The idea was in the 64 bit version, I'm going to use it as a int, so I do the int conversion once here.

mcfedr:
So I change it again, I think it makes more sense to do it this way, always use string version of microtime - means we can do more acurate float calculation

src/Php73/Php73.php
```diff
@@ -0,0 +1,55 @@
...
+    public static function hrtime($asNum = false)
+    {
+        if (null === self::$startAt) {
+            self::$startAt = self::microtime();
+        }
+
+        $time = self::microtime();
+
+        $secs = $time[1] - self::$startAt[1];
+        $msecs = $time[0] - self::$startAt[0];
```
stof: so `$msecs` is still in seconds here, right ?

nicolas-grekas:
Got it for the +/-1.
I'd suggest to remove the private method and remove as many extra calls as possible. Having two separate path for asNum true/false also.
The reason is that this is a perf-measuring function, we should add as little overhead as possible.

mcfedr:
Ok, So I'm putting it back how it was, did a little test, the float version comes out about 4x faster than array version. Then I'm thinking, maybe just always use the float version? - so I wrote that, and tested the difference in performance

Its a super simple perf check:
```
for ($i = 0; $i < 10000000; $i++) {
    hrtime();
}
```

|                          | hrtime(true) | hrtime() |
|--------------------------|--------------|----------|
| [Separate array and float](https://github.com/mcfedr/polyfill/blob/872c12569534cb9762e054b88c61e1d0d36cfffe/src/Php73/Php73.php) | 0m2.094s     | 0m9.478s |
| [Array only](https://github.com/mcfedr/polyfill/blob/5d114de9572ac54f2de3f5764c0fde5c97243c84/src/Php73/Php73.php)               | 0m9.375s     | 0m9.491s |
| [Float only](https://github.com/mcfedr/polyfill/blob/04f81ad5c5961c10a00face1e3b88abdea7ead72/src/Php73/Php73.php)               | 0m1.985s     | 0m3.239s |

The float only version clearly wins, so pushing that for the moment, I dont think this is any advantage to the array version, I am *pretty sure* we don't lose any accuracy...

mcfedr:
@nicolas-grekas is there anything else preventing this being finished?

PR #131, opened 2018-07-05
Update for PHPUnit 7.2

DavidPrevot:
Here are some changes I had to make in order to run the test suite with PHPUnit 7.2. Posting it here in case it becomes relevant for you.

Regards

David


nicolas-grekas:
@DavidPrevot, @keradus is right: can you work on a patch that would keep this working with PHP 5.3 please? We need it to test all polyfills.

nicolas-grekas:
I'm closing due to lack of feedback. I created #144 to keep track of the topic.
Please resubmit if you have to fix the comments above.
Thank you for raising this.

PR #132, opened 2018-07-10
Allow NOP version of paragonie/random_compat

z38:
`paragonie/random_compat` ships a [special version 9.99.99](https://github.com/paragonie/random_compat#version-99999) which is only installable on PHP ≥7.0. It consists of an empty `composer.json` and adds no files to the autoloader. It might provide a minor perfomance improvement.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": ">=5.3.3",
         "ircmaxell/password-compat": "~1.0",
-        "paragonie/random_compat": "~1.0|~2.0",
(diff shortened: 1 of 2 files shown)


BackEndTea:
Should we do the same for our polyfills as well?

nicolas-grekas:
Thank you @z38.

nicolas-grekas:
@BackEndTea looks like a good idea, would you like to submit a PR doing so? I just created a `v99` branch so you could submit your PR against it.

PR #133, opened 2018-07-17
PHP 7.3: Add array_key_first and array_key_last functions

pierredup:
RFC: https://wiki.php.net/rfc/array_key_first_last
Added to PHP in php/php-src@50516a6e1f357f63324ec76e26623f2a7185bd4f

### DIFF
--- a/src/Php73/bootstrap.php
+++ b/src/Php73/bootstrap.php
@@ -20,4 +20,12 @@ if (PHP_VERSION_ID < 70300) {
         p\Php73::$startAt = (int) microtime(true);
         function hrtime($asNum = false) { return p\Php73::hrtime($asNum); }
     }
+
+    if (!function_exists('array_key_first')) {
+        function array_key_first(array $array) { foreach ($array as $key => $value) { return $key; } }
+    }
+
+    if (!function_exists('array_key_last')) {
+        function array_key_last(array $array) { $key = null; foreach ($array as $key => $value); return $key; }
+    }
 }
--- a/tests/Php73/Php73Test.php
+++ b/tests/Php73/Php73Test.php
@@ -15,6 +15,7 @@ use PHPUnit\Framework\TestCase;
 
 /**
  * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ * @author Pierre du Plessis <pdples@gmail.com>
  */
 class Php73Test extends TestCase
 {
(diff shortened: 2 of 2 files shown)


src/Php73/bootstrap.php
```diff
@@ -13,4 +13,12 @@
     if (!function_exists('is_countable')) {
         function is_countable($var) { return is_array($var) || $var instanceof Countable || $var instanceof ResourceBundle || $var instanceof SimpleXmlElement; }
     }
+
+    if (!function_exists('array_key_first')) {
+        function array_key_first(array $array) { reset($array); return key($array); }
```
BackEndTea: If i'm correct this will change the internal state of the array, while the php 7.3 functionality will keep the internal stat intact.

How about changing this function to not mess with the internal state as well?

stof: that's indeed necessary to emulate the correct behavior.

BackEndTea: [`array_slice`](http://php.net/manual/en/function.array-slice.php) with preserved keys should do the trick

stof: well, actually, this will not mess the external array. But it will trigger a copy of the array (to be able to alter the one passed to the function).
So we are emulating the right behavior, but with a performance cost. This should rather make a foreach loop with a return statement in the first iteration.

pierredup: I went for `array_slice` in both cases (passing the array by reference to prevent a copy).
This works as expected and some initial tests shows the internal pointer of the array isn't modified and there is no additional memory overhead

stof: no need to pass the array by reference just to avoid a copy. That could actually *create* a copy, and that would not polyfill the PHP 7.3 function, as it would have a different signature

BackEndTea:
How does this, and the php 7.3 implementation, behave on an object that isn't an array, but array accessible?

What about a string, or something that isn't anything like an array?

stof:
The usage of a reference should be removed:

- it creates a different signature than the official function (making the polyfill buggy)
- it can trigger 2 array copies on PHP 5.6 (once to pass the array by reference if it was already referenced twice, and once again to pass it to `array_slice` which does not expect a reference. Don't try to be smarter than PHP about avoiding copies.

stof:
> How does this, and the php 7.3 implementation, behave on an object that isn't an array, but array accessible?
> 
> What about a string, or something that isn't anything like an array?

The array typehint would break, making it a TypeError (or the relevant error reporting on PHP 5.x).
Btw, ArrayAccess does not enforce the object to be iterable, so there is no concept of *first* key (and it does not require being able to enumerate keys either)

pierredup:
By not passing the array by reference, won't that anyway trigger a copy of the array? (since the values need to be copied to the function). If that is the case, then the original implementation can stay (using `reset` and `end`) since the performance hit would already be there

stof:
@pierredup no. PHP performs copy-on-write. The usage of a reference argument for `reset` or `end` was the trigger for copying the memory, not passing the argument by value.

pierredup:
Right, so how should we get the last key of the array then? The first key is easy with a foreach, but getting the last key will then have the performance hit (either copying the memory, or iterating through the entire array to get to the last key)

stof:
@pierredup copying the array is `O(n)` in time as well. So we should rather be `O(n)` in time and `O(1)` in memory than `O(n)` for both

src/Php73/bootstrap.php
```diff
@@ -13,4 +13,12 @@
     if (!function_exists('is_countable')) {
         function is_countable($var) { return is_array($var) || $var instanceof Countable || $var instanceof ResourceBundle || $var instanceof SimpleXmlElement; }
     }
+
+    if (!function_exists('array_key_first')) {
+        function array_key_first(array $array) { foreach ($array as $key => $value) { return $key; } }
+    }
+
+    if (!function_exists('array_key_last')) {
+        function array_key_last(array $array) { return key(array_slice($array, -1, 1, true)); }
```
nicolas-grekas: I'd suggest to go with a foreach here also:
`$key = null; foreach ($array as $key => $value); return $key;`

nicolas-grekas:
How about smaller arrays?

pierredup:
Smaller arrays perform slightly better (https://blackfire.io/profiles/compare/604ecab6-1d23-4486-832d-216f27fc862c/graph)

nicolas-grekas:
you mean, even for smaller arrays foreach is always slower? let's always use array_slice then, and sorry for the extra work :)

pierredup:
No, I mean for smaller arrays foreach is faster (I tested against 100 and 1000 items and foreach was faster in both cases). This would probably be more common in real-world scenarios than large arrays, so will keep the foreach

nicolas-grekas:
Thank you @pierredup.

PR #137, opened 2018-08-04
Add hrtime() function

nicolas-grekas:
@mcfedr I'm wondering: wouldn't this be enough?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+* v1.9.0
+
+ * added polyfill for `hrtime()`
+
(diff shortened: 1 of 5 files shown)


mcfedr:
Yea, I think its enough. I tried to avoid float calculations in the 64 bit version, but probably thats just adding complications as the accuracy gains are unlikely to make a difference

nicolas-grekas:
Actually, I updated the implementation: the precision loss made tests fail on 32-bit. Ready now.

nicolas-grekas:
Thank you @IonBazan.

PR #138, opened 2018-08-05
Introduce NOOP versions of polyfills

BackEndTea:
In the spirit of paragonie/random_compat v9.99.99.

This would be tagged as a high version, to make sure that composer installs these if the poly fill is not 'needed'. I removed the auto loads, so there should be 0 run time impact when this version of the poly fill is present.

For now i have opted to leave the actual php code in the package, but it could be removed if desired.

@nicolas-grekas Could you add a v99 branch that i can target this PR to?


nicolas-grekas:
target updated

nicolas-grekas:
Oh, and I think e.g. for mbstring this is a bit more complex as this polyfill can add functions even when the extension is installed (on older version of php)

BackEndTea:
> The type of package should be updated to metapackage, the empty autoload section should be removed, and the extra also.
Last but not least, all the .php + tests-related files should be removed.

Done

> Oh, and I think e.g. for mbstring this is a bit more complex as this polyfill can add functions even when the extension is installed (on older version of php)

What would be the best solution here? require both the php version, and the extension?

composer.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "symfony/polyfill",
-    "type": "library",
+    "type": "metapackage",
```
nicolas-grekas: actually, because of polyfill-util, the root package should stay as a library, and its autoload section should list Util.

src/Apcu/bootstrap.php
```diff
@@ -1,48 +0,0 @@
...
- */
-
-use Symfony\Polyfill\Apcu as p;
-
-if (!extension_loaded('apc') && !extension_loaded('apcu')) {
-    return;
-}
-
-if (!function_exists('apcu_add')) {
-    if (extension_loaded('Zend Data Cache')) {
```
nicolas-grekas: this case is not covered by just requiring ext-apcu+PHP7.3

src/Ctype/composer.json
```diff
@@ -1,34 +0,0 @@
-{
-    "name": "symfony/polyfill-ctype",
-    "type": "library",
```
nicolas-grekas: not sure this file should be removed

src/Mbstring/composer.json
```diff
@@ -16,19 +16,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3"
+        "php": ">=5.3.3",
```
nicolas-grekas: should be bumped to 7.2 to account for `mb_chr`

composer.json
```diff
@@ -16,10 +16,12 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3",
-        "ircmaxell/password-compat": "~1.0",
-        "paragonie/random_compat": "~1.0|~2.0|~9.99",
```
nicolas-grekas: password-compat and password-compat could be adde to the replace section: `"...compat": "*"` (same in the PHP7 polyfill)

BackEndTea: I have added them to the replace section, as well as the `ircmaxell/password-compat` to php 5.5

src/Apcu/bootstrap.php
BackEndTea: Isn't this just to make use of a 'better' solution if the Zend Data Cache is available? It provides the same functions even when this isn't available.

nicolas-grekas: this allows using the apcu functions when only Zend Data Cache is installed.
maybe add a conflict rule with "ext-Zend Data Cache": "*"?

BackEndTea: I have added the conflict rule

nicolas-grekas: thanks. can you add it to the root composer.json also?

nicolas-grekas:
What about updating the root README to explain what this version is?
And also, could you send another PR against the master branch to tell about this branch and how it works?

BackEndTea:
I have added the following section to the readme. Would that be okay or should i word thing differently.
```
It is recommended to allow installation of  both the `1.*` and `9.*`versions of
this package. The `9.*` versions are a special noop version that can only be
installed if the extension, or PHP version required is present. These special
versions do no populate the autoloader and do therefore not add any `require` 
statements to the bootstrap process, resulting in 0 overhead during runtime.
```

nicolas-grekas:
Thanks for the PR, now we can see the precise picture.
I've been thinking a bot more about this and I think we should not do this actually.
My reasoning is that there are two downsides to this approach:
- it requires cooperation from every lib author that requires a polyfill.
- it relies on composer to resolve the version, with a different result based on the runtime at that time. By experience on Symfony, this is a foot-gun for many, who create their lock file on a different PHP/extension set than their prod.

Instead, there is another strategy that is free from these: putting the polyfills in the "replace" section of your root composer.json. The symfony skeleton does it, see https://github.com/symfony/skeleton/blob/master/composer.json

Since this way has no downside, I think that's what we should encourage doing instead (and put it in the readme then).

Does that make sense?

BackEndTea:
It does make sense. However these polyfills are generally not a direct dependency of a project, but rather a dependency of another dependency somewhere. Which means that for a lot of users may not know they are using it, or know that they can add it to the replace section.

When we look at [paragonie/random_compat](https://packagist.org/packages/paragonie/random_compat/stats#v9.99.99) on packagist, we can see that it is already at 16k downloads per day, and that is before the [symfony/polyfill-70](https://packagist.org/packages/symfony/polyfill-php70/stats#v1.8.0) allowed its installation. So it does require cooperation from as many lib authors as possible, but it should see installations. I don't think it should fall to the end users who don't interact with this polyfill directly.

> It relies on composer to resolve the version, with a different result based on the runtime at that time. A By experience on Symfony, this is a foot-gun for many, who create their lock file on a different PHP/extension set than their prod.

This is of course an issue, which will end up in a difficult to debug situation. I'm not sure what the best way would be to tackle this, if possible at all.

All in all i think this version could provide a benefit, but it does come with a (potential) risk, and the benefit is somewhat marginal.

BackEndTea:
Another option for this would be https://github.com/composer/composer/issues/7557

nicolas-grekas:
Thank you for proposing and implementing this approach @BackEndTea 
I'm closing as explained: "replacing" the packages is the approach we'd like to follow instead.

PR #146, opened 2018-09-20
Duplicate mb_ord/mb_chr/mb_scrub in PHP 7.2  polyfill

nicolas-grekas:
When our skeleton "replace" polyfill-mbstring but PHP 7.1 is used, it means these functions are *not* polyfilled (since they've been introduced in PHP 7.2).

Let's duplicate them in the 7.2 polyfill.


fabpot:
Thank you @nicolas-grekas.

PR #147, opened 2018-09-21
DX: PHP CS Fixer configuration

keradus:
as you requested, @nicolas-grekas


keradus:
ah, old PHP :D will fix in a minute

nicolas-grekas:
Thank you @keradus.

PR #148, opened 2018-09-21
Add polyfill for Intl\MessageFormatter

nicolas-grekas:
This is a fork of Yii's [fallback MessageFormatter](https://github.com/yiisoft/yii2/blob/master/framework/i18n/MessageFormatter.php) implementation, ported as a polyfill of intl's MessageFormatter. I also reformatted the code to make it more in line with our practices and added some notes if we want to push the polyfill forward. The test suite compares both native and PHP variants, with similar results.
Kudos to them!

ping @cebe and @samdark FYI

### DIFF
--- a/src/Intl/MessageFormatter/MessageFormatter.php
+++ b/src/Intl/MessageFormatter/MessageFormatter.php
@@ -1,0 +1,316 @@
+<?php
+
+/*
+ * Copyright © 2008 by Yii Software LLC (http://www.yiisoft.com)
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *  * Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  * Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *  * Neither the name of Yii Software LLC nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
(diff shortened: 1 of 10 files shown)


src/Intl/MessageFormatter/MessageFormatter.php
```diff
@@ -0,0 +1,314 @@
...
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Originally forked from
+ * https://github.com/yiisoft/yii2/blob/master/framework/i18n/MessageFormatter.php
```
cebe: you should probably make this a permalink pointing to a tag instead of master to make sure the link does not get broken later.

https://github.com/yiisoft/yii2/blob/2.0.15/framework/i18n/MessageFormatter.php

tests/Intl/MessageFormatter/MessageFormatterTest.php
```diff
@@ -0,0 +1,279 @@
...
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * Originally forked from
+ * https://github.com/yiisoft/yii2/blob/master/tests/framework/i18n/FallbackMessageFormatterTest.php
```
cebe: same here:

https://github.com/yiisoft/yii2/blob/2.0.15/tests/framework/i18n/FallbackMessageFormatterTest.php

src/Intl/MessageFormatter/MessageFormatter.php
nicolas-grekas: Good idea, done, thanks!

src/Intl/MessageFormatter/MessageFormatter.php
```diff
@@ -0,0 +1,316 @@
...
+    public static function formatMessage($locale, $pattern, array $args)
+    {
+        if (null === $formatter = self::create($locale, $pattern)) {
+            return false;
+        }
+
+        return $formatter->format($args);
+    }
+
+    public function __construct($locale, $pattern)
```
fabpot: should be first method

fabpot:
Thank you @nicolas-grekas.

Issue #113, opened 2018-02-13
Polyfill for intl-idn


gharlan:
What do you think about adding a polyfill for [intl-idn](http://php.net/manual/en/ref.intl.idn.php)?

nicolas-grekas:
why not :) do you have code?

keradus:
why not.
perhaps https://github.com/true/php-punycode ?

gharlan:
> do you have code?

Not yet. And I'm not sure, if I am able to do this. But maybe I will try.

> perhaps https://github.com/true/php-punycode ?

I guess we have to re-implement this to be compatible with the original functions.

Should the polyfill support the deprecated variant `INTL_IDNA_VARIANT_2003`? (I don't think so)

nicolas-grekas:
It's OK to have a partial polyfill, so no, no need to support deprecated features (and if someone comes with a PR later, fine :) )

gharlan:
This is the php-src part of the functions: https://github.com/php/php-src/blob/master/ext/intl/idn/idn.c
But the main conversion is done by functions `uidna_nameToASCII_UTF8` and `uidna_nameToUnicodeUTF8` from ICU.

keradus:
> I guess we have to re-implement this to be compatible with the original functions.

well, they tried to be compatible.

we can cover it with tests, sure, see what's missing
simply, there may be no need to reimplement everything from scratch

gharlan:
Thanks @lbassin for the pull request!

PR #149, opened 2018-09-24
Add polyfill for Intl\Idn

lbassin:
This PR add the polyfill intl-idn suggested in #113 
The original code came from https://github.com/true/php-punycode


src/Intl/Idn/README.md
```diff
@@ -0,0 +1,12 @@
+Symfony Polyfill / Ctype
```
stof: wrong title

src/Intl/Idn/Idn.php
```diff
@@ -0,0 +1,382 @@
...
+     * @param string $domain Domain to convert in IDNA ASCII-compatible format.
+     * @param int $options Conversion options - combination of IDNA_* constants (except IDNA_ERROR_* constants).
+     * @param int $variant Either INTL_IDNA_VARIANT_2003 for IDNA 2003 or INTL_IDNA_VARIANT_UTS46 for UTS #46.
+     * @param array &$idna_info This parameter can be used only if INTL_IDNA_VARIANT_UTS46 was used for variant.
+     *
+     * @return string Punycode representation in ASCII
+     */
+    public static function idn_to_ascii($domain, $options, $variant, &$idna_info = array())
+    {
+        if ($variant !== INTL_IDNA_VARIANT_2003) {
```
stof: So, no support for `INTL_IDNA_VARIANT_UTS46`, while `INTL_IDNA_VARIANT_2003` is deprecated as of PHP 7.2 ?

src/Intl/Idn/Idn.php
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        return $output;
+    }
+
+    private static function encodePart($input)
+    {
+        $codePoints = self::listCodePoints($input);
+
+        $n = static::INITIAL_N;
```
stof: should use `self::`. It does not make sense to support overriding the constants, so using late static binding is useless (and removes opportunities for OPCache to optimize things as it forces runtime evaluation)

src/Intl/Idn/bootstrap.php
```diff
@@ -0,0 +1,22 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Intl\Idn as p;
+
+if (!function_exists('idn_to_ascii')) {
+    define('IDNA_DEFAULT', 0);
```
stof: what about all the other `IDNA` constants allowing to configure it ?

stof:
Github says there is some conflicts in your branch, which also prevent the CI from running. Please rebase your branch (see https://symfony.com/doc/current/contributing/code/patches.html#step-3-submit-your-patch for documentation about that process if needed)

src/Intl/Idn/README.md
```diff
@@ -0,0 +1,12 @@
+Symfony Polyfill / Intl: Idn
+========================
```
fabpot: missing some `=`

tests/Intl/Idn/IdnTest.php
```diff
@@ -0,0 +1,214 @@
...
+ * @author Dmitry Lukashin <dmitry@lukashin.ru>
+ * @author Fahad Ibnay Heylaal <fahad19@gmail.com>
+ * @author ceeram <c33ram@gmail.com>
+ * @author Laurent Bassin <laurent@bassin.info>
+ *
+ * @covers Symfony\Polyfill\Intl\Idn\Idn::<!public>
+ */
+class IdnTest extends TestCase
+{
+
```
fabpot: extra blank line

stof:
As the `INTL_IDNA_VARIANT_UTS46` variant is the recommended one (`INTL_IDNA_VARIANT_2003` is deprecated as of PHP 7.2), I think it does not make sense to merge a polyfill supporting only `INTL_IDNA_VARIANT_2003`

src/Intl/Idn/Idn.php
```diff
@@ -0,0 +1,287 @@
...
+        'g' => 6, 'h' => 7, 'i' => 8, 'j' => 9, 'k' => 10, 'l' => 11,
+        'm' => 12, 'n' => 13, 'o' => 14, 'p' => 15, 'q' => 16, 'r' => 17,
+        's' => 18, 't' => 19, 'u' => 20, 'v' => 21, 'w' => 22, 'x' => 23,
+        'y' => 24, 'z' => 25, '0' => 26, '1' => 27, '2' => 28, '3' => 29,
+        '4' => 30, '5' => 31, '6' => 32, '7' => 33, '8' => 34, '9' => 35,
+    );
+
+    public static function idn_to_ascii($domain, $options, $variant, &$idna_info = array())
+    {
+        if (INTL_IDNA_VARIANT_2003 === $variant && version_compare(PHP_VERSION, '7.2.0') >= 0) {
```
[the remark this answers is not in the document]
stof: and the version check should be first in the `if`, to let OPCache optimize as much of the condition as it can (`... && false` cannot be removed safely at compile time)

src/Intl/Idn/bootstrap.php
```diff
@@ -0,0 +1,27 @@
...
+    define('INTL_IDNA_VARIANT_2003', 0);
+    define('INTL_IDNA_VARIANT_UTS46', 1);
+
+    if (\PHP_VERSION_ID < 70400) {
+        define('INTL_IDNA_VARIANT_DEFAULT', INTL_IDNA_VARIANT_2003);
```
stof: is this a constant defined in Intl ? If no, we should maybe define a namespaced constant to keep it in the polyfill)

lbassin: No this is a custom constant
What do you mean by namespaced constant ? Something like 
`define('Symfony\Polyfill\Tests\Intl\Idn\VARIANT_DEFAULT', ...);` ?

stof: yes, so that it is clear that this is not part of the polyfilled API

lbassin:
I am kind of stuck with the last error i got with php 5.3
The problem is the function signature changed after php 5.3. There are only two arguments in 5.3...

The way tests are running in Util/TestListenerTrait.php in line 100 doesn't allow us to change signature depending of the php version

I have to find a way to fix that :confused: I'll continue to search a solution but if someone has an idea it would be great

src/Intl/Idn/bootstrap.php
lbassin: I just added them this morning

lbassin:
I think i fixed all the feedbacks :)

Here a quick update for this PR : 

- The polyfill works perfectly with 2003 variant
The IDNA_* flags and the last argument idna_info are used for UTS46 only 

- I made some research about UTS46 implementation :
From what I understand, the algorithm is the same but the differences are on the characters the function is allowed to encode / decode
A full explanation : https://unicode.org/reports/tr46/#Validity_Criteria
Official test cases for UTS46 : https://www.unicode.org/Public/idna/6.0.0/IdnaTest.txt
This repo may help: https://github.com/jcranmer/idna-uts46/blob/master/uts46.js (UTF46 implementation by in javascript. I didn't find php implementation for UTS46)

In my opinion this PR is a good start, there is at least 2003 variant available and we may add UTS46 later

nicolas-grekas:
> about UTS46 implementation: From what I understand, the algorithm is the same

does this mean we could duplicate the test cases but use TS46 instead while still having them green?
If yes, then let's do it! (don't forget removing the `@group legacy` annotation on them.)

lbassin:
I just added the UTS46 tests (Same as 2003) and still green

The next step would be to add some that only work with 2003 and others with UTS46

nicolas-grekas:
I push forced on your fork. The 1st commit is your work, the 2nd are some changes on top.
This PR is ready for a first version.
As far as polyfilling is concerned, this is a very basic implementation: it misses handling the option bitfield, misses properly populating $idna_info and misses proper stringprep (NFKC normalization, maybe more).
But that's what forking https://github.com/true/php-punycode provides us for now and that's a good start.
People could later contribute the missing bits.
Thank you.

fabpot:
Thank you @lbassin.

PR #150, opened 2018-09-25
array_key_last: improving performance

h4kuna:
First PR is [here](https://github.com/symfony/polyfill-php73/pull/1), where was discussion.

Q: The issue is that this mutates the internal pointer.
A: created [benchmark](https://github.com/symfony/polyfill-php73/pull/1#issuecomment-424145092) and implementation does not change internal pointer.

Q: about performance for small vs big arrays
A: look like implementation is [every time faster](https://github.com/symfony/polyfill-php73/pull/1#issuecomment-424217955)

This implementation is approx twice faster than cycle foreach.

Benchmark is [on my gist](https://gist.github.com/h4kuna/6353ce99bd45b9c5fcdf49ef6d0e3da8). 

Results on my machine.

```
array_key_last_symfony
0.1387140750885

array_key_last
0.055586814880371
```

### DIFF
--- a/src/Php73/bootstrap.php
+++ b/src/Php73/bootstrap.php
@@ -26,6 +26,6 @@ if (PHP_VERSION_ID < 70300) {
     }
 
     if (!function_exists('array_key_last')) {
-        function array_key_last(array $array) { $key = null; foreach ($array as $key => $value); return $key; }
+        function array_key_last(array $array) { end($array); return key($array); }
     }
 }


BackEndTea:
`end` changes the internal pointer of the array. And the 'official' 7.3 `array_key_last` (and first), don't change the internal pointer.

The `foreach` approach was taken to make sure that the polyfill behaves the same as the official version.

nicolas-grekas:
Actually it doesn't change the internal pointer, because the "end" operates on a copy of the array (argument is passed by value and not by reference to the function)

h4kuna:
Why this test [does not failed](https://github.com/symfony/polyfill/blob/master/tests/Php73/Php73Test.php#L125)?

stof:
but on PHP 5.x, the reference mismatch forces to create a copy of the array, and that's also not the behavior of the core method.

h4kuna:
@stof you have right. Could you check if the benchmarks are relevant?

Array size is 2000, on site is execute limit 3s for php.

### Memory usage
Here is [benchmark](http://sandbox.onlinephpfunctions.com/code/0fe160275c8dae5528a2c4d969d832c8603c26f8).

#### php 5.6.29
```
Array size: 2000
array_key_last_symfony: time: 1.5190160274506 s, memory: 80 B
array_key_last: time: 0.70938301086426 s, memory: 112 B
``` 

#### php 7+ return this result
```
Array size: 2000
array_key_last_symfony: time: 0.29880118370056 s, memory: 0 B
array_key_last: time: 0.13791918754578 s, memory: 0 B
```

### Peak of memory

[array_key_last](http://sandbox.onlinephpfunctions.com/code/a0854f92b5aa32ce9dcdf9e64dfbe7bc62d60a3b)
- php 7.0.1: 140.03 kB
- php 5.6.29: 468.34 kB

[array_key_last_symfony](http://sandbox.onlinephpfunctions.com/code/b3461ccf3edb4d5f758de0cd4259e4af0235d64e)
- php 7.0.1: 140.02 kB
- php 5.6.29: 468.34 kB

nicolas-grekas:
Thank you @h4kuna.

PR #153, opened 2018-10-02
Add a note about reducing overhead

BackEndTea:
As talked about in #138

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,27 @@ It adds only a few lightweight `require` statements to the bootstrap process
 to support all polyfills. Implementations are then loaded on-demand when
 needed during code execution.
 
+If your project requires a minimum PHP version it is advisable to add polyfills
+for lower PHP versions to the `replace` section of your `composer.json`. 
+This removes any overhead from these polyfills as they are no longer part of your project.
+The same can be done for polyfills for extensions that you require.
+
+If your project requires php 7.0, and needs the mb extension, the replace section would look
+something like this.
+
+```json
+{
+    "replace": {
+        "symfony/polyfill-php54": "1.99",
+        "symfony/polyfill-php55": "1.99",
+        "symfony/polyfill-php56": "1.99",
+        "symfony/polyfill-php70": "1.99",
+        "symfony/polyfill-mbstring": "1.99"
+    }
+}
+
+```
+
 Polyfills are unit-tested alongside their native implementation so that
 feature and behavior parity can be proven and enforced in the long run.
 


README.md
```diff
@@ -86,6 +86,11 @@ It adds only a few lightweight `require` statements to the bootstrap process
 to support all polyfills. Implementations are then loaded on-demand when
 needed during code execution.
 
+If your project requires a minimum PHP version it is advisable to add polyfills
+for lower PHP versions to the `replace` section of your `composer.json`. 
+This removes any overhead from these polyfills as they are no longer part of your project.
+The same can be done for polyfills for extensions that you require.
```
fabpot: What about adding a `composer.json` snippet to illustrate how the `replace` section would look like for a give version of PHP?

README.md
```diff
@@ -86,6 +86,27 @@ It adds only a few lightweight `require` statements to the bootstrap process
...
+something like this.
+
+```json
+{
+    "replace": {
+        "symfony/polyfill-php54": "*",
+        "symfony/polyfill-php55": "*",
+        "symfony/polyfill-php56": "*",
+        "symfony/polyfill-php70": "*",
+        "symfony/polyfill-mbstring": "*"
```
stof: replacing `*` has a drawback: it creates issues in case we got advisories for some old releases of the polyfill and projects use https://packagist.org/packages/roave/security-advisories to block them, as `*` will match the conflict rules and so be affected by any advisory.

I suggest recommending `1.99` instead as version.

BackEndTea: > ... as * will match the conflict rules ...

Will this replace cause an issue in combination with conflict rules? Because i don't completely understand what you are saying.

stof: well, if you have a conflict rule for `symfony/polyfill-mbstring: <1.3.2` for instance (let's imagine that we would have a security issue in older releases for instance, and so that you want to prevent them from being installed, which is what `roave/security-advisories` is about), a root package replacing `*` would be in conflict, as it would replace conflicting versions too.

The general rule should be to always use fixed versions for `replace` or `provide`, not ranges (because your own code provides a single implementation anyway).

BackEndTea: That seems like unexpected behavior to me, didn't know it works like that. I'll go ahead and update the PR

stof: Well, as soon as you use a range for `replace`, you face weird behavior, because it means that your root package has **multiple** versions when considering it as the replaced name.
I'm not even sure there is a valid use case for using a range there.

fabpot:
Thank you @BackEndTea.

BackEndTea:
Thank you @fabpot @stof & @xabbuh

README.md
```diff
@@ -86,6 +86,27 @@ It adds only a few lightweight `require` statements to the bootstrap process
 to support all polyfills. Implementations are then loaded on-demand when
 needed during code execution.
 
+If your project requires a minimum PHP version it is advisable to add polyfills
+for lower PHP versions to the `replace` section of your `composer.json`. 
+This removes any overhead from these polyfills as they are no longer part of your project.
+The same can be done for polyfills for extensions that you require.
+
+If your project requires php 7.0, and needs the mb extension, the replace section would look
```
javiereguiluz: I like this pull request and I was thinking about making some PR to Symfony Docs about this. However, this phrase is confusing:

```
If your project requires php 7.0, and needs the mb extension, the replace section
would look something like this.
```

But if your project requires PHP 7, the PHP constraint in composer.json wouldn't let you use it unless you have PHP 7, so you don't need the old PHP polyfills.

Moreover, these polyfills only provide some features of PHP 7, but they can't make a PHP 7 app run on a PHP 5.x server.

Same for the MB String extension. If you add it to the Composer requirements, you can't install the project unless you have that extension ... so you don't need the polyfill.

I think we miss here some comments about not adding PHP or PHP extensions constraints to composer.json or the "ignore-platform-reqs" option. What do you think @BackEndTea?

BackEndTea: > But if your project requires PHP 7, the PHP constraint in composer.json wouldn't let you use it unless you have PHP 7, so you don't need the old PHP polyfills.

You are right, this message is more aimed to users who use package that has a polyfill as a dependency. This means that your project also requires the polyfill. So you add it to the replace section to reduce overhead.

BackEndTea: As far as adding it to the symfony docs, please do, its probably a better place for that documentation then here.

README.md
teohhanhui: https://getcomposer.org/doc/04-schema.md#replace

I don't understand how `1.99` could work for `replace`. You're supposed to specify which version(s) you're replacing, and version `1.99` does not exist.

BackEndTea: The replace basically says: "I have the package, at this version." We chose 1.99 as that is a version that won't be out for a long time, if ever. 

We do so because of security checkers, which will match against replaced packages as well. With 1.99 not being out for a long time that means security checkers won't see it as a wrong version, if older versions contain a security fault.

teohhanhui: > The replace basically says: "I have the package, at this version."

That's incorrect.

But the Composer docs doesn't really explain it, so...

> You may use the `replace` key to specify which versions of the original package your fork is an API compliant replacement for. You should avoid using too generic version constraints (e.g. * or >=1.0.0) with replacements as with all other dependencies.

As @naderman explained here: https://blog.naderman.de/2014/02/17/replace-conflict-forks-explained/

PR #166, opened 2019-01-30
Redefine INTL_IDN_VARIANT_* constants on the Idn class

derrabus:
I'm currently working on a project where we have to deploy the application to servers with a strange configuration. The servers run php 5.6, but instead of the bundled `ext-intl` the very outdated version 1.1 of `pecl/intl` is installed.

If I install the polyfill there, the `bootstrap.php` scrit will find an already defined `idn_to_ascii` function and do nothing. This is why I'd like to force the usage of the polyfill code here.

The problem that I have is that the old `pecl/intl` does not define all constants. To be more precise, `INTL_IDNA_VARIANT_UTS46` is missing because version 1.1 did not yet support the `$variant` parameter.

In order to solve this problem for me, I have redefined all constants on the `Idn` class.

### DIFF
--- a/src/Intl/Idn/Idn.php
+++ b/src/Intl/Idn/Idn.php
@@ -43,6 +43,9 @@ namespace Symfony\Polyfill\Intl\Idn;
  */
 final class Idn
 {
+    const INTL_IDNA_VARIANT_2003 = 0;
+    const INTL_IDNA_VARIANT_UTS46 = 1;
+
     private static $encodeTable = array(
         'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l',
         'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x',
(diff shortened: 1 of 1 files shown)


derrabus:
I don't know, if anyone checks for the availability of the constants for feature detection, but if someone did, we would break that code.

The problem is that in my scenario, the extension already provided a `idn_to_ascii` function with a reduced feature set that the polyfill cannot replace. Meaning that this code would right now fail with an error about an undefined constant.

```php
idn_to_ascii($domain, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
```

If we defined the constant, that code above would be executed, but do unexpected things because the the native `idn_to_ascii` ignores the third parameter and does not implement the UTF46 variant yet.

nicolas-grekas:
and so, how do you plan to use the polyfill?

derrabus:
I would call the static class method directly.

```php
Idn::idn_to_ascii($domain, Idn::IDNA_DEFAULT, Idn::INTL_IDNA_VARIANT_UTS46);
```

nicolas-grekas:
OK, that's `@internal` so you're on your own :)
I'd suggest to seek for the minimum viable patch that would allow what you want.

derrabus:
> OK, that's `@internal` so you're on your own :)

I am aware of that.

> I'd suggest to seek for the minimum viable patch that would allow what you want.

I could reduce the list of class constants to only the two `INTL_IDNA_VARIANT_*` constants because the current implementation of the polyfill does not make use of any of the other constants.

derrabus:
Done. The class now only defines the constants it actually uses.

stof:
this might be OK for me, even though it only fixes things for people using internal APIs (as we cannot replace an existing implementation with a more complete polyfill)

nicolas-grekas:
Given #168, I think we should polyfill the const actually.

stof:
@nicolas-grekas I would vote against it for the same reason that in #168 actually.

nicolas-grekas:
Thank you @derrabus.

PR #175, opened 2019-06-28
ensure Php73 class is loaded for bootstrap

baohx2000:
Some composer plugins (like monorepo) will cause files to be loaded before the autoloader is fully registered, so class files must be manually loaded

### DIFF
--- a/src/Php73/bootstrap.php
+++ b/src/Php73/bootstrap.php
@@ -17,6 +17,7 @@ if (PHP_VERSION_ID < 70300) {
     }
 
     if (!function_exists('hrtime')) {
+        require_once __DIR__.'/Php73.php';
         p\Php73::$startAt = (int) microtime(true);
         function hrtime($asNum = false) { return p\Php73::hrtime($asNum); }
     }


src/Php73/bootstrap.php
```diff
@@ -17,6 +17,9 @@ function is_countable($var) { return is_array($var) || $var instanceof Countable
     }
 
     if (!function_exists('hrtime')) {
+        if (!class_exists(p\Php73::class)) {
+            require_once('Php73.php');
```
nicolas-grekas: I'd suggest removing the "if" and let `require_once` do its conditional job.
Also, this should be written as: require_once __DIR__.'/Php73.php';

src/Php73/bootstrap.php
```diff
@@ -17,6 +17,7 @@ function is_countable($var) { return is_array($var) || $var instanceof Countable
     }
 
     if (!function_exists('hrtime')) {
+        require_once(__DIR__.'/Php73.php');
```
BackEndTea: `require_once` is generally written without parenthesis

fabpot:
Thank you @baohx2000.

PR #178, opened 2019-07-14
add php 7.4snapshot

andreybolonin:

PR #181, opened 2019-07-22
add new PHP 7.4 functions

### LINKED ISSUE #179: [PHP 7.4] Add get_mangled_object_vars()
There is a new function coming in PHP 7.4 called `get_mangled_object_vars()` which returns same result as `(array) $object` with the exception that it ignores overloaded array casts, such as used by ArrayObject.
Output is also similar to `get_object_vars()` but it also shows protected and private fields.

To implement that we will probably need to use `ReflectionObject`.

PHP implementation and `phpt` test:
https://github.com/php/php-src/commit/eecd8961d94c50cc6cdc94ec80df8c1ce4881a76

IonBazan:
This PR adds new PHP 7.4 functions:
- [x] `get_mangled_object_vars` (https://github.com/php/php-src/commit/eecd8961d94c50cc6cdc94ec80df8c1ce4881a76)
- [x] `mb_str_split` (https://wiki.php.net/rfc/mb_str_split)
- [x] `password_algos` (https://wiki.php.net/rfc/password_registry)

Closes #179 and replaces #178

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,10 @@ matrix:
         - php: 7.2
         - php: 7.3
           env: SYMFONY_PHPUNIT_VERSION=7.2
+        - php: 7.4snapshot
+          env: SYMFONY_PHPUNIT_VERSION=7.2
         - php: nightly
+          env: SYMFONY_PHPUNIT_VERSION=7.2
     allow_failures:
         - php: nightly
     fast_finish: true
@@ -32,7 +35,7 @@ before_install:
     - phpenv config-rm xdebug.ini || echo "xdebug not available"
     - if [[ $TRAVIS_PHP_VERSION = 5.* ]]; then echo yes | pecl install -f apcu-4.0.11; fi
     - if [[ $TRAVIS_PHP_VERSION = 7.* ]]; then echo yes | pecl install -f apcu_bc-1.0.4; fi
-    - if [[ $TRAVIS_PHP_VERSION = 7.* ]]; then echo yes | pecl install -f apcu-5.1.11; fi
+    - if [[ $TRAVIS_PHP_VERSION = 7.* ]]; then echo yes | pecl install -f apcu-5.1.17; fi
     - php -i
 
 install:
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -35,6 +35,7 @@ namespace Symfony\Polyfill\Mbstring;
  * - mb_strlen               - Get string length
  * - mb_strpos               - Find position of first occurrence of string in a string
  * - mb_strrpos              - Find position of last occurrence of a string in a string
+ * - mb_str_split            - Convert a string to an array
  * - mb_strtolower           - Make a string lowercase
  * - mb_strtoupper           - Make a string uppercase
  * - mb_substitute_character - Set/Get substitution character
@@ -521,6 +522,34 @@ final class Mbstring
         $pos = iconv_strrpos($haystack, $needle, $encoding);
 
         return false !== $pos ? $offset + $pos : false;
+    }
+
+    public static function mb_str_split($string, $split_length = 1, $encoding = null)
+    {
+        if (null !== $string && !\is_scalar($string) && !(\is_object($string) && \method_exists($string, '__toString'))) {
+            trigger_error('mb_str_split() expects parameter 1 to be string, '.\gettype($string).' given', E_USER_WARNING);
+
+            return null;
+        }
+
+        if ($split_length < 1) {
+            trigger_error('The length of each segment must be greater than zero', E_USER_WARNING);
+
+            return false;
+        }
+
+        if (null === $encoding) {
+            $encoding = mb_internal_encoding();
+        }
+
+        $result = array();
+        $length = mb_strlen($string, $encoding);
+
+        for ($i = 0; $i < $length; $i += $split_length) {
+            $result[] = mb_substr($string, $i, $split_length, $encoding);
+        }
+
+        return $result;
     }
 
     public static function mb_strtolower($s, $encoding = null)
--- a/src/Php74/Php74.php
+++ b/src/Php74/Php74.php
@@ -1,0 +1,90 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Php74;
+
+/**
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
+ * @internal
+ */
+final class Php74
+{
+    public static function get_mangled_object_vars($obj)
+    {
+        if (!\is_object($obj)) {
+            trigger_error('get_mangled_object_vars() expects parameter 1 to be object, '.\gettype($obj).' given', E_USER_WARNING);
+
+            return null;
+        }
+
+        if ($obj instanceof \ArrayIterator || $obj instanceof \ArrayObject) {
+            $reflector = new \ReflectionClass($obj instanceof \ArrayIterator ? 'ArrayIterator' : 'ArrayObject');
+            $flags = $reflector->getMethod('getFlags')->invoke($obj);
+            $reflector = $reflector->getMethod('setFlags');
+
+            $reflector->invoke($obj, ($flags & \ArrayObject::STD_PROP_LIST) ? 0 : \ArrayObject::STD_PROP_LIST);
+            $arr = (array) $obj;
+            $reflector->invoke($obj, $flags);
+        } else {
+            $arr = (array) $obj;
+        }
+
+        return array_combine(array_keys($arr), array_values($arr));
+    }
+
+    public static function mb_str_split($string, $split_length = 1, $encoding = null)
+    {
+        if (null !== $string && !\is_scalar($string) && !(\is_object($string) && \method_exists($string, '__toString'))) {
+            trigger_error('mb_str_split() expects parameter 1 to be string, '.\gettype($string).' given', E_USER_WARNING);
+
+            return null;
+        }
+
+        if ($split_length < 1) {
+            trigger_error('The length of each segment must be greater than zero', E_USER_WARNING);
+
+            return false;
+        }
+
+        if (null === $encoding) {
+            $encoding = mb_internal_encoding();
+        }
+
+        $result = array();
+        $length = mb_strlen($string, $encoding);
+
+        for ($i = 0; $i < $length; $i += $split_length) {
+            $result[] = mb_substr($string, $i, $split_length, $encoding);
+        }
+
+        return $result;
+    }
+
+    public static function password_algos()
+    {
+        $algos = array();
+
+        if (\defined('PASSWORD_BCRYPT')) {
+            $algos[] = PASSWORD_BCRYPT;
+        }
+
+        if (\defined('PASSWORD_ARGON2I')) {
+            $algos[] = PASSWORD_ARGON2I;
+        }
+
+        if (\defined('PASSWORD_ARGON2ID')) {
+            $algos[] = PASSWORD_ARGON2ID;
+        }
+
+        return $algos;
+    }
+}
--- a/src/Php74/README.md
+++ b/src/Php74/README.md
@@ -1,0 +1,16 @@
+Symfony Polyfill / Php74
+========================
+
+This component provides functions added to PHP 7.4 core:
+
+- [`get_mangled_object_vars`](https://php.net/get_mangled_object_vars)
+- [`mb_str_split`](https://php.net/mb_str_split)
+- [`password_algos`](https://php.net/password_algos)
+
+More information can be found in the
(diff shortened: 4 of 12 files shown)


src/Php74/Php74.php
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @return array
+     */
+    public static function get_mangled_object_vars($obj)
+    {
+        if (!$obj instanceof \ArrayIterator && !$obj instanceof \ArrayObject) {
+            $result = array();
+
+            foreach ((array) $obj as $key => $value) {
+                $result[(string) $key] = $value;
```
stof: isn't this exactly `return (array) $obj` ?

src/Php74/Php74.php
```diff
@@ -0,0 +1,87 @@
...
+final class Php74
+{
+    /**
+     * @param object $obj
+     *
+     * @return array
+     */
+    public static function get_mangled_object_vars($obj)
+    {
+        if (!$obj instanceof \ArrayIterator && !$obj instanceof \ArrayObject) {
```
stof: I would rather write the special cases here, and have the final code path being the generic one. This will avoid having to update this condition each time we need to add another special case.

src/Php74/Php74.php
IonBazan: This fixes a BC break introduced in PHP 7.2: https://3v4l.org/MuD5X

IonBazan: @stof However I am not sure how exactly does it fix the issue - the casting seems optional here :man_shrugging:

src/Php74/Php74.php
IonBazan: I flipped the condition. If a new special case appears, we should consider moving it to a new method.

.travis.yml
```diff
@@ -19,8 +19,10 @@ matrix:
         - php: 7.2
         - php: 7.3
           env: SYMFONY_PHPUNIT_VERSION=7.2
+        - php: 7.4snapshot
         - php: nightly
     allow_failures:
+        - php: 7.4snapshot
```
nicolas-grekas: let's remove this line, we don't use allow_failures because everybody ignores unenforced jobs + it's green

src/Php74/Php74.php
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
+ * @internal
+ */
+final class Php74
+{
+    /**
+     * @param object $obj
```
nicolas-grekas: would be great to mimic PHP7.4's behavior when a non object is passed (+add test case for it)

src/Php74/Php74.php
```diff
@@ -0,0 +1,94 @@
...
+            $reflector = $reflector->getMethod('setFlags');
+
+            $reflector->invoke($obj, ($flags & \ArrayObject::STD_PROP_LIST) ? 0 : \ArrayObject::STD_PROP_LIST);
+            $result = (array) $obj;
+            $reflector->invoke($obj, $flags);
+
+            return $result;
+        }
+
+        foreach ((array) $obj as $key => $value) {
```
nicolas-grekas: This breaks references + should be fixed when using ArrayIterator/ArrayObject too.
But do we care really? That's mostly a theoretical issue to me, better ignore?

src/Php74/Php74.php
```diff
@@ -0,0 +1,94 @@
...
+
+        return $result;
+    }
+
+    /**
+     * @return array
+     */
+    public static function password_algos()
+    {
+        $algos = array(PASSWORD_BCRYPT);
```
nicolas-grekas: this one too should be under an "if" statement, as the polyfill runs on PHP< 5.5, isn't it?

src/Php74/composer.json
```diff
@@ -0,0 +1,31 @@
...
+        "php": ">=5.3.3"
+    },
+    "autoload": {
+        "psr-4": { "Symfony\\Polyfill\\Php74\\": "" },
+        "files": [ "bootstrap.php" ]
+    },
+    "minimum-stability": "dev",
+    "extra": {
+        "branch-alias": {
+            "dev-master": "1.11-dev"
```
nicolas-grekas: 1.12

nicolas-grekas:
Oh, please list the polyfill in the root composer.json and README files too!

src/Php74/Php74.php
IonBazan: Actually the test was passing because we've had the polyfill bootstapped. I added the `if` anyway.

src/Php74/Php74.php
IonBazan: Do you suggest `foreach ($arr as $key => &$value)`?

src/Php74/Php74.php
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
+ * @internal
+ */
+final class Php74
+{
+    /**
+     * @param object|mixed $obj
```
nicolas-grekas: no need to document unexpected types,
same for the return type

src/Php74/Php74.php
```diff
@@ -0,0 +1,105 @@
...
+            $reflector->invoke($obj, $flags);
+        } else {
+            $arr = (array) $obj;
+        }
+
+        foreach ($arr as $key => $value) {
+            $result[(string) $key] = $value;
+        }
+
+        return $result;
```
nicolas-grekas: `return array_combine(array_keys($arr), array_values($arr));` would fix the issue with references, if we care. (do we?)

src/Php74/Php74.php
stof: do we need the foreach at all ?

src/Php74/Php74.php
IonBazan: Wow, thanks for this! :clap:

src/Php74/composer.json
```diff
@@ -0,0 +1,31 @@
+{
+    "name": "symfony/polyfill-php74",
+    "type": "library",
+    "description": "Symfony polyfill backporting some PHP 7.4+ features to lower PHP versions",
+    "keywords": ["polyfill", "shim", "compatibility", "portable"],
+    "homepage": "https://symfony.com",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Nicolas Grekas",
```
BackEndTea: This should probably be your name/email 😃

tests/Php74/Php74Test.php
```diff
@@ -0,0 +1,99 @@
...
+        $this->assertSame(array(
+            "\0".'Symfony\Polyfill\Tests\Php74\B'."\0".'priv' => 4,
+            'pub' => 1,
+            "\0".'*'."\0".'prot' => 2,
+            "\0".'Symfony\Polyfill\Tests\Php74\A'."\0".'priv' => 3,
+            'dyn' => 5,
+            6 => 6,
+        ), get_mangled_object_vars($obj));
+
+        $this->assertNull(@get_mangled_object_vars(1));
```
BackEndTea: Can we test that an error was triggered with PHPUnit?

If its  possible it may be nice to make sure that happens as well.

tests/Php74/Php74Test.php
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        if (\defined('PASSWORD_ARGON2ID')) {
+            $this->assertContains(PASSWORD_ARGON2ID, $algos);
+        }
+    }
+
+    public function testMbStrSplit()
+    {
+        $this->assertFalse(@mb_str_split('победа', 0));
```
BackEndTea: Same as previous comment

tests/Php74/Php74Test.php
BackEndTea: Also, it may be nicer to move this to its own test.

src/Php74/Php74.php
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        if ($split_length < 1) {
+            trigger_error('The length of each segment must be greater than zero', E_USER_WARNING);
+
+            return false;
+        }
+
+        $result = array();
+        $length = mb_strlen($string, $encoding);
```
BackEndTea: Looks like we need `mb_` functions for this polyfill, how about adding `symfony/polyfill-mb` as a direct dependency? Otherwise this could break on certain systems.

Or maybe the mb polyfill should require this one, so that it polyfills all the `mb_` functions possible, i'm not sure what the standard is for this project.

IonBazan: I was also wondering about that. Seems that the code should be duplicated - `mb_chr` is in both mbstring and PHP 7.2 polyfills.

IonBazan: Added `symfony/polyfill-mbstring` as a dependency and duplicated the code following `IDN` and `mb_chr` polyfills.

src/Php74/README.md
```diff
@@ -0,0 +1,16 @@
+Symfony Polyfill / Php74
+========================
+
+This component provides functions added to PHP 7.4 core:
+
+- [`get_mangled_object_vars`](https://php.net/get_mangled_object_vars)
```
IonBazan: These links are not yet available. The should work after PHP 7.4 release.

IonBazan:
@stof @nicolas-grekas @BackEndTea PTAL

src/Mbstring/Mbstring.php
```diff
@@ -523,6 +524,41 @@ public static function mb_strrpos($haystack, $needle, $offset = 0, $encoding = n
         return false !== $pos ? $offset + $pos : false;
     }
 
+    /**
+     * @param string      $string
+     * @param int         $split_length
+     * @param string|null $encoding
+     *
+     * @return array|null|false
```
nicolas-grekas: let's remove the dockblock, other functions don't have it and it's not needed as tooling already know the signature of internal function - same for other classes/methods

src/Php74/composer.json
```diff
@@ -0,0 +1,36 @@
...
+            "email": "p@tchwork.com"
+        },
+        {
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "require": {
+        "php": ">=5.3.3",
+        "symfony/polyfill-mbstring": "^1.3"
```
nicolas-grekas: should be removed, that's an optional dep, we don't want to force the polyfill to everyone

IonBazan: @nicolas-grekas I disagree. 
Calling `mb_str_split()` while using PHP 7.3 with `symfony/polyfill-php74` installed and without `ext-mbstring` will cause an error. Until https://github.com/composer/composer/issues/7557 is resolved, there is no better solution than requiring `symfony/polyfill-mbstring` for this package and using `replaces` key in project root when `ext-mbstring` is installed.

nicolas-grekas: The dep is optional, because when you have ext-mbstring, the polyfill is not needed.
The responsibility is on the consumer. If they have to require the polyfill, they can. Or they can require only ext-mbstring.
That's always been our policy.

IonBazan: OK, I'll remove it then. I just seen it in https://github.com/symfony/polyfill/blob/6afbb6c3cfa6ca217c3e07f8aa7306c56b546792/src/Intl/Idn/composer.json#L20 and followed that.

nicolas-grekas: Looks like we missed this during the review in #149, sorry for the confusion it creates now.

stof: what we could do is having `symfony/polyfill-php74` polyfilling `mb_str_split` only if `mb_substr` is available. This way, you get the polyfill for it only if the other parts of mbstring are available (either through the mbstring polyfill or through the native extension).

And `symfony/polyfill-mbstring` should polyfill it all the time

IonBazan: @nicolas-grekas, thank you for explanation :smiley: Removed the polyfill requirement.
@stof Good idea, implemented :+1:

nicolas-grekas:
Thank you @IonBazan.

kamil-tekiela:
FYI there has been multiple workarounds suggested on SO for `mb_str_split()` function already. Maybe you could optimise the case when the split length is 1?

https://stackoverflow.com/a/41664898/1839439
https://stackoverflow.com/a/42122239/1839439
https://stackoverflow.com/a/55783469/1839439

nicolas-grekas:
@kamil-tekiela feel free to submit a PR!

kamil-tekiela:
Done https://github.com/symfony/polyfill/pull/199

This is my first PR, so I am not sure I followed the process correctly or if my PR is useful at all. I appreciate feedback.

src/Php74/Php74.php
```diff
@@ -0,0 +1,90 @@
...
+        if (!\is_object($obj)) {
+            trigger_error('get_mangled_object_vars() expects parameter 1 to be object, '.\gettype($obj).' given', E_USER_WARNING);
+
+            return null;
+        }
+
+        if ($obj instanceof \ArrayIterator || $obj instanceof \ArrayObject) {
+            $reflector = new \ReflectionClass($obj instanceof \ArrayIterator ? 'ArrayIterator' : 'ArrayObject');
+            $flags = $reflector->getMethod('getFlags')->invoke($obj);
+            $reflector = $reflector->getMethod('setFlags');
```
glensc, line 32: not very good coding style to overwrite variable with different purpose (overwriting $reflector here)

makes refactoring difficult

Issue #182, opened 2019-07-30
mb_substr return type


orklah:
Hi,

This library is a dependancy of a library we included on our project.

When our IDE(PHPStorm) try to resolve the mb_substr function, there is two results: [PHPStorm stubs](https://github.com/JetBrains/phpstorm-stubs) and symfony/polyfill.

However, when PHPStorm try to resolve return types for mb_substr on symfony/polyfill, it return string|bool which is not the same as the stubs (string only) nor the php.net documentation (string only).

This is due to that line: [Mbstring.php#L549](https://github.com/symfony/polyfill/blob/master/src/Mbstring/Mbstring.php#L549)

It directly return substr(), which, contrary to mb_substr, can return false on old PHP versions: [php.net/substr](https://php.net/substr#refsect1-function.substr-changelog)

It create issues with PHPStorm static analysis of code because of the extra type resolved for this function.

I created this post to discuss about this. 

If there is grunt work to do (for example, adding PHPDoc to functions), I can create a PR, I wanted to have different point of views before doing so because I don't know if this is an issue that should be dealt with within the code or in PHPDoc (or not at all?)

Thanks

nicolas-grekas:
We could add a test case with an input that make substr return false, then compare the return value with that of the native mb_substr (the test suite will do it automatically), then add a cast to `(string)` after the return statement.

Up for giving it a try?

orklah:
I'll try !

PR #183, opened 2019-07-30
[Mbstring ] fix mb_substr return type

orklah:
WIP. Trying to fix issue #182

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -546,7 +546,7 @@ final class Mbstring
     {
         $encoding = self::getEncoding($encoding);
         if ('CP850' === $encoding || 'ASCII' === $encoding) {
-            return substr($s, $start, null === $length ? 2147483647 : $length);
+            return (string) substr($s, $start, null === $length ? 2147483647 : $length);
         }
 
         if ($start < 0) {
--- a/tests/Mbstring/MbstringTest.php
+++ b/tests/Mbstring/MbstringTest.php
@@ -257,9 +257,13 @@ class MbstringTest extends TestCase
         $this->assertSame("\xFF", mb_substr("\x00\xFF", -1, 1, 'ASCII'));
         $this->assertSame("\x00", mb_substr("\x00\xFF", 0, 1, 'ASCII'));
         $this->assertSame("\x00\xFF", mb_substr("\x00\xFF", 0, 2, 'ASCII'));
+        $this->assertSame('', mb_substr("\x00\xFF", 2, 1, 'ASCII'));
+        $this->assertSame('', mb_substr("\x00\xFF", 3, 1, 'ASCII'));
         $this->assertSame("\xFF", mb_substr("\x00\xFF", -1, 1, 'CP850'));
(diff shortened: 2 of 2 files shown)


orklah:
So, contrary to what I thought when I created #182 , substr can also return false on newer versions of PHP7. The doc only explain that when $start is EQUAL to string length, PHP5 returned false and PHP7 return empty string. When $start is superior, substr continue to return false.

This is not the way mb_substr works. mb_substr always return an empty string when $start is equal or superior to string length

I added two tests to each encoding that happen to call substr internally (ASCII and CP850). 
Every test should fail on PHP5. On PHP7, only the tests with $start longer than the string should fail.

Here is the 3v4l.org link which I used to display my results
https://3v4l.org/QKdXP
(CP850 seems to be missing before PHP5.2.17, but given that this package minimum required version is 5.3.3, I ignored this issue)

I will now begin to modify the mb_substr method to pass the tests

orklah:
The test are passing.

Please tell me if there is something left to do or if there's something I did wrong. This is one of my first PR so I may have made mistakes.

Thanks

src/Mbstring/Mbstring.php
```diff
@@ -546,7 +546,11 @@ public static function mb_substr($s, $start, $length = null, $encoding = null)
     {
         $encoding = self::getEncoding($encoding);
         if ('CP850' === $encoding || 'ASCII' === $encoding) {
-            return substr($s, $start, null === $length ? 2147483647 : $length);
```
nicolas-grekas: `return (string) substr($s, $start, null === $length ? 2147483647 : $length);`

tests/Mbstring/MbstringTest.php
```diff
@@ -257,9 +257,13 @@ public function testSubstr()
         $this->assertSame("\xFF", mb_substr("\x00\xFF", -1, 1, 'ASCII'));
         $this->assertSame("\x00", mb_substr("\x00\xFF", 0, 1, 'ASCII'));
         $this->assertSame("\x00\xFF", mb_substr("\x00\xFF", 0, 2, 'ASCII'));
+        $this->assertSame("", mb_substr("\x00\xFF", 2, 1, 'ASCII'));
+        $this->assertSame("", mb_substr("\x00\xFF", 3, 1, 'ASCII'));
         $this->assertSame("\xFF", mb_substr("\x00\xFF", -1, 1, 'CP850'));
         $this->assertSame("\x00", mb_substr("\x00\xFF", 0, 1, 'CP850'));
         $this->assertSame("\x00\xFF", mb_substr("\x00\xFF", 0, 2, 'CP850'));
+        $this->assertSame("", mb_substr("\x00\xFF", 2, 1, 'CP850'));
+        $this->assertSame("", mb_substr("\x00\xFF", 3, 1, 'CP850'));
```
nicolas-grekas: please use single quotes for the empty string

src/Mbstring/Mbstring.php
orklah: Done :)

tests/Mbstring/MbstringTest.php
orklah: Yes, my bad!

fabpot:
Thank you @orklah.

PR #188, opened 2019-09-15
[grapheme] fix handling negative offsets

nicolas-grekas:
and some cleanups found meanwhile

### DIFF
--- a/src/Intl/Grapheme/Grapheme.php
+++ b/src/Intl/Grapheme/Grapheme.php
@@ -179,10 +179,17 @@ final class Grapheme
         if ($offset > 0) {
             $s = self::grapheme_substr($s, $offset);
         } elseif ($offset < 0) {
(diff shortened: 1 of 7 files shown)


src/Intl/Grapheme/Grapheme.php
```diff
@@ -179,10 +179,17 @@ private static function grapheme_position($s, $needle, $offset, $mode)
         if ($offset > 0) {
             $s = self::grapheme_substr($s, $offset);
         } elseif ($offset < 0) {
-            if (\PHP_VERSION_ID < 50535 || (50600 <= \PHP_VERSION_ID && \PHP_VERSION_ID < 50621) || (70000 <= \PHP_VERSION_ID && \PHP_VERSION_ID < 70006)) {
```
gharlan: @nicolas-grekas why did you remove the version check here? I think this is the cause for the current failing tests.

nicolas-grekas: I don't remember :)

PR #189, opened 2019-09-17
[Mbstring] fix mb_strrpos with negative offset

nicolas-grekas:
Will fix the last failures on https://github.com/symfony/symfony/pull/33553


BackEndTea:
What about `mb_strripos` ?

nicolas-grekas:
it used mb_strrpos, see implementation.

stof:
@nicolas-grekas I still suggest adding the test for it (even though the implementation is already patched by the `mb_strrpos` fix), so that a future refactoring of `mb_strripos` cannot reintroduce the bug

PR #192, opened 2019-10-08
Intl Transliterator - polyfill

voku:
TODO:
- [ ] add more tests - https://github.com/voku/polyfill/blob/intl_transliterator/tests/Intl/Transliterator/TransliteratorTest.php
- [x] not working as expected for PHP <= 5.5
- [ ] some tests did not work on travis-ci (old ICU version)

QUESTIONS:
- [ ] do we need support for "Transliterator->direction"?

reference: https://github.com/voku/portable-ascii/issues/32

---
This change is [Reviewable](https://reviewable.io/reviews/symfony/polyfill/192)


voku:
@nicolas-grekas Hi, here is my first quick try to move the ASCII stuff into a "Transliterator"-polyfill. Maybe you can take a look at this if you have time, thanks.

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,676 @@
...
+        return $s.(null !== $start ? mb_substr($s, $end) : '');
+    }
+    
+    private static function to_translit($s) {
+        if (null === self::$TRANSLIT) {
+            $array = self::getData('translit');
+
+            self::$TRANSLIT = array(
+                'orig' => array_keys($array),
+                'replace' => array_values($array),
```
nicolas-grekas: It would be great to have a source data format that doesn't require any processing, as that would allow using shared memory directly.

src/Intl/Transliterator/bootstrap.php
```diff
@@ -0,0 +1,22 @@
...
+use Symfony\Polyfill\Intl\Transliterator as p;
+
+if (!function_exists('transliterator_create')) {
+    function transliterator_create($id, $direction = null) { return p\Transliterator::create($id, $direction); }
+    function transliterator_create_from_rules($rules, $direction = null) { return p\Transliterator::createFromRules($rules, $direction); }
+    function transliterator_list_ids() { return p\Transliterator::listIDs(); }
+    function transliterator_create_inverse() { return p\Transliterator::createInverse(); }
+    function transliterator_transliterate(p\Transliterator $trans, $subject, $start = null, $end = null) { return $trans->transliterate($subject, $start, $end); }
+    function transliterator_get_error_code(p\Transliterator $trans) { return $trans->getErrorCode(); }
+    function transliterator_get_error_message(p\Transliterator $trans) { return $trans->getErrorMessage(); }
```
nicolas-grekas: this should type-hint for \Transliterator directly, and stubs should exist for the class,
see `src/Intl/Normalizer/Resources/stubs/Normalizer.php` and the classmap in the corresponding composer.json

nicolas-grekas (commented):
Looks nice! Here are some first comments.

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,676 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Intl\Transliterator;
+
+/**
+ * Partial intl implementation in pure PHP.
+ *
+ * Implemented:
```
stof: this phpdoc seems incomplete

src/Intl/Transliterator/data/ascii_by_languages.php
```diff
@@ -0,0 +1,2670 @@
...
+        '៛' => 'KR',
+        '₠' => 'ECU',
+        '¤' => '$?',
+        '฿' => 'THB',
+        '؋' => 'AFN',
+    ),
+);
+
+$result =& $data;
+unset($data);
```
stof: why do we need to replace the variable here ?

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,676 @@
...
+     *
+     * @param string $s <p>The string to check.</p>
+     *
+     * @return bool
+     *              <p>
+     *              <strong>true</strong> if it is ASCII<br>
+     *              <strong>false</strong> otherwise
+     *              </p>
+     */
+    private static function is_ascii($s)
```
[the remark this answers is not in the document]
voku: "is_ascii" is removed and is replaced with this

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,676 @@
...
+
+    /**
+     * Private constructor to deny instantiation.
+     *
+     * @see https://php.net/manual/en/transliterator.construct.php
+     */
+    private function __construct() {
+    }
+
+    public static function create($id, $direction = null) {
```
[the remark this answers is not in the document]
voku: done

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,676 @@
...
+     * @var int
+     */
+    private $direction = self::FORWARD;
+
+    /**
+     * Private constructor to deny instantiation.
+     *
+     * @see https://php.net/manual/en/transliterator.construct.php
+     */
+    private function __construct() {
```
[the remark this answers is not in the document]
voku: done (didn't see that it will not be auto-fixed)

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,676 @@
...
+                normalizer_is_normalized($s, \Normalizer::FORM_D) ?: $s = normalizer_normalize($s, \Normalizer::NFD);
+            } elseif ('NFKD' === $rule) {
+                normalizer_is_normalized($s, \Normalizer::FORM_KD) ?: $s = normalizer_normalize($s, \Normalizer::NFKD);
+            } elseif ('NFKC' === $rule) {
+                normalizer_is_normalized($s, \Normalizer::FORM_KC) ?: $s = normalizer_normalize($s, \Normalizer::NFKC);
+            } elseif ('[:NONSPACING MARK:] REMOVE' === $rule) {
+                $s = preg_replace('/\p{Mn}++/u', '', $s);
+            } elseif (false !== strpos($rule, 'REMOVE')) {
+                $s = preg_replace(self::$REGEX_ASCII, '', $s);
+            } elseif (false !== strpos($rule, 'LATIN-ASCII')) {
```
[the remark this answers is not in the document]
voku: done

src/Intl/Transliterator/Transliterator.php
voku: done

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,676 @@
...
+            } elseif (
+                false !== strpos($rule, '-ASCII')
+                &&
+                $lang = str_replace('-ASCII', '', $rule)
+            ) {
+                $s = self::to_ascii($s, $lang);
+            }
+        }
+
+        return $s.(null !== $start ? mb_substr($s, $end) : '');
```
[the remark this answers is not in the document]
voku: noop, but now it's fixes :-) + tested

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,676 @@
...
+
+    /**
+     * @var int
+     */
+    private $direction = self::FORWARD;
+
+    /**
+     * Private constructor to deny instantiation.
+     *
+     * @see https://php.net/manual/en/transliterator.construct.php
```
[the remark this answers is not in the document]
voku: removed

src/Intl/Transliterator/data/ascii_by_languages.php
voku: I just copy and past the code from the other polyfills. I think in PHP >= 7 we can return the array without this stuff? @nicolas-grekas

src/Intl/Transliterator/bootstrap.php
voku: Now I added a stubs file, but this stubs are not used in other "bootstrap.php"-files? Should we use it here? e.g. ```src/Intl/Normalizer/bootstrap.php```

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,676 @@
...
+    public function transliterate($s, $start = null, $end = null) {
+        if ('' === $s) {
+            return '';
+        }
+
+        if (null !== $start) {
+            $s = mb_substr($s, $start, $end);
+        }
+
+        $s = self::clean($s);
```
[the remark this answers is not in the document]
voku: removed + changed the tests

src/Intl/Transliterator/bootstrap.php
stof: the file should not be required in bootstrap.php. It should be referenced in the composer autoload config

voku: @stof now I use the autoloader + type-hint for \Transliterator directly (in this bootstrap.php) but it's not used this way in other "bootstrap.php" file e.g. 

src/Intl/Normalizer/bootstrap.php


```
use Symfony\Polyfill\Intl\Normalizer as p;

if (!function_exists('normalizer_is_normalized')) {
    function normalizer_is_normalized($s, $form = p\Normalizer::NFC) { return p\Normalizer::isNormalized($s, $form); }
    function normalizer_normalize($s, $form = p\Normalizer::NFC) { return p\Normalizer::normalize($s, $form); }
}
```

stof: that's not a typehint. That's a constant access. So what's important here is the value being used. For the typehint, using the `Symfony\Polyfill\Intl\Normalizer` class rather than the class from the global namespace produces a different function signature.

voku: @stof ok, understood and thanks for the explanation. Is this code correct now?

```php
if (!function_exists('transliterator_create')) {
    function transliterator_create($id, $direction = null) { return \Transliterator::create($id, $direction); }
    function transliterator_create_from_rules($rules, $direction = null) { return \Transliterator::createFromRules($rules, $direction); }
    function transliterator_list_ids() { return \Transliterator::listIDs(); }
    function transliterator_create_inverse() {return \Transliterator::createInverse(); }
    function transliterator_transliterate(\Transliterator $trans, $subject, $start = null, $end = null) { return $trans->transliterate($subject, $start, $end); }
    function transliterator_get_error_code(\Transliterator $trans) { return $trans->getErrorCode(); }
    function transliterator_get_error_message(\Transliterator $trans) { return $trans->getErrorMessage(); }
}
```

src/Intl/Transliterator/Transliterator.php
voku: for now I added only "create" and "transliterate" here

src/Intl/Transliterator/composer.json
```diff
@@ -0,0 +1,34 @@
+{
+    "name": "symfony/polyfill-intl-transliterator",
```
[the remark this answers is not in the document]
voku: done: added ```"symfony/polyfill-intl-transliterator": "self.version",``` into the ```replace``` rule of the root composer.json

src/Intl/Transliterator/bootstrap.php
stof: yes, looks good to me.

voku:
@nicolas-grekas do we want to polyfill support for ICU Rule at all? And if yes do we need to implement all rules? -> https://github.com/symfony/polyfill/pull/192/commits/3e5aef3d59d08ff1ea39c4884ca507c1cc95e33f#diff-ea9e37c915050e124a9d9fcb89522607R136

voku:
@nicolas-grekas @stof can we do something for the old php versions? e.g. PHP 5.4

```diff
-‹ťéśţ› - öäü - 123 - abc - …
+‹��������› - ������ - 123 - abc - …
```

-> https://travis-ci.org/symfony/polyfill/jobs/599397260

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,761 @@
...
+     *
+     * For example, German will map 'ä' to 'ae', while other languages
+     * will simply return e.g. 'a'.
+     *
+     * @psalm-suppress InvalidNullableReturnType - we use the prepare* methods here, so we don't get NULL here
+     *
+     * @param string $language [optional] <p>Language of the source string e.g.: en, de_at, or de-ch.
+     *                         (default is 'en') | ASCII::*_LANGUAGE_CODE</p>
+     *
+     * @return array{orig: string[], replace: string[]}
```
stof: this is a psalm-specific type, which is not valid in phpdoc.

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,761 @@
...
+     *                     <p>An array of replacements.</p>
```
stof: why using HTML tags ?

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,761 @@
...
+        /* @psalm-suppress UnresolvableInclude */
+        return include __DIR__.'/Resources/unidata/'.$file.'.php';
+    }
+
+    /**
+     * Get data from "/data/*.php".
+     *
+     * @param string $file
+     *
+     * @return array|false
```
stof: `array|null` would be better as signature.

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,761 @@
...
+     * e.g.: de_at -> de_at
+     *       de_DE -> de
+     *       DE_DE -> de
+     *       de-de -> de
+     *
+     * @param string $language
+     *
+     * @return string
+     */
+    private static function get_language($language)
```
stof: We use camelCase names in Symfony.

src/Intl/Transliterator/Transliterator.php
```diff
@@ -0,0 +1,761 @@
...
+            )
+        );
+    }
+
+    public static function listIDs()
+    {
+        return array_values(self::$LOCALE_TO_TRANSLITERATOR_ID);
+    }
+
+    private function normalize_regex_helper($rule)
```
stof: using replacement in the rule to convert it to a regex has a big drawback: it could very easily allow injecting whather we want in the regex, breaking everything. I haven't tested this, but I'm almost sure that garbage is rules could go through (at least some of it)

stof:
@voku well, debugging why some garbage is returned when running on such old version ?

src/Intl/Transliterator/Transliterator.php
voku: Jep, I have tried to make this more stable in my last commit. So now we only have one "preg_replace" + one "preg_replace_callback" call and we will check the regex via "@preg_match" before we use it. 

---

Or we only try to polyfill "create" and not "createFromRules" in the first place?

src/Intl/Transliterator/Transliterator.php
voku: fixed

src/Intl/Transliterator/Transliterator.php
voku: now we return always an array here

src/Intl/Transliterator/Transliterator.php
voku: Good question. I added them because I saw them in the PhpStorm stubs and you can easily add something like <strong> ... but anyway, I have removed all html tags with one of my last commit here.

src/Intl/Transliterator/Transliterator.php
voku: fixed

voku:
@stof Garbage was return because old PHP versions do not use "UTF-8" encoding by default. 

But in PHP 5.3 I see this error message: "Fatal error: Class 'Transliterator' not found", is this something like a known issue somehow? Do I need a extra step for this PHP version somehow? -> https://travis-ci.org/symfony/polyfill/jobs/599921509

src/Intl/Transliterator/Transliterator.php
stof: my point is that you should probably not replace in the original string, while letting any unmatched part unchanged, but rather parse the original string and explicitly deal with *all* parts.

A single implementation of replacement does not solve the fact that you could inject stuff between placeholders being replaced.

tests/Intl/Transliterator/TransliteratorTest.php
```diff
@@ -0,0 +1,529 @@
...
+ */
+class TransliteratorTest extends TestCase
+{
+    private $old_mb_internal_encoding;
+
+    public function setUp()
+    {
+        if (function_exists('mb_internal_encoding')) {
+            $this->old_mb_internal_encoding = mb_internal_encoding();
+            mb_internal_encoding('UTF-8');
```
stof: this looks wrong to me. We should not write a polyfill assuming that the global state of mbstring is set to UTF-8, especially if as it is not the default on old PHP versions. We should instead pass an encoding explicitly in our polyfill code.

We cannot expect users of the polyfill to do that change (especially given it is useless for the non-polyfill implementation which is not based on mbstring).

stof (changes requested):

tests/Intl/Transliterator/TransliteratorTest.php
voku: I added the encoding parameter into the polyfill code, so that it now also work for php 5.3.

src/Intl/Transliterator/Transliterator.php
voku: Can you please give me a example? I didn't 100% understand what you mean.

We already split the "rules" (by ";") into chunks, and then normalize the regex and we use only the part in brackets for the regex.

voku: @stof ping

voku: I tried, but the problem is that the syntax from the ICU's regex is not 100% compatible with the PCRE regex, so that it looks like I have to modify the original regex, so that PHP can run it. :-/ Any further idea how to solve this problem?

voku: @ThomasWeinert maybe you can help me with this problem? I try to write a polyfill for " Transliterator", but it use ICU regex. Is there a library that I can relay on or do you have any idea how to handle this?

ThomasWeinert: The clean way would be to implement a small parser for the transliteration rules and compile the PCRE pattern from the tokens/ast. It would allow to throw specific errors.

stof: ICU rules are not a regex, so that's expected that they are not compatible with PCRE. My whole point is that you should not be transforming the ICU rules to a PCRE regex (silently keeping anything that your transformation does not explicitly transforms) but you should parse the rules and generate a PCRE regex, so that nothing can be silently injected in the regex.

voku: @stof ICU rules are using the ICU regular expression syntax, (http://userguide.icu-project.org/strings/regexp) so I think we need an parser for that, otherwise I don't know how to generate a PCRE regex from a ICU regex. And because there are so many options, I don't think that I can simply parse the regex stuff.

@ThomasWeinert Thanks for the feedback. Do you think I can use something like https://github.com/hoaproject/Regex to parse the ICU regex grammar https://github.com/gagolews/stringi/blob/master/src/icu61/i18n/regexcst.txt ?

ThomasWeinert: I used only transliteration **identifiers** before: http://userguide.icu-project.org/transforms/general#TOC-Formal-ID-Syntax

Transliteration **rules** would be for designing your own transliterations: http://userguide.icu-project.org/transforms/general#TOC-Designing-Transliterators 

*btw* http://userguide.icu-project.org/transforms/general#TOC-Regular-Expression

src/Intl/Transliterator/Resources/unidata/ascii_ord.php
```diff
@@ -0,0 +1,266 @@
...
+    "\xFC" => 252,
+    "\xFD" => 253,
+    "\xFE" => 254,
+    "\xFF" => 255,
+);
+
+$result = &$data;
+unset($data);
+
+return $result;
```
glensc, line 266: I'm curious what does this static+reference+unset trick does? what problem does it solve?

```php
static $data = array(
  // ...
);
$result = &$data;
unset($data);

return $result;
```

voku: @glensc It is used in this way for all symfony polyfills, so I think it's something related to PHP OPcache and old PHP versions? Maybe @nicolas-grekas knows why it's written this way?

nicolas-grekas: this used to be more performant under php5, i'm fine removing now

chx:
My knowledge on these matters are limited, however https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/text/Transliterator.html mentions

>  the Transliterator rule syntax is available in section Transform Rules Syntax of UTS #35: Unicode LDML. The rules are shown there using arrow symbols ← and → and ↔. ICU supports both those and the equivalent ASCII symbols < and > and <>.

I have searched for → in this codebase and haven't found anything relevant. And ICU does support the ASCII equivalent, if were to [dump](https://stackoverflow.com/a/17015063/308851) the Transliterator rules `intl` ships with, you'd see they are using → and so I believe for compatibility support for that is necessary as some people might just write their rules following the many examples shipped with PHP.

src/Intl/Transliterator/Transliterator.php
chx: AFAICS this is hopeless. The [Transform Rules Syntax of UTS #35: Unicode LDML](http://www.unicode.org/reports/tr35/tr35-39/tr35-general.html#Transform_Rules_Syntax) is easy to understand for a human (and I am glad it is so because the ICU docs are anything but) but I do not see a formal grammar.

As far as I can see the [reference source code](https://github.com/unicode-org/icu/blob/d3315d98ef82b09aabef12e3f7fb46d171d8bc32/icu4c/source/i18n/rbt_pars.cpp) does _not_ use the regex engine and I can see significant [differences](https://github.com/unicode-org/icu/blob/d3315d98ef82b09aabef12e3f7fb46d171d8bc32/icu4c/source/i18n/rbt_pars.cpp#L647) differences even in matching. And then there's variable definitions and function calls. All this makes me question whether anything else but porting this monstrosity wholesale would achieve full compatibility. In other words, here's the beginning of the Latin-ASCII ruleset 

`:: [[:Latin:][:Common:][:Inherited:][〇]] ;:: NFD() ;`

I think a discussion of what is feasible to support and then sharply limiting the rules to those is the way forward.

nicolas-grekas:
Sorry we didn't catch up on this topic. The amount of work is too high I suppose. Please let me close. Thanks for the PR!

PR #193, opened 2019-10-13
support grapheme_extract() with negative $start

gharlan:
found via https://github.com/symfony/symfony/pull/33971


nicolas-grekas:
Thank you @gharlan.

nicolas-grekas:
Up for having a look at the remaining failure on 5.3? :)

PR #197, opened 2019-10-30
[Uuid] Added the polyfill

lyrixx:


src/Uuid/Uuid.php
```diff
@@ -0,0 +1,274 @@
...
+            return self::UUID_VARIANT_MICROSOFT;
+        }
+
+        return self::UUID_VARIANT_OTHER;
+    }
+
+    public static function uuid_time(string $uuid)
+    {
+        if (PHP_INT_SIZE !== 8) {
+            throw new \RuntimeException('UUID time generation is not supported on 32bits system.');
```
javiereguiluz: `[...] 32bits system.` -> `[...] 32-bit systems.`

nicolas-grekas: should we suggest to install the extension?

src/Util/TestListenerTrait.php
```diff
@@ -27,6 +27,11 @@ public function startTestSuite($mainSuite)
         $SkippedTestError = class_exists('PHPUnit\Framework\SkippedTestError') ? 'PHPUnit\Framework\SkippedTestError' : 'PHPUnit_Framework_SkippedTestError';
 
         foreach ($mainSuite->tests() as $suite) {
+            // We do not want to test the uuid polyfill on PHP < 7
+            // Since this script try to load the SUT, we must skip it
```
nicolas-grekas: s/try/tries

src/Util/TestListenerTrait.php
```diff
@@ -27,6 +27,11 @@ public function startTestSuite($mainSuite)
...
+            if ('Symfony\Polyfill\Tests\Uuid\UuidTest' === $suite->getName() && version_compare('7.0.0', PHP_VERSION, '>=')) {
```
nicolas-grekas: PHP_VERSION_ID?

src/Uuid/Uuid.php
lyrixx: Thanks. I added the suggestion

src/Util/TestListenerTrait.php
```diff
@@ -109,8 +114,10 @@ function {$f['name']}{$f['signature']}
                 $mainSuite->addTest(new TestListener($suite));
             }
         }
-        foreach ($warnings as $w) {
-            $mainSuite->addTest($w);
+        if (isset($warnings)) {
+            foreach ($warnings as $w) {
```
dunglas: `$warning ?? []`? (nitpicking)

nicolas-grekas: Hello PHP 5.3 :)

src/Util/TestListenerTrait.php
```diff
@@ -27,6 +27,11 @@ public function startTestSuite($mainSuite)
         $SkippedTestError = class_exists('PHPUnit\Framework\SkippedTestError') ? 'PHPUnit\Framework\SkippedTestError' : 'PHPUnit_Framework_SkippedTestError';
 
         foreach ($mainSuite->tests() as $suite) {
+            // We do not want to test the uuid polyfill on PHP < 7
+            // Since this script tries to load the SUT, we must skip it
+            if ('Symfony\Polyfill\Tests\Uuid\UuidTest' === $suite->getName() && PHP_VERSION_ID < 70000) {
```
stof: can we implemenent that in a more generic way by looking at the `@requires php` annotations on the test class instead ?

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,274 @@
...
+        if (($parsed['clock_seq'] & 0x2000) == 0) {
+            return self::UUID_VARIANT_MICROSOFT;
+        }
+
+        return self::UUID_VARIANT_OTHER;
+    }
+
+    public static function uuid_time(string $uuid)
+    {
+        if (PHP_INT_SIZE !== 8) {
```
stof: should be fully-qualified to be resolved at compile time and let OPCache optimize dead code due to this condition (same reason that for our PHP version checks)

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,274 @@
...
+            // four most significant bits holds version number 4
+            random_int(0, 0x0fff) | 0x4000,
+
+            // 16 bits:
+            // * 8 bits for "clk_seq_hi_res",
+            // * 8 bits for "clk_seq_low",
+            // two most significant bits holds zero and one for variant DCE1.1
+            random_int(0, 0x3fff) | 0x8000,
+
+            // 48 bits for "node"
```
stof: will this work on 32-bits PHP ?

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,274 @@
...
+        );
+    }
+
+    private static function uuid_parse_as_array(string $uuid)
+    {
+        if (36 != \strlen($uuid)) {
+            return null;
+        }
+
+        static $regex = '{^(?<time_low>[0-9a-f]{8})-(?<time_mid>[0-9a-f]{4})-(?<time_hi_and_version>[0-9a-f]{4})-(?<clock_seq>[0-9a-f]{4})-(?<node>[0-9a-f]{12})$}i';
```
stof: do we even need static variable, or should we inline the regex ?

src/Util/TestListenerTrait.php
stof: btw, why do we need to make that filtering here ? Is it only because the polyfill implementation uses PHP 7 typehints ? If yes, it might make more sense to drop them (especially given that the polyfilled functions don't have them).

lyrixx: > can we implemenent that in a more generic way by looking at the @requires php annotations on the test class instead ?

We could, but there is already some harcoded check in this file. More over we don't have an annotation reader in this project.But maybe phpunit expose that information. If it's the case, I will update the code. If not, I prefer to keep something simple

> Is it only because the polyfill implementation uses PHP 7 typehints ? 

yes I guess

> If yes, it might make more sense to drop them (especially given that the polyfilled functions don't have them).

If I do that, I will have to manually check the argument type by hand to have the same behavior than the ext.

src/Uuid/Uuid.php
lyrixx: I don't know. How could I test it ?

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,274 @@
...
+        );
+
+        if (null === self::uuid_parse_as_array($uuid)) {
+            return false;
+        }
+
+        return $uuid;
+    }
+
+    private static function uuid_generate_random(): string
```
stof: isn't it possible for this implementation to generate a null UUID with a non-zero probability ? And is it the same for the extension ?

src/Uuid/Uuid.php
lyrixx: I did it for readability, but indeed :+1: I will move it

src/Uuid/Uuid.php
lyrixx: a null UUID is still a string. Did I miss something?

src/Util/TestListenerTrait.php
stof: does the ext actually throw TypeError or warnings ?

src/Uuid/Uuid.php
stof: well, the question is whether `uuid_generate_random()` is allowed to return this null UUID in the original implementation, or whether it never returns it (given that a null UUID won't be of type `UUID_TYPE_RANDOM` AFAICT)

src/Util/TestListenerTrait.php
lyrixx: It seems it triggers warning: 
```
$ php -a
php > uuid_create('a');
PHP Warning:  uuid_create() expects parameter 1 to be int, string given in php shell code on line 1

Warning: uuid_create() expects parameter 1 to be int, string given in php shell code on line 1
```

stof: then the polyfill must trigger warnings too

src/Uuid/Uuid.php
lyrixx: It looks like nothing [prevent](https://github.com/karelzak/util-linux/blob/6a735c2c5bfd64b06fd210663c1ebfeac6eef061/libuuid/src/gen_uuid.c#L502-L523) the extension to return a null UUID

**BUT** This is not possible since we "pad" some value for ensure the type:
```
            // 16 bits for "time_hi_and_version",
            // four most significant bits holds version number 4
            random_int(0, 0x0fff) | 0x4000,
```

src/Uuid/Uuid.php
lyrixx: Fixed

src/Uuid/Uuid.php
lyrixx: Fixed. Thanks

src/Uuid/Uuid.php
lyrixx:
```
php > echo random_int(0, 0xffffffffffff);
PHP Warning:  Uncaught TypeError: random_int() expects parameter 2 to be integer, float given in php shell code:1
Stack trace:
#0 php shell code(1): random_int(0, 2.8147497671066E+14)
#1 {main}
```

So I updated this :)

src/Util/TestListenerTrait.php
lyrixx: OK, I used PHPUnit groups to fixed this issue. I was not able to retrieve requirements :(

lyrixx:
I think this PR is ready :tada:

src/Util/TestListenerTrait.php
stof: well, if you change the polyfill to properly trigger warnings rather than TypeErrors, you might not need this at all anymore (as it will become compatible with PHP 5.x too)

stof:
based on https://github.com/symfony/polyfill/pull/197#discussion_r342492555 (and my answer just after), it is not, as the error handling is not the same in the polyfill and in the original extension.

lyrixx:
OK, I removed all the type hints and codes incompatible with PHP > 5.3
Tests run on PHP 5.4+. They don't run on PHP 5.3 because the PECL does not compile.

Should be really ready now :smile:

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,349 @@
...
+    public static function uuid_create($uuid_type = UUID_TYPE_DEFAULT)
+    {
+        if (!\is_int($uuid_type)) {
+            trigger_error(sprintf("uuid_create() expects parameter 1 to be int, %s given", gettype($uuid_type)), E_USER_WARNING);
+
+            return null;
+        }
+
+        switch ($uuid_type) {
+            case UUID_TYPE_NAME:
```
jderusse: should be `self::UUID_TYPE_NAME` or move constantes outside the class

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,349 @@
...
+                    goto static_node;
+                }
+            }
+        } else {
+            static_node:
+            static $node;
+            if (null === $node) {
+                $node = random_int(0, 0xffffffffffff);
+            }
+        }
```
jderusse: Wouldn't it be faster to do avoid calling `function_exists` and `apcu_fetch` each time we generate a uuid?
```php
        static $node;
        if (null === $node) {
            if (\function_exists('apcu_fetch')) {
                $node = apcu_fetch('__symfony_uuid_node');
                if (false === $node) {
                    $node = random_int(0, 0xffffffffffff);
                    apcu_store('__symfony_uuid_node', $node);
                }
            } else {
                $node = random_int(0, 0xffffffffffff);
        }

```

src/Uuid/Uuid.php
lyrixx: It works because the polyfill was already loaded.
But indeed, if one use the class directly (this is forbidden by `@internal` tag) it will crash.

But you are right, I will update this

src/Uuid/Uuid.php
lyrixx: Much better (and bye bye my goto :cry: )

lyrixx: Note: I also fixed the support for 32-bits Sytems

src/Uuid/Uuid.php
```diff
@@ -0,0 +1,352 @@
...
+        // not able to get a super precise timeOfDay as a unique sequence
+        $clockSeq = random_int(0, 0x3fff);
+
+        static $node;
+        if (null === $node) {
+            if (\function_exists('apcu_fetch')) {
+                $node = apcu_fetch('__symfony_uuid_node');
+                if (false === $node) {
+                    $node = sprintf('%06x%06x',
+                        random_int(0, 0xffffff),
```
xelan: First of all, thanks for the nice polyfill. Having UUID generation directly intergrated is great :+1:

Just a small note for the node value: According to [RFC4122](https://tools.ietf.org/html/rfc4122#section-4.1.6), the MAC address [multicast bit](https://www.geeksforgeeks.org/introduction-of-mac-address-in-computer-network/) must be set if randomly or pseudo-randomly generated node values are used.

lyrixx: Hello @xelan

Thanks :) 

 What does that mean exactly? What is the mylticasst bit?

xelan: Hi @lyrixx! Thanks for your prompt response :) According to the RFC, the UUIDs generated with random node values must be distinguishable from regular MACs via setting that bit  to 1 (LSB of first MAC octet according to the linked source). Probably it would also make sense to set the locally administered bit to 1, as the random node values are not supposed to be globally unique.

xelan: [Here ](https://tools.ietf.org/html/rfc4122#section-4.5)the reasoning behind that requirement is described. Based on that, I'd guess that setting both the multicast bit and the local bit would be the best solution.

lyrixx: done, Thanks

src/Util/TestListenerTrait.php
```diff
@@ -109,8 +109,10 @@ function {$f['name']}{$f['signature']}
                 $mainSuite->addTest(new TestListener($suite));
             }
         }
-        foreach ($warnings as $w) {
-            $mainSuite->addTest($w);
+        if (isset($warnings)) {
```
stof: is this change still necessary, given that other changes in this file have been reverted ?

lyrixx: reverted

nicolas-grekas:
Thank you @lyrixx.

PR #198, opened 2019-11-08
Compatibility fix for PHP 7 and higher for polyfill-php70 package

MykhailoSukovitsyn:
symfony/polyfill-php70 package, as composer.json says, requires PHP >=5.3.3 but on practice if my environment uses PHP >= 7.0.0, name conflict occurs. PR fixes this issue.


stof:
the autoloader won't ever try to load these files if the classes already exist in PHP. How did you trigger these name conflicts ?

nicolas-grekas:
What @stof said. If this creates an issue on your use case, please check the tool that loads these files and report the bug there.

PR #199, opened 2019-11-10
Optimization of mb_str_split

kamil-tekiela:
`mb_strlen` + `mb_substr` can be very slow for large strings. In the common case of split length 1 and charset UTF=8 we can use `preg_split` instead.

Added an early exit to the function:

    if (1 === $split_length && 'UTF-8' === $encoding) {
        return preg_split('//u', $string, null, PREG_SPLIT_NO_EMPTY);
    }

See https://3v4l.org/U6PPf

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -534,7 +534,7 @@ final class Mbstring
             return null;
         }
 
-        if ($split_length < 1) {
+        if (1 > $split_length = (int) $split_length) {
             trigger_error('The length of each segment must be greater than zero', E_USER_WARNING);
 
             return false;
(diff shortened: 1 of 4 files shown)


BackEndTea:
This probably needs a test case with exactly this combo (split of 1, and 'UTF-8'encoding)

kamil-tekiela:
The first test kind of handles this, because 1 and UTF-8 should be the default values. 

We could add the following test:

    $this->assertSame(array('e', '́', '💩', '𐍈'), **mb_str_split('é💩𐍈', 1, 'UTF-8'));**

BackEndTea:
I'd say  👍 for a more explicit test for this case

voku:
```preg_match_all``` seems to be faster, or?

-> https://3v4l.org/8XPK7 (preg_match_all)
-> https://3v4l.org/Dl2ml (preg_split)

PS: I think for short strings it's even faster to use substr ? e.g. https://github.com/voku/portable-utf8/commit/e70be3849f826fb63299a9bb4a5fa4581c53fd9a#diff-890805f35966ea20340b9076d7b47329R7212

nicolas-grekas:
Thank you @kamil-tekiela.

PR #200, opened 2019-11-26
PHP 8.0 - add fdiv() function polyfill

IonBazan:
This PR adds simple `fdiv()` implementation by suppressing warnings on the division, just like in original implementation: https://github.com/php/php-src/commit/e35bdb49121ac451384b6c443850ce67e7c5be63

This feature works only on PHP 7+ because suppressing warnings does not produce valid results and casting to string does not detect negative zero on older versions. 
Full logic implementation is available at https://github.com/IonBazan/polyfill/blob/php8-php5/src/Php80/Php80.php but lacks negative-zero detection for PHP 5. If anyone figures out how to detect it, we can simply replace `LogicException` with it.

### DIFF
--- a/src/Php80/Php80.php
+++ b/src/Php80/Php80.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Php80;
+
+/**
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
+ * @internal
+ */
+final class Php80
+{
+    public static function fdiv($dividend, $divisor)
+    {
+        $dividend = self::floatArg($dividend, __FUNCTION__, 1);
+        $divisor = self::floatArg($divisor, __FUNCTION__, 2);
+
+        return (float) @($dividend / $divisor);
+    }
+
+    private static function floatArg($value, $caller, $pos)
+    {
+        if (\is_float($value)) {
+            return $value;
+        }
+
+        if (!\is_numeric($value)) {
+            throw new \TypeError(sprintf('%s() expects parameter %d to be float, %s given', $caller, $pos, \gettype($value)));
+        }
+
+        return (float) $value;
+    }
+}
--- a/src/Php80/bootstrap.php
+++ b/src/Php80/bootstrap.php
@@ -1,0 +1,18 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Php80 as p;
+
+if (PHP_VERSION_ID < 80000 && PHP_VERSION_ID >= 70000) {
+    if (!function_exists('fdiv')) {
+        function fdiv($dividend, $divisor) { return p\Php80::fdiv($dividend, $divisor); }
+    }
+}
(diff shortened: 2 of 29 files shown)


src/Php80/Php80.php
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
+ * @internal
+ */
+final class Php80
+{
+    public static function fdiv($dividend, $divisor)
+    {
+        if (PHP_VERSION_ID >= 70000) {
```
stof: please use a fully-qualified name for this constant, so that OPCache resolves it at compile-time and drops dead code.

src/Php80/composer.json
```diff
@@ -0,0 +1,35 @@
...
+        "php": ">=5.3.3"
+    },
+    "autoload": {
+        "psr-4": { "Symfony\\Polyfill\\Php80\\": "" },
+        "files": [ "bootstrap.php" ]
+    },
+    "minimum-stability": "dev",
+    "extra": {
+        "branch-alias": {
+            "dev-master": "1.12-dev"
```
stof: As this adds a new feature, the dev version needs to be bumped to 1.13 everywhere.

src/Php80/composer.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "symfony/polyfill-php80",
```
stof: this must be added in the `replace` rule of the root package.

src/Php80/Php80.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Symfony\Polyfill\Php80;
+
+/**
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
+ * @internal
+ */
+final class Php80
+{
+    public static function fdiv($dividend, $divisor)
```
stof: what about the error handling for invalid types ?

stof:
@IonBazan your idea based on `var_dump` will now work when xdebug overloads it (and it might not work on Windows due to line endings, but that's easier to solve).

src/Php80/composer.json
nicolas-grekas: 1.14 actually now

IonBazan:
@nicolas-grekas @stof changes applied. I decided to drop PHP 5 because found it impossible to detect negative zeros on it.

src/Php80/composer.json
```diff
@@ -0,0 +1,35 @@
...
+            "name": "Nicolas Grekas",
+            "email": "p@tchwork.com"
+        },
+        {
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "require": {
+        "php": ">=5.3.3"
```
BackEndTea: If this needs php 7 to function correctly, then should this be bumped as well?

src/Php80/bootstrap.php
```diff
@@ -0,0 +1,18 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Php80 as p;
+
+if (PHP_VERSION_ID < 80000) {
+    if (!function_exists('fdiv')) {
```
stof: As the polyfill is not compatible with PHP 5 (while the main package is), I think we should add a version check here. Not registering the polyfilled function is better than registering it but throwing an exception when using it. Code checking for `function_exists('fdiv')` would break if we register a non-working polyfill.

src/Php80/composer.json
IonBazan: That would prevent you from installing `symfony/polyfill` meta package on PHP 5.

nicolas-grekas: PHP 7 here yes. The root json file is a separate one so no issue with it.

IonBazan: @nicolas-grekas I bumped the version but I wonder if we should do that now. I'm planning to submit another PR with `ValueError` class which could potentially be used on PHP 5 using `symfony/polyfill-php70` which includes `*Error` classes.

nicolas-grekas:
Thank you @IonBazan.

PR #187, opened 2019-08-25
Ensure hrtime is monotonic

GrahamCampbell:
Closes #186.

PR #202, opened 2019-11-27
[UUID] Fixed tests when the PECL is not available

lyrixx:
fixes https://github.com/symfony/polyfill/pull/187#issuecomment-559123404


nicolas-grekas:
Thank you @lyrixx.

PR #203, opened 2019-11-27
Revert "bug #187 Ensure hrtime is monotonic (GrahamCampbell)"

nicolas-grekas:
This reverts commit 64d2d7d5b4595443c1ad0c667597ea4f5ae1e7a5, reversing
changes made to b501e55fbc45de6bc4112efeb7cb5f4b6a283d7b.

/cc @GrahamCampbell FYI I just realized the implementation doesn't work:
microtime(false) return something like `0.00018200 1574780469`
casting this to float and comparing with the previous checks only the sub-seconds part. But it should account for seconds too.
That's why tests fail transiently on the CI.

Maybe we don't care about monotonicity eventually, do we?

### DIFF
--- a/src/Php73/Php73.php
+++ b/src/Php73/Php73.php
@@ -21,8 +21,6 @@ final class Php73
 {
     public static $startAt = 1533462603;
 
-    private static $previousValue = '0';
-
     /**
      * @param bool $asNum
      *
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
> Maybe we don't care about monotonicity eventually, do we?

I mean, yes. That's the only real advantage of this function compared to microtime?

nicolas-grekas:
That's correct :) But nobody will notice the diff in practice - or they should upgrade if they care, don't you think?

GrahamCampbell:
Hmmm. Well, I've experienced lots of issues with the time on my local machine shifting causing timing results from symfony to be negative. If I run a script for an hour straight, I always always see a time shift cause a negative timing. Symfony's timing code should really be using hrtime, and the hrtime poylfill should at least prevent negative timings.

nicolas-grekas:
OK, if you encountered the issue, then it's legit fixing it.
New PR welcome :)

PR #204, opened 2019-11-29
Make the UUID polyfill compatible with PHP >=5.3.3

javiereguiluz:


nicolas-grekas:
Thank you @javiereguiluz.

BackEndTea:
The uuid implementation uses `random_bytes`, which is only present in php 7+, or with the polyfill. So this project should then require the `random_compat` polyfill.

The reason the tests pass anyways is probably because one of the other polyfills has it as a dependency.

lyrixx:
@BackEndTea It seems you are right. Would you mind submitting a PR for that ?

BackEndTea:
@lyrixx I did in #207

PR #205, opened 2019-11-29
[UUID] Fix undefined variable in error message, and other minor fixes

guilliamxavier:
(all commits are actually independent)


nicolas-grekas:
Thank you @guilliamxavier.

PR #212, opened 2019-12-02
[Mbstring] fix `mb_str_split()` ignoring new-line characters

### LINKED ISSUE #211: mb_str_split() bug when string contains newline
I stumbled upon a bug where mb_str_split() returned an odd response. After some research I found out it happened when input string has one or more newlines.

Example code to reproduce:
```
$str = "Hello!\nThis is an example with a string with new line.\nI'm just making an example to point out how mb_str_split() was broken in 1.13.0 when string has new lines. Thanks for reading this long example string! Have a nice day!";

var_dump(mb_strlen($str));
var_dump(mb_str_split($str, 153));
```

Expected output (https://3v4l.org/L6M1K):
```
int(223)
array(2) {
  [0]=>
  string(153) "Hello!
This is an example with a string with new line.
I'm just making an example to point out how mb_str_split() was broken in 1.13.0 when string has ne"
  [1]=>
  string(70) "w lines. Thanks for reading this long example string! Have a nice day!"
}
```

The problem started in 1.13.0 where https://github.com/symfony/polyfill/pull/199 was merged.
If I understand the intention of the pull request correctly he wanted to return early when split length is 1, his initial example was:
```
if (1 === $split_length && 'UTF-8' === $encoding) {
    return preg_split('//u', $string, null, PREG_SPLIT_NO_EMPTY);
}
```

Actual output with polyfill 1.13.0+:
```
int(223)
array(3) {
  [0]=>
  string(55) "Hello!
This is an example with a string with new line.
"
  [1]=>
  string(153) "I'm just making an example to point out how mb_str_split() was broken in 1.13.0 when string has new lines. Thanks for reading this long example string! H"
  [2]=>
  string(15) "ave a nice day!"
}
```

The problem seems to be that in the final commit you are not checking if $split_length is actually 1, which results in that preg_split() is always used:

https://github.com/symfony/polyfill/pull/199/commits/8dd88d7792f07487f7ced96be1a6b06cd4f4d745#diff-e8706b247d3e17dac17a413b65b64193R547

nicolas-grekas:
Closes #211

### DIFF
--- a/tests/Mbstring/MbstringTest.php
+++ b/tests/Mbstring/MbstringTest.php
@@ -318,6 +318,8 @@ class MbstringTest extends TestCase
      */
     public function testStrSplit()
     {
+        $this->assertSame(array('H', "\r", "\n", 'W'), mb_str_split("H\r\nW", 1));
+        $this->assertSame(array('Hell', "o\nWo", 'rld!'), mb_str_split("Hello\nWorld!", 4));
         $this->assertSame(array('한', '국', '어'), mb_str_split('한국어'));
         $this->assertSame(array('по', 'бе', 'да'), mb_str_split('победа', 2));
         $this->assertSame(array('źre', 'bię'), mb_str_split('źrebię', 3));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 * v1.13.2
 
  * use correct block size for SHA1 in `hash_pbkdf2()` polyfill
+ * fixed `mb_str_split()` ignoring new-line characters
+ * updated Unicode maps
 
 * v1.13.1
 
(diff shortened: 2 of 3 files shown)


tests/Mbstring/MbstringTest.php
```diff
@@ -318,6 +318,7 @@ public function testStrposNegativeOffset()
      */
     public function testStrSplit()
     {
+        $this->assertSame(array('Hell', "o\nWo", 'rld!'), mb_str_split("Hello\nWorld!", 4));
```
IonBazan: Let's test `\r\n` too.

guilliamxavier:
Thanks but I cannot get the fix by updating `symfony/polyfill-mbstring` (keeps at `v1.13.1`), it seems that the `v1.13.2` release tag of `symfony/polyfill` has not been applied to individual sub-packages

fabpot:
@guilliamxavier Fixed now

PR #213, opened 2019-12-03
Update Unicode maps

nicolas-grekas:
Updated using https://github.com/tchwork/utf8/blob/master/src/Patchwork/Utf8/Compiler.php


stof:
should we bring this updater tooling in this repo (or maybe at least a script doing that, with a dev dependency on this compiler) ?

nicolas-grekas:
The CS of the compiler class is not up to the standards of this repo. I won't be able to port it here is the short term.

PR #217, opened 2020-01-13
Fix nighlty builds

nicolas-grekas:


nicolas-grekas:
Tests on PHP8 are not green, but at least they run.

stof:
Most of the failures are related to PHP 8 converting warnings to TypeError for bad arguments.

PR #221, opened 2020-01-17
Raise minimum polyfill-php72 version in polyfill-intl-idn

### LINKED ISSUE #220: idn-polyfill is not working for php < 7.2 without the symfony mb-polyfill
I saw today a very strange error, because of using php 7.1 + this exclusion in the composer config...

```
{
    "replace": {
        "symfony/polyfill-php70": "1.99",
        "symfony/polyfill-php71": "1.99",
        "symfony/polyfill-mbstring": "1.99"
    }
}
```

... another library required ```symfony/polyfill-idn``` but that is using ```mb_ord()``` (=> PHP 7.2). This was only a local problem, so we switched the php version and everything was good. But I wanted to create an issue here, so that someone else can find it, if he/she had the same issue.

jdreesen:
Fixes #220

### DIFF
--- a/src/Intl/Idn/composer.json
+++ b/src/Intl/Idn/composer.json
@@ -18,7 +18,7 @@
     "require": {
         "php": ">=5.3.3",
         "symfony/polyfill-mbstring": "^1.3",
-        "symfony/polyfill-php72": "^1.9"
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Thank you @jdreesen.

PR #224, opened 2020-01-25
Add polyfill for PHP8's Stringable

nicolas-grekas:
- [x] Should wait for https://wiki.php.net/rfc/stringable to be approved before merging.

### DIFF
--- a/src/Php80/Resources/stubs/Stringable.php
+++ b/src/Php80/Resources/stubs/Stringable.php
@@ -1,0 +1,9 @@
+<?php
+
+interface Stringable
+{
+    /**
+     * @return string
+     */
+    public function __toString();
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 1.15.0
 
   * added `preg_last_error_msg()` to the PHP 8 polyfill
+  * added interface `Stringable` to the PHP 8 polyfill
 
 # 1.14.0
 
--- a/src/Php80/README.md
+++ b/src/Php80/README.md
@@ -1,8 +1,9 @@
 Symfony Polyfill / Php80
 ========================
 
-This component provides functions added to PHP 8.0 core:
+This component provides features added to PHP 8.0 core:
 
+- `Stringable` interface
 - [`fdiv`](https://php.net/fdiv)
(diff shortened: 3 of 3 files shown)


src/Php80/Resources/stubs/Stringable.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+interface Stringable
+{
+    /**
+     * @return string
+     */
+    public function __toString();
```
BackEndTea: I think this needs a `: string` to be compatible with the RFC right?

nicolas-grekas: Nope, it doesn't - and adding it would force a BC break on code that wants to adopt the polyfill. See the RFC, this is discussed.

nicolas-grekas: This one runs on php7, the stub on php8, where the BC layer is in place. So yes, you're misunderstanding something :)

ohartl: > This one runs on php7, the stub on php8, where the BC layer is in place. So yes, you're misunderstanding something :)

this ensured by the test `Zend/tests/stringable_automatic_implementation.phpt`. https://github.com/php/php-src/pull/5083/files#diff-f3a64c6266183080afe34570b2b90282

@nicolas-grekas does `symfony/polyfill` already test compliance with the `php/php-src` Test Suite? It should be to integrate, partial usage of their test suite to cover `symfony/polyfill` easy to run those tests over the ployfill code. Let me know if this is worth a shot and i could help.

nicolas-grekas: We don't do it systematically. In this specific case, the behavior on php8 cannot be replicated on php7. But that does invalidate the polyfill, which still works, in an opt in way.

filipeoliveira97, line 8: public function __toString() : string;

nicolas-grekas: missing return type on purpose, see https://wiki.php.net/rfc/stringable:
> code in PHP < 8 can adopt a polyfill interface immediately (one that doesn't declare the return type for the same BC reasons.)

PR #226, opened 2020-02-05
[Php80] Add get_debug_type()

nicolas-grekas:
Waiting for
- [ ] https://github.com/php/php-src/pull/5143
- [ ] https://wiki.php.net/rfc/get_debug_type
- [x] php/php-src#5153

### DIFF
--- a/src/Php80/Php80.php
+++ b/src/Php80/Php80.php
@@ -14,6 +14,7 @@ namespace Symfony\Polyfill\Php80;
 /**
  * @author Ion Bazan <ion.bazan@gmail.com>
  * @author Nico Oelgart <nicoswd@gmail.com>
+ * @author Nicolas Grekas <p@tchwork.com>
  *
  * @internal
  */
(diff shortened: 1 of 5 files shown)


src/Php80/bootstrap.php
```diff
@@ -19,4 +19,8 @@ function fdiv($dividend, $divisor) { return p\Php80::fdiv($dividend, $divisor);
     if (!defined('FILTER_VALIDATE_BOOL')) {
         define('FILTER_VALIDATE_BOOL', FILTER_VALIDATE_BOOLEAN);
     }
+
+    if (!function_exists('get_debug_type')) {
+        function get_debug_type($value) { return is_object($value) ? get_class($value) : gettype($value); }
```
BackEndTea: I don't think `gettype` is correct, since `get_debug_type` will return `int` instead of `integer`, looking at the implementation.

nicolas-grekas:
Now in sync with latest changes, including https://github.com/php/php-src/pull/5153

src/Php80/Php80.php
```diff
@@ -18,12 +18,35 @@
...
+            case \is_int($value): return 'int';
+            case \is_float($value): return 'float';
+            case \is_object($value): break;
+            case $value instanceof \__PHP_Incomplete_Class: return '__PHP_Incomplete_Class';
+            default: return 'resource';
+        }
+
+        $class = \get_class($value);
+
+        if ("\0" !== ($class[15] ?? '')) {
```
guilliamxavier: I guess that `15` magic number is for `\strlen('class@anonymous')`? Will it need to be changed after php/php-src#5153?

nicolas-grekas:
Review addressed, implem synced with the PR on php-src.

PR #227, opened 2020-02-05
Check if `FILTER_VALIDATE_BOOLEAN` is available

BackEndTea:
PHP can be compiled without the `filter` extension, so there is a possibility that `FILTER_VALIDATE_BOOLEAN` isn't available.


fabpot:
Thank you @BackEndTea.

BackEndTea:
Thank you @fabpot, @IonBazan & @nicolas-grekas

PR #229, opened 2020-02-18
[PHP 8.0] add `str_contains` function

IonBazan:
Waiting for https://wiki.php.net/rfc/str_contains RFC.

### DIFF
--- a/src/Php80/Php80.php
+++ b/src/Php80/Php80.php
@@ -45,4 +45,9 @@ final class Php80
                 return 'Unknown error';
         }
     }
+
+    public static function str_contains(string $haystack, string $needle): bool
+    {
+        return '' === $needle || false !== strpos($haystack, $needle);
+    }
 }
--- a/tests/Php80/Php80Test.php
+++ b/tests/Php80/Php80Test.php
@@ -87,6 +87,24 @@ class Php80Test extends TestCase
         );
     }
 
+    /**
+     * @covers \Symfony\Polyfill\Php80\Php80::str_contains
+     */
+    public function testStrContains()
+    {
+        $this->assertTrue(str_contains('abc', ''));
+        $this->assertTrue(str_contains('abc', 'a'));
+        $this->assertTrue(str_contains('abc', 'bc'));
+        $this->assertTrue(str_contains('abc', 'abc'));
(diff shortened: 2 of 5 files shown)


src/Php80/Php80.php
```diff
@@ -26,6 +26,11 @@ public static function fdiv($dividend, $divisor)
         return (float) @($dividend / $divisor);
     }
 
+    public static function str_contains($haystack, $needle)
```
nicolas-grekas: please add all type declarations

IonBazan: Are you sure? We are not using them in other functions yet.

IonBazan:
Rebased 🚀

src/Php80/Php80.php
nicolas-grekas: See #234

src/Php80/Php80.php
```diff
@@ -49,6 +49,11 @@ public static function preg_last_error_msg()
         }
     }
 
+    public static function str_contains($haystack, $needle)
+    {
+        return '' === $needle || false !== \strpos($haystack, $needle);
```
nicolas-grekas: no need for the `\`

src/Php80/Php80.php
nicolas-grekas: (please rebase again while doing so)

tests/Php80/Php80Test.php
```diff
@@ -87,6 +87,24 @@ public function testPregMalformedUtf8Offset()
...
+    /**
+     * @covers \Symfony\Polyfill\Php80\Php80::str_contains
+     */
+    public function testStrContains()
+    {
+        $this->assertTrue(str_contains('abc', ''));
+        $this->assertTrue(str_contains('abc', 'a'));
+        $this->assertTrue(str_contains('abc', 'bc'));
+        $this->assertTrue(str_contains('abc', 'abc'));
+        $this->assertTrue(str_contains('abc', 'abc'));
```
noniagriconomie: Twice same test here

src/Php80/Php80.php
IonBazan: OK, typehints added, changes rebased again 😉

nicolas-grekas:
Thank you @IonBazan.

PR #231, opened 2020-02-25
[PHP 8.0] Add preg_last_error_msg function

nicoSWD:
This PR adds the function `preg_last_error_msg` added in [php-src/pull/5185](https://github.com/php/php-src/pull/5185)

_Note_: I only tested the cases that don't require special `.ini` settings as I didn't want to set them dynamically in the test suite.


nicolas-grekas:
Thank you @nicoSWD.

PR #240, opened 2020-03-10
[uuid] provide compat with 32bit systems

nicolas-grekas:
So that we can pretend to have a truely portable polyfill.

### DIFF
--- a/src/Uuid/Uuid.php
+++ b/src/Uuid/Uuid.php
@@ -32,6 +32,13 @@ final class Uuid
     const UUID_TYPE_NULL = -1;
     const UUID_TYPE_INVALID = -42;
 
+    // https://tools.ietf.org/html/rfc4122#section-4.1.4
+    // 0x01b21dd213814000 is the number of 100-ns intervals between the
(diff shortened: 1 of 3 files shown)


src/Uuid/Uuid.php
```diff
@@ -360,16 +364,18 @@ private static function uuid_generate_random()
...
-        if (\PHP_INT_SIZE === 4) {
-            throw new \RuntimeException('UUID time generation is not supported on 32-bit systems. Use the uuid extension instead.');
-        }
+        $time = microtime(false);
+        $time = substr($time, 11).substr($time, 2, 7);
 
-        // https://tools.ietf.org/html/rfc4122#section-4.1.4
-        // 0x01b21dd213814000 is the number of 100-ns intervals between the
-        // UUID epoch 1582-10-15 00:00:00 and the Unix epoch 1970-01-01 00:00:00.
         $offset = 0x01b21dd213814000;
```
lyrixx: This can be removed, you don't use it anymore.
and I would move this special value to a constant since you are using 3 times (and move the doc on top of the constant)

nicolas-grekas: addressed.

PR #241, opened 2020-03-13
[Uuid] fix edge cases

nicolas-grekas:


src/Uuid/Uuid.php
```diff
@@ -435,6 +429,15 @@ private static function uuid_parse_as_array($uuid)
         );
     }
 
+    private static function toString($v)
```
stof: the name is confusing, as it does not always return a string.

nicolas-grekas: I get that - naming things... Feel free to submit a PR...

PR #244, opened 2020-03-20
[uuid] improve perf

nicolas-grekas:
Follows https://dev.to/fredbouchery/performance-des-uuids-29dj

### DIFF
--- a/src/Uuid/Uuid.php
+++ b/src/Uuid/Uuid.php
@@ -76,20 +76,20 @@ final class Uuid
             return null;
         }
 
-        if (null === self::uuid_parse_as_array($uuid_ns)) {
+        if (!self::isValid($uuid_ns)) {
             return false;
         }
 
         $hash = md5(hex2bin(str_replace('-', '', $uuid_ns)).$name);
 
-        return sprintf('%08s-%04s-%04x-%04x-%012s',
+        return sprintf('%08s-%04s-3%03s-%04x-%012s',
             // 32 bits for "time_low"
             substr($hash, 0, 8),
             // 16 bits for "time_mid"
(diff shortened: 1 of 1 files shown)


src/Uuid/Uuid.php
```diff
@@ -146,7 +146,7 @@ public static function uuid_is_valid($uuid)
             return null;
         }
 
-        return null !== self::uuid_parse_as_array($uuid);
+        return (bool) self::isValid($uuid);
```
lyrixx: let's cast in bool in `isValid` instead. It feels more natural to me that `isValid` returns a bool

src/Uuid/Uuid.php
```diff
@@ -305,40 +303,31 @@ public static function uuid_unparse($uuidAsBinary)
...
         }
 
-        $uuid = sprintf('%s-%s-%s-%s-%s',
-            substr($data, 0, 8),
-            substr($data, 8, 4),
-            substr($data, 12, 4),
-            substr($data, 16, 4),
-            substr($data, 20, 12)
-        );
+        $uuidAsBinary = bin2hex($uuidAsBinary);
```
stof: Overriding the variable `$uuidAsBinary` to contain the uuid as hex looks wrong to me.

src/Uuid/Uuid.php
nicolas-grekas: updated

src/Uuid/Uuid.php
nicolas-grekas: updated

fabpot:
Thank you @nicolas-grekas.

PR #219, opened 2020-01-14
[PHP8.0] add ValueError and FILTER_VALIDATE_BOOL

IonBazan:
This change adds support for new PHP 8.0 features:

- [x] `ValueError` class
- [x] `FILTER_VALIDATE_BOOL` constant

Issue #218, opened 2020-01-13
Fix PHP nightly builds


nicolas-grekas:
Here is a build that fails on PHP8:
https://travis-ci.org/symfony/polyfill/jobs/636310496

Help wanted to fix them!

stof:
Most of the failures are related to PHP 8 converting warnings to TypeError for bad arguments.

We need to figure out what to do with them.

Note that if we want to polyfill the behavior exactly both for strict mode and loose mode, we might have to use typehints on the polyfill methods directly.

Ayesh:
In regards to the exceptions thrown in nightly builds for type errors:

[PHP 8 will throw `\TypeError`](https://php.watch/versions/8.0/internal-function-exceptions) and [`\ValueError`](https://php.watch/versions/8.0/ValueError) exceptions, and this can be a huuuge break :)

This is no where near complete, but I think one approach would be to throw a `\TypeError` or `\ValueError` exception from the polyfill code, and conditionally expect a warning (PHPUnit-assisted warning) or an exception of type `\TypeError` or `\ValueError`. 

Ideally, we can drop PHP 5.x version support for good and add scalar typing. We will still need to work on `\ValueError` exceptions and work out differences thrown from within the function and at type-checking stage. 

Please see the second commit above. This approach worked to bring number of failures from [88](https://travis-ci.com/github/Ayesh/polyfill/jobs/314929103#L2199) to [76](https://travis-ci.com/github/Ayesh/polyfill/jobs/315205839#L2137) (this includes all errors throughout the polyfill suite).

stof:
@Ayesh please open a pull request. It makes the review much easier.

stof:
Well, for ValueError, we won't be able to make our polyfill turn existing core functions to the new behavior, as the native implementation would be used, not our polyfill.
So any code needing to support both PHP 7.x (and maybe 5.x) and PHP 8 will need to handle both kind of error handling (returning `false` with a warning, and throwing a `ValueError`) anyway.

What the polyfill should do is making sure that when it runs on PHP 8 for extensions that are not installed, it throws a `ValueError` rather than returning a `false` with a warning.

Ayesh:
Thanks @stof - I did not feel like the commit above was complete, so I did not open a PR yet. 

> What the polyfill should do is making sure that when it runs on PHP 8 for extensions that are not installed, it throws a ValueError rather than returning a false with a warning.

[...]

PR #247, opened 2020-04-06
[WIP] Fixes for mbstring with PHP 8 throwing exceptions instead of raising warnings

Ayesh:
Related #218.

This is an attempt to workaround the PHP 8's behavior of throwing `\TypeError` and [`\ValueError`](https://php.watch/versions/8.0/ValueError) exceptions on internal functions[1]. This is a rather massive change, so we will be taking a look at in bite-sizes. 

Both `Mbstring` and its test classes get two new helper methods. They check the PHP version, and if the PHP version is 8.0 or later, it will throw an exception from the polyfill class, and the complementing helper method will expect the same type of error. Both helper methods accept the name of the new exception class as the last parameter. We currently use `\TypeError`, but we will need to manually review each test to make sure the type of errors they throw. In most cases, this will be a `\ValueError` to follow the semantics of the error message. 

This PR is to get some feedback on if others would agree if this approach indeed makes sense. #219 already adds the new `\ValueError` class.

I considered to make the conditional expect/trigger helper methods into a trait, but traits are only available in PHP 5.4, but the polyfill needs to support PHP 5.3. However, because we need to split the repo to separate packages, this trait wouldn't be as handy as I thought it would be, so the helper methods are baked into the test and polyfill themselves.

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -77,6 +77,25 @@ final class Mbstring
         array('μ', 's', 'ι',        'σ', 'β',        'θ',        'φ',        'π',        'κ',        'ρ',        'ε',        "\xE1\xB9\xA1", 'ι'),
     );
 
+    /**
+     * A helper method to either trigger an error, or throw an exception in PHP 8.
+     *
+     * @param string $error_msg  the error message to be used in the error or exception
+     * @param int    $error_type Type of the error. Not used when an exception is thrown.
+     * @param null   $exception  in case an exception must be thrown in PHP 8,
+     *                           the fully-qualified name of the exception
+     */
+    private static function triggerErrorOrException(
+      $error_msg,
+      $error_type = E_USER_NOTICE,
+      $exception = null
+    ) {
+        if (null !== $exception && PHP_VERSION_ID >= 80000) {
+            throw new $exception($error_msg);
+        }
+        trigger_error($error_msg, $error_type);
+    }
+
     public static function mb_convert_encoding($s, $toEncoding, $fromEncoding = null)
     {
         if (\is_array($fromEncoding) || false !== strpos($fromEncoding, ',')) {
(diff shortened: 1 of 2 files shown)


Ayesh:
Thanks a lot @IonBazan - I have applied CS fixes on hunks this PR changes.

src/Mbstring/Mbstring.php
```diff
@@ -77,6 +77,25 @@ final class Mbstring
...
+     * @param int    $error_type Type of the error. Not used when an exception is thrown.
+     * @param null   $exception  in case an exception must be thrown in PHP 8,
+     *                           the fully-qualified name of the exception
+     */
+    private static function triggerErrorOrException(
+      $error_msg,
+      $error_type = E_USER_NOTICE,
+      $exception = null
+    ) {
+        if (null !== $exception && PHP_VERSION_ID >= 80000) {
```
stof, line 93: I'd rather inline this than each place that creating this method. It will be more efficient, as OPCache can optimize the `PHP_VERSION_ID` check at compile time, and it will avoid passing class names as string.

And btw, this might also help in case the message needs to change in the PHP 8 exception.

nicolas-grekas:
What about creating polyfills v2 instead? We would maintain two branches: v1 for <8 and v2 for >=8.

stof:
> What about creating polyfills v2 instead? We would maintain two branches: v1 for <8 and v2 for >=8.

seems like a good idea. Then the v2 can use native typehints in functions and let PHP throw the TypeError btw.

derrabus:
> > What about creating polyfills v2 instead? We would maintain two branches: v1 for <8 and v2 for >=8.
> 
> seems like a good idea. Then the v2 can use native typehints in functions and let PHP throw the TypeError btw.

Creating a v2 with a higher php requirement would certainly ease future maintenance, but I wouldn‘t require php 8 just yet. But php 7.2 as a minimum would be fine imho.

stof:
@derrabus the goal is not to abandon v1. the goal is to make it easier to implement polyfills that respect the warning to exception migration which happens in PHP 8.

derrabus:
Yes, but if the consequence is that I need a different set of dependencies for php 7.4 and 8.0 in my projects, I don‘t think that we have a good solution.

derrabus:
Also, it‘s way too soon to create a release that requires php 8. The ProxyManager experiment last year (a ProxyManager release with php 7.4 as min was tagged while 7.4 was still in development) showed this.

jderusse:
> Yes, but if the consequence is that I need a different set of dependencies for php 7.4 and 8.0 in my projects, I don‘t think that we have a good solution.

not necessary. If `symfony/polyfill: 2` requires `php: >=8` and `symfony/polyfill: 1` conflicts with `php: >=8`, then you can requires `symfony/polyfill: 1 || 2` in your application and let composer choosing the right version of the polyfill.

I think following major versions of PHP make sens for polyfills.

derrabus:
> then you can requires `symfony/polyfill: 1 || 2` in your application and let composer choosing the right version of the polyfill.

That is my point exactly. Composer has to chose a different version depending on my php version. The consequence is that I cannot produce a build of my application that can be tested on php 7.4 and 8.0 alike. This is the kind of stuff that makes migrations painful. Please let‘s not di this.

nicolas-grekas:
I agree with @derrabus, relying on composer for this is risky practice, better avoid it.

nicolas-grekas:
Closing in favor of #316

PR #248, opened 2020-04-12
Fix the polyfill for spl_object_id()

TysonAndre:
spl_object_hash will be a string such as
`000000001b19a247000000004e677fa5`.
The first half is 8 to 16 zero-padded **hex digits** of
data (PHP_INT_SIZE=4 **bytes** on 32-bit systems, 8 on 64-bit systems)

This meant that this polyfill for spl_object_id previously couldn't accurately
represent values that were greater than 65536 (`16**4`) on 32-bit systems,
~~or 4294967296 (`16**8`) on 64-bit systems.~~

This was noticed because 32-bit tests failed for Phan in php 7.1.
(https://github.com/phan/phan/blob/master/src/spl_object_id.php had no
issues before the dependency on symfony/polyfill-72 was added for a
different function)

- Aside: The object handle of `struct _zend_object` was a uint32_t in php 7.1,
  so it's always 8 hex digits for official php builds.

### DIFF
--- a/tests/Php72/Php72Test.php
+++ b/tests/Php72/Php72Test.php
@@ -67,6 +67,20 @@ class Php72Test extends TestCase
     }
 
     /**
+     * @covers \Symfony\Polyfill\Php72\Php72::spl_object_id
+     */
+    public function testSplObjectIdUniqueValues()
+    {
+        // Should be able to represent more than 2**16 ids on 32-bit systems.
+        $result = array();
+        for ($i = 0; $i < 70000; $i++) {
+            $obj = new \stdClass();
+            $result[spl_object_id($obj)] = $obj;
+        }
+        $this->assertSame(70000, count($result));
+    }
+
+    /**
      * @covers \Symfony\Polyfill\Php72\Php72::sapi_windows_vt100_support
      */
     public function testSapiWindowsVt100Support()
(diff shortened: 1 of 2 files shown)


tests/Php72/Php72Test.php
```diff
@@ -66,6 +66,20 @@ public function testSplObjectId()
         $this->assertNull(@spl_object_id(123));
     }
 
+    /**
+     * @covers \Symfony\Polyfill\Php72\Php72::spl_object_id
+     */
+    public function testSplObjectIdUniqueValues()
+    {
+        // Should be able to represent more than 2**16 ids on 32-bit systems.
+        $result = [];
```
staabm, line 75: For BC
```suggestion
        $result = array();
```

TysonAndre: Done

TysonAndre:
This is ready for another look.

The build failure seems to be unrelated

https://travis-ci.org/github/symfony/polyfill/jobs/674458882

```
1.57s$ composer --prefer-source install
                                            
  [Composer\Downloader\TransportException]  
  Peer fingerprint did not match            
```

nicolas-grekas:
Thank you @TysonAndre

PR #250, opened 2020-04-21
Add polyfill for PHP 8's str_starts_with() and str_ends_with() functions

Ayesh:
Adds polyfills for str_starts_with() and str_ends_with() functions added in PHP 8.0.
RFC: https://wiki.php.net/rfc/add_str_starts_with_and_ends_with_functions
Tests from PHP source:
 - ext/standard/tests/strings/str_starts_with.phpt
 - ext/standard/tests/strings/str_ends_with.phpt

Sample implementation: https://php.watch/versions/8.0/str_starts_with-str_ends_with

*Marking this PR draft because the RFC voting has yet to be closed (on May 4)*

### DIFF
--- a/src/Php80/Php80.php
+++ b/src/Php80/Php80.php
@@ -83,4 +83,14 @@ final class Php80
     {
         return '' === $needle || false !== strpos($haystack, $needle);
     }
+
+    public static function str_starts_with(string $haystack, string $needle): bool
+    {
+        return 0 === \strncmp($haystack, $needle, \strlen($needle));
+    }
+
+    public static function str_ends_with(string $haystack, string $needle): bool
+    {
+        return '' === $needle || $needle === \substr($haystack, -\strlen($needle));
+    }
 }
--- a/src/Php80/bootstrap.php
+++ b/src/Php80/bootstrap.php
@@ -24,6 +24,14 @@ if (PHP_VERSION_ID < 80000) {
         function str_contains(string $haystack, string $needle): bool { return p\Php80::str_contains($haystack, $needle); }
     }
 
+    if (!function_exists('str_starts_with')) {
+        function str_starts_with(string $haystack, string $needle): bool { return p\Php80::str_starts_with($haystack, $needle); }
+    }
+
+    if (!function_exists('str_ends_with')) {
+        function str_ends_with(string $haystack, string $needle): bool { return p\Php80::str_ends_with($haystack, $needle); }
+    }
+
     if (!defined('FILTER_VALIDATE_BOOL') && defined('FILTER_VALIDATE_BOOLEAN')) {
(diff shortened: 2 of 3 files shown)


Ayesh:
Thank you. I have added some elaborate tests that use my native language (Sinhalese) in it. Internally, these functions use memory compare (using C `memcmp`), so they always work the same way `subtr`, `strlen`, and `strncmp` functions which we use in the polyfill.

Ayesh:
The RFC vote is currently at 38-4 in favor. The highest number of votes cast on an RFC was 71, and even if everyone remaining votes no, 38/71 = 53%, so this RFC can be considered a pass. I suppose it's OK to submit this for review.

stof:
@Ayesh 53% is not a passing vote. The threshold is 2/3.

Ayesh:
@stof - shouldn't it be 50%+1 because the RFC is not a language change? Syntax changes sure would require 2/3, but this RFC has no BC impact.

jdreesen:
I think it's 2/3 every time, since https://wiki.php.net/rfc/abolish-narrow-margins

nicolas-grekas:
Thank you @Ayesh.

jrfnl:
Just a heads-up: looks like the `readme` of the PHP 8 polyfill still needs updating for this change.

Ayesh:
Thanks for the good catch @jrfnl - I have pushed a commit with the changes 👍.

jrfnl:
@nicolas-grekas Re: Travis not reporting back - you can find the most likely solution here: https://twitter.com/jrf_nl/status/1255854838597984262

src/Php80/Php80.php
```diff
@@ -83,4 +83,14 @@ public static function str_contains(string $haystack, string $needle): bool
...
     }
+
+    public static function str_starts_with(string $haystack, string $needle): bool
+    {
+        return 0 === \strncmp($haystack, $needle, \strlen($needle));
+    }
+
+    public static function str_ends_with(string $haystack, string $needle): bool
+    {
+        return '' === $needle || $needle === \substr($haystack, -\strlen($needle));
```
guilliamxavier, line 94: Using `substr()` is suboptimal (as [pointed in the RFC](https://wiki.php.net/rfc/add_str_starts_with_and_ends_with_functions#str_ends_with)). The [actual implementation](https://github.com/php/php-src/pull/5300/files?file-filters%5B%5D=.c) rather does something akin to `substr_compare()`, which could give:

```php
        return '' === $needle || ('' !== $haystack && 0 === \substr_compare($haystack, $needle, -\strlen($needle)));
```

(see https://3v4l.org/NUS87 and the "based on" links for why the two guards)

nicolas-grekas: Please provide a PR and some benchmark if possible.

guilliamxavier: See #255

PR #251, opened 2020-04-29
Add checks for each `mb_*` function separately

capuderg:
As discussed in this PR: https://github.com/symfony/polyfill-mbstring/pull/6, this would be a potential fix for any corner cases.

Currently, we have issue with `mb_strtolower` not being polyfilled, because the current check only checks if `mb_strlen` exists.

In WordPress the `mb_strlen` function is polyfilled before this package can be required, but the `mb_strtolower` is not (don't know why WP is polyfilling just a few).

With this change, all mb_* functions can be polyfilled separately.

### DIFF
--- a/src/Mbstring/bootstrap.php
+++ b/src/Mbstring/bootstrap.php
@@ -13,49 +13,161 @@ use Symfony\Polyfill\Mbstring as p;
 
 if (!defined('MB_CASE_UPPER')) {
     define('MB_CASE_UPPER', 0);
+}
+
+if (!defined('MB_CASE_LOWER')) {
     define('MB_CASE_LOWER', 1);
+}
+
+if (!defined('MB_CASE_TITLE')) {
     define('MB_CASE_TITLE', 2);
 }
 
-if (!function_exists('mb_strlen')) {
+if (!function_exists('mb_convert_encoding')) {
     function mb_convert_encoding($s, $to, $from = null) { return p\Mbstring::mb_convert_encoding($s, $to, $from); }
+}
+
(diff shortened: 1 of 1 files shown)


stof:
If we decide that we want to go that road, we should do it for all polyfills.

nicolas-grekas:
> If we decide that we want to go that road, we should do it for all polyfills.

I agree - and then, what is now a few tenths of extra check might become hundreds. We should definitely run some benchmarks before doing this.

JanJakes:
@nicolas-grekas In many cases (like PHP version polyfill) you won't add much, almost nothing in fact, since most of the IFs are already there (i.e. https://github.com/symfony/polyfill/blob/master/src/Php54/bootstrap.php).

And then some polyfills may not need this at all (classes? not sure which ones).

And I guess even if you add hundreds it will still be a fraction of a millisecond but yes, a benchmark would be good to prove such a guess.

stof:
@nicolas-grekas isn't `function_exists` optimized by OPCache, so that it is optimized-away entirely when you have the extension loaded anyway ?

JanJakes:
@stof Seems you're right: https://github.com/php/php-src/blob/e141592da6cdb15709e4aeaab5421c9a46564849/ext/opcache/Optimizer/pass1_5.c#L349

See also: https://stackoverflow.com/questions/21619939/does-function-exists-cache-its-queries/21620062#21620062

JanJakes:
I prepared a PR handling this across all polyfills - https://github.com/symfony/polyfill/pull/252. Let's see where it leads us.

PR #252, opened 2020-04-30
Add separate checks for all polyfilled functions and constants

JanJakes:
As originally discussed in https://github.com/symfony/polyfill-mbstring/pull/6 and later in https://github.com/symfony/polyfill/pull/251 in some scenarios it may be better to check for each constant/function before defining it. Such scenarios are typically those where some project or a library defines its polyfills for some parts of extension functionality but not all of it.

Probably the most popular example is WordPress, defining some polyfills in `wp-includes/compat.php` but not including complete functionalities (i.e. it defines only a few `mb_` functions). Plugin developers then may need other `mb_` functions and need to load those that WordPress didn't.

There may be some performance concerns but it seems that checks such as `function_exists` are [very fast](https://github.com/symfony/polyfill-mbstring/pull/6#issuecomment-621117197) - in fact, [PHP tries to evaluate both `function_exists` and `defined` at compile time](https://github.com/php/php-src/blob/e141592da6cdb15709e4aeaab5421c9a46564849/ext/opcache/Optimizer/pass1_5.c#L349) (see also [StackOverflow](https://stackoverflow.com/questions/21619939/does-function-exists-cache-its-queries/21620062#21620062)). This means on installs that do have the polyfilled extensions installed, there should be zero overhead (the conditions will evaluate to `true` at compile time).

Note that many of the currently existing polyfills already perform checks for each function separately (most of the PHP-version polyfills).

Closes https://github.com/symfony/polyfill/pull/251.

### DIFF
--- a/src/Apcu/bootstrap.php
+++ b/src/Apcu/bootstrap.php
@@ -15,25 +15,56 @@ if (!extension_loaded('apc') && !extension_loaded('apcu')) {
     return;
 }
 
-if (!function_exists('apcu_add')) {
-    if (extension_loaded('Zend Data Cache')) {
+if (extension_loaded('Zend Data Cache')) {
+    if (!function_exists('apcu_add')) {
         function apcu_add($key, $var = null, $ttl = 0) { return p\Apcu::apcu_add($key, $var, $ttl); }
+    }
+    if (!function_exists('apcu_delete')) {
         function apcu_delete($key) { return p\Apcu::apcu_delete($key); }
+    }
+    if (!function_exists('apcu_exists')) {
         function apcu_exists($keys) { return p\Apcu::apcu_exists($keys); }
+    }
+    if (!function_exists('apcu_fetch')) {
         function apcu_fetch($key, &$success = null) { return p\Apcu::apcu_fetch($key, $success); }
+    }
+    if (!function_exists('apcu_store')) {
         function apcu_store($key, $var = null, $ttl = 0) { return p\Apcu::apcu_store($key, $var, $ttl); }
-    } else {
+    }
+} else {
+    if (!function_exists('apcu_add')) {
         function apcu_add($key, $var = null, $ttl = 0) { return apc_add($key, $var, $ttl); }
+    }
+    if (!function_exists('apcu_delete')) {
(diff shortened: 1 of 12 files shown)


nicolas-grekas:
To validate the change, can you please run a simple benchmark with nothing else but a require for vendor/autoload.php, in three situations:
- all extensions loaded in php, polyfills installed
- no extensions loaded in php, polyfills installed
- again but with NO polyfills installed
?

If you think about other relevant benchmarks, please proceed :) Thanks.

capuderg:
I've run some basic time benchmarks. I've only tested the [mbstrings polyfill library](https://github.com/symfony/polyfill-mbstring) (not this whole library)

A new PHP 7.4.5 CLI docker container was used as an environment.

I've run the basic `require_once __DIR__ . '/vendor/autoload.php';` 1000 times and here are the averages (in seconds) for each situation: 

- "empty" composer (no packages required) with `mbstring` extension disabled: 0.0179128654
- "empty" composer (no packages required) with `mbstring` extension enabled: 0.01777232742

- "symfony/polyfill-mbstring = 1.15" required with `mbstring` extension disabled: 0.02518279123
- "symfony/polyfill-mbstring = 1.15" required with `mbstring` extension enabled: 0.02492605114

- "symfony/polyfill-mbstring with updated bootstrap.php" required with `mbstring` extension disabled: 0.02665189552
- "symfony/polyfill-mbstring with updated bootstrap.php" required with `mbstring` extension enabled: 0.02570387244

I hope this helps :)

JanJakes:
@capuderg Great, thanks! I didn't have time to get to this yet, only started to prepare a PHP Docker container without mbstring. This looks good, not sure if we need any more tests.

@nicolas-grekas Do you have any clues why the builds are failing? The `Parse error: syntax error, unexpected T_OBJECT_OPERATOR in C:\projects\polyfill\vendor\bin\.phpunit\phpunit-4.8\phpunit on line 13` doesn't say much.

nicolas-grekas:
Thank you @JanJakes.

nicolas-grekas:
And thank you @capuderg also.

PR #257, opened 2020-05-06
Fail on leading or trailing hyphen-minus (RFC 3492)

### LINKED ISSUE #256: Discrepancy between the logic of idn_to_ascii and the polyfilled implementation
The native (PHP 7.4.5, ICU 66.1) `idn_to_ascii` implementation is unable to process a string that contains a leading or a trailing dash while the polyfilled implementation is able:
```php
$part = 'xyz-';
var_dump(
    idn_to_ascii($part, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46)
);
// bool(false)

var_dump(
    Symfony\Polyfill\Intl\Idn\Idn::idn_to_ascii($part, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46)
);
// string(4) "xyz-"
```
The native implementation seems to be the correct one according to [RFC 3492](https://tools.ietf.org/rfc/rfc3492):

> […] IDNA using Punycode conforms to the RFC 952 rule that host name labels neither begin nor end with a hyphen-minus […].

morozov:
Fixes #256.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
   * added `str_starts_with()` and `str_ends_with()` to the PHP 8 polyfill
   * fixed `spl_object_id()` on 32-bit systems
+  * fixed `idn_to_ascii()` not failing on leading or trailing hyphen-minus
 
 # 1.15.0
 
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Thank you @morozov.

PR #260, opened 2020-05-11
[PHP 8.0] add `get_resource_id` function

### LINKED ISSUE #259: Polyfill for get_resource_id()
As described in https://github.com/php/php-src/pull/5427

Anyone up to give it a try?

Be careful: we must also polyfill the behavior when a non-resource is passed (I don't know the behavior, it might be a warning or an exception - to be investigated).

IonBazan:
Resolves #259 

PHP 8.0 output for reference:
```
php > $res = fopen('test', 'w');
php > var_dump((int) $res, get_resource_id($res));
int(2)
int(2)
php > fclose($res);
php > var_dump((int) $res, get_resource_id($res));
int(2)
int(2)
php > $res = 10;
php > var_dump((int) $res, get_resource_id($res));
PHP Warning:  Uncaught TypeError: get_resource_id(): Argument #1 ($res) must be of type resource, int given in php shell code:1
Stack trace:
#0 php shell code(1): get_resource_id(10)
#1 {main}

Warning: Uncaught TypeError: get_resource_id(): Argument #1 ($res) must be of type resource, int given in php shell code:1
Stack trace:
#0 php shell code(1): get_resource_id(10)
#1 {main}
```

### DIFF
--- a/src/Php80/Php80.php
+++ b/src/Php80/Php80.php
@@ -57,6 +57,15 @@ final class Php80
         return (get_parent_class($class) ?: key(class_implements($class)) ?: 'class').'@anonymous';
     }
 
+    public static function get_resource_id($res): int
+    {
+        if (!\is_resource($res) && null === @get_resource_type($res)) {
+            throw new \TypeError(sprintf('Argument 1 passed to get_resource_id() must be of the type resource, %s given', get_debug_type($res)));
+        }
+
+        return (int) $res;
+    }
+
     public static function preg_last_error_msg(): string
     {
         switch (preg_last_error()) {
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Polyfills are provided for:
 - the `str_starts_with` and `str_ends_with` functions introduced in PHP 8.0;
 - the `ValueError` class introduced in PHP 8.0;
 - the `FILTER_VALIDATE_BOOL` constant introduced in PHP 8.0;
+- the `get_resource_id` function introduced in PHP 8.0;
 
 It is strongly recommended to upgrade your PHP version and/or install the missing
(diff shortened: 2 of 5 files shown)


src/Php80/Php80.php
```diff
@@ -57,6 +57,15 @@ public static function get_debug_type($value): string
         return (get_parent_class($class) ?: key(class_implements($class)) ?: 'class').'@anonymous';
     }
 
+    public static function get_resource_id($res): int
+    {
+        if (@\get_resource_type($res) === null) {
```
nicolas-grekas, line 62: let's optimize for the happy path:
`if (!\is_resource($res) && null === @get_resource_type($res)) {`

src/Php80/Php80.php
```diff
@@ -57,6 +57,15 @@ public static function get_debug_type($value): string
...
+            throw new \TypeError(sprintf('Argument #1 ($res) must be of type resource, %s given', \gettype($res)));
```
nicolas-grekas, line 63: the message should match the one from PHP8
looking at https://wiki.php.net/rfc/consistent_type_errors, this should be more like
`Argument 1 passed to get_resource_id() must be of the type int, string given`

stof: well, `resource` rather than `int` of course :smile:

src/Php80/Php80.php
stof: do we even need the `get_resource_type` check ?

nicolas-grekas: we need it for closed resources yes, that's an odd edge case.

nicolas-grekas:
Thank you @IonBazan.

PR #263, opened 2020-05-12
Fix declaring extra constants when `intl` is loaded

### LINKED ISSUE #262: symfony/polyfill-intl-idn 1.16 broken Guzzle for CentOS 6 (old ICU) servers?
FWIW, today the update to `symfony/polyfill-intl-idn` v1.16 (from 1.15) just completely broke all my Guzzle usages. This is on a server running CentOS 6 (and therefore old ICU - 4.2.1?) and PHP 7.3

nicolas-grekas:
Fix #262
Best to [ignore whitespace](https://github.com/symfony/polyfill/pull/263/files?w=1) when reviewing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 1.17.0
 
   * added `get_resource_id()` to the PHP 8 polyfill
+  * fix declaring extra constants when `intl` is loaded
 
 # 1.16.0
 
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # 1.17.0
 
   * added `get_resource_id()` to the PHP 8 polyfill
+  * fix declaring extra constants when `intl` is loaded
```
stof, line 4: This also applies to `mbstring`

PR #266, opened 2020-06-04
Bump unicode data

nicolas-grekas:
This PR updates the unicode maps and imports the [`Compiler`](https://github.com/tchwork/utf8/blob/master/src/Patchwork/Utf8/Compiler.php) class from patchwork/utf8 to help with future updates.


fabpot:
Thank you @nicolas-grekas.

PR #269, opened 2020-06-17
Align `APCuIterator` case with its implementation

carusogabriel:
This is the correct case defined in the source code: https://github.com/krakjoe/apcu/blob/1f98e34d93/apc_iterator.h#L29

This is an attempt to fix errors that a project started to report on PHPStan `v0.12.26`:

```
 ------ ------------------------------------------------------------------
  Line   src/State/StateStorage.php
 ------ ------------------------------------------------------------------
  6      Class APCUIterator referenced with incorrect case: APCuIterator.
  109    Class APCUIterator referenced with incorrect case: APCuIterator.
 ------ ------------------------------------------------------------------
```

### DIFF
--- a/src/Apcu/README.md
+++ b/src/Apcu/README.md
@@ -1,7 +1,7 @@
 Symfony Polyfill / APCu
 ========================
 
-This component provides `apcu_*` functions and the `APCUIterator` class to users of the legacy APC extension.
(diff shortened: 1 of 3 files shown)


carusogabriel:
cc @nicolas-grekas @fabpot

nicolas-grekas:
Good catch, thanks @carusogabriel.

PR #270, opened 2020-07-03
Add `UnhandledMatchError` exception class

Ayesh:
Along with the `match` expression syntax in PHP 8, it [registers](https://github.com/php/php-src/blob/c2d3cd2595481af4b2edd7cf26fb148de2d475aa/Zend/zend_exceptions.c#L819-L820) a new exception class `UnhandledMatchError` that extends `\Error`.

While the `match` expression itself cannot be polyfilled, we can backport this new class for the feature parity, so exceptions can be unserialized, or exceptions can be thrown in other code if they choose to throw `\UnhandledMatchError` exceptions.

 - [RFC](https://wiki.php.net/rfc/match_expression_v2)
 - [Implementation](https://github.com/php/php-src/blob/c2d3cd2595481af4b2edd7cf26fb148de2d475aa/Zend/zend_exceptions.c#L819-L820)
 - [`match` expressions # `UnhandledMatchError`](https://php.watch/versions/8.0/match-expression#UnhandledMatchError)
 - PR for `ValueError` (#219)

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Polyfills are provided for:
 - the `str_contains` function introduced in PHP 8.0;
 - the `str_starts_with` and `str_ends_with` functions introduced in PHP 8.0;
 - the `ValueError` class introduced in PHP 8.0;
+- the `UnhandledMatchError` class introduced in PHP 8.0;
 - the `FILTER_VALIDATE_BOOL` constant introduced in PHP 8.0;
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Thank you @Ayesh.

Ayesh:
Thank you @nicolas-grekas.

PR #272, opened 2020-07-08
Use https on appveyor

TRowbotham:
Tell OpenSSL to use the cert file from the [curl](https://curl.haxx.se/docs/caextract.html) website, which should eliminate the need to store a composer config that has SSL/TLS protection disabled. It should also remove all the warnings from SSL/TLS protection being disabled and download failure messages.


TRowbotham:
Well, those results are unfortunate. Closing.

PR #274, opened 2020-07-10
Fix registering polyfill-php80 in the meta composer.json

### LINKED ISSUE #273: Installing symfony/polyfill as well as symfony/polyfill-php80 doesn't work
The functions never get defined. symfony/polyfill shouldn't have claimed to replace the polyfill if it didn't?

nicolas-grekas:
Fix #273

### DIFF
--- a/src/Php80/bootstrap.php
+++ b/src/Php80/bootstrap.php
@@ -9,6 +9,8 @@
  * file that was distributed with this source code.
  */
 
+// this file should be kept in sync with src/bootstrap.php
+
 use Symfony\Polyfill\Php80 as p;
 
 if (PHP_VERSION_ID >= 80000) {
--- a/src/bootstrap.php
+++ b/src/bootstrap.php
@@ -9,6 +9,42 @@
  * file that was distributed with this source code.
  */
 
-if (PHP_VERSION_ID < 80000 && PHP_VERSION_ID >= 70000) {
+// this file should be kept in sync with src/Php80/bootstrap.php
+
+use Symfony\Polyfill\Php80 as p;
+
+if (PHP_VERSION_ID >= 80000) {
+    return;
+}
+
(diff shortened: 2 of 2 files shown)


src/Php80/bootstrap.php
```diff
@@ -20,23 +20,23 @@
 }
 
 if (!function_exists('fdiv')) {
-    function fdiv(float $dividend, float $divisor): float { return p\Php80::fdiv($dividend, $divisor); }
```
GrahamCampbell, line 23: Could we conditionally define these functions so that on PHP 7 they have the types?

nicolas-grekas: yes, updated.

nicolas-grekas:
Wait, this is wrong. The php8 polyfill is not compatible with PHP5.

stof:
Then we should not gave the `replace` rule either in the root composer.json, as it does not provide the feature.

nicolas-grekas:
yes.it.does.

stof:
@nicolas-grekas I don't understand your answer. Without this PR, installing `symfony/polyfill` does not provide the features as they are not autoloaded.

nicolas-grekas:
The PHP 8 polyfill supports only PHP >= 7.

For this reason, on PHP 5, the root polyfill package does not define the functions of the php8 one.
On PHP 7, these functions are declared as expected.

See #234 for how this works.

GrahamCampbell:
But there is still a bug. The polyfill package incorrectly claims to replace it, so if I'm on PHP 7 and the main polyfill package gets loaded and a library requires the php8 polyfill, then the library gets broken...

GrahamCampbell:
Oh, wait. I see there's a dedicated bootstrap file for this case. Sorry for the confusion!

PR #277, opened 2020-07-21
Only define classes if they don't exist

someonewithpc:
I noticed specifically with `Normalizer` that I get an error saying:

> Compile Error: Cannot declare class Normalizer, because the name is already in use

This occurred with the `intl` extension installed and trying to use the class without importing it with `use`.

### DIFF
--- a/src/Intl/MessageFormatter/Resources/stubs/IntlException.php
+++ b/src/Intl/MessageFormatter/Resources/stubs/IntlException.php
@@ -1,5 +1,7 @@
 <?php
 
-class IntlException extends Exception
-{
+if (!class_exists('IntlException')) {
+    class IntlException extends Exception
+    {
+    }
 }
--- a/src/Intl/MessageFormatter/Resources/stubs/MessageFormatter.php
+++ b/src/Intl/MessageFormatter/Resources/stubs/MessageFormatter.php
@@ -1,5 +1,7 @@
 <?php
 
-class MessageFormatter extends Symfony\Polyfill\Intl\MessageFormatter\MessageFormatter
-{
+if (!class_exists('MessageFormatter')) {
+    class MessageFormatter extends Symfony\Polyfill\Intl\MessageFormatter\MessageFormatter
+    {
+    }
 }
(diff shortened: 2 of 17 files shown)


stof:
The autoloader is responsible for requiring these files, and it will never be triggered at all if the class already exists.

Please describe how you get this error. It seems to indicate that you load the files in an unsupported way.

someonewithpc:
In a symfony project, attempting to use `Normalizer` without a `use` statement, like I said.

stof:
Please provide a full reproducer. The case you describe in one sentence is precisely the case that has been tested by the core team without any issue, meaning that you probably miss something.

xabbuh:
How exactly do you use the class without a `use` statement?

someonewithpc:
I created a barebones project at:
https://github.com/someonewithpc/symfony-polyfill-test/blob/master/src/Kernel.php#L29

If you clone it and run `php -S localhost:8080 public/index.php` you should see an error like
```
Attempted to load class "Normalizer" from namespace "App".
Did you forget a "use" statement for "Symfony\Polyfill\Intl\Normalizer\Normalizer"?
```
----
 (as an aside `symfony server:start` and `bin/console server:start` don't work, like suggested in the docs).

someonewithpc:
> How exactly do you use the class without a `use` statement?

With it being in the global namespace, I'm not supposed to need to `use` it, so I just write `$foo = Normalizer::normalize('foo');`

stof:
@someonewithpc this error message tells me that your code is **not** in the global namespace but in the `App` namespace. To access a class of the global namespace, you must either add a `use` statement for it (to import it in your current namespace temporarily) or use a fully-qualified name (`\Normalizer`).
The rule to use classes of the global namespace is not different from the rule to use namespaced classes. Functions and constants have a fallback to the global namespace in their name resolution, but classes don't have it.

xabbuh:
As @stof said you need to change your code to `$foo = \Normalizer::normalize('foo');` if you don't want to use a `use` statement. I am closing here as that's not related to the polyfill.

PR #281, opened 2020-08-05
Avoid using any IDNA constants in the internal polyfill implementation

### LINKED ISSUE #278: [BUG] polyfill-intl-idn breaks on large hoster installation
We use the intl-idn package in TYPO3 CMS and it helps a lot! Thank you for providing such a package.

However there is one large TYPO3 hoster (having roughly 200.000 TYPO3 installations), where they have a messed up Server setup regarding to ICU version on their system. In short: They provide the php-intl extension, but do not define all constants as needed for PHP 7.2+. We had solved this issue in the past by a hard workaround, but with the update to polyfill-intl-idn 1.17.1 and 1.18.0 then (respective change) https://github.com/symfony/polyfill-intl-idn/commit/3bff59ea7047e925be6b7f2059d60af31bb46d6a (or https://github.com/symfony/polyfill/commit/f968149bedaaa9013d39fecd9cab1139aba1334a#diff-eaa55ac0b04a1928e5455b4f48efe417R14-R17) our setups don't work on the hoster anymore, as they _clearly_ have a misconfiguration.

Still, the previous solution when not checking on the `extension_loaded()` was more robust for these kind of weird setups.

In any case, I'm (personally) fine if you decide to keep the `extension_loaded()` check in your library, but wanted to let you know that this actually is an issue for a lot of people. It would be great to know if you're able to fix it or planning to fix/revert the three lines or if we should find a workaround for ourselves in our project.

TRowbotham:
Fixes #278.

Sometimes we can get into an awkward configuration where `ext-intl` is enabled, but is compiled with a version of ICU < 4.6, which means that a number of constants aren't defined and the polyfill doesn't get loaded and therefore doesn't define the constants for us. This causes breakage for users using the internal API. Fix this by avoiding the use of any `IDNA_*` constants.

@bmack let me know if this fixes the issue for you.

### DIFF
--- a/src/Intl/Idn/Idn.php
+++ b/src/Intl/Idn/Idn.php
@@ -42,6 +42,14 @@ final class Idn
     const INTL_IDNA_VARIANT_2003 = 0;
     const INTL_IDNA_VARIANT_UTS46 = 1;
 
+    const IDNA_DEFAULT = 0;
+    const IDNA_ALLOW_UNASSIGNED = 1;
+    const IDNA_USE_STD3_RULES = 2;
+    const IDNA_CHECK_BIDI = 4;
+    const IDNA_CHECK_CONTEXTJ = 8;
+    const IDNA_NONTRANSITIONAL_TO_ASCII = 16;
+    const IDNA_NONTRANSITIONAL_TO_UNICODE = 32;
+
     const MAX_DOMAIN_SIZE = 253;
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @TRowbotham.

bmack:
@stof @TRowbotham works perfectly for my scenario. Tested on the "broken system" again, and it works. Thanks a ton for your support.

PR #283, opened 2020-08-05
Handle null being passed to PHP80 str_* methods

### LINKED ISSUE #282: str_contains behaviour differs from php8 behaviour
It looks like the PHP8 behaviour for `str_contains` actually silently accepts a null value (casting it to the empty string), whereas the polyfill doesn't. [See this comparison on 3v4l](https://3v4l.org/LPetC), where you can see that the polyfill throws a TypeError but PHP8 doesn't. The same is probably true for the other str methods, but I haven't confirmed it yet.

Happy to send a PR (possibly just easiest to remove the types from the params?)

PatrickRose:
The previous version of the polyfills didn't accept null as a value as
either the haystack or the needle because of the string typehint. In
PHP8, these are silently cast to an empty string.

This matches that behaviour, by making the typehints nullable.

Fixes #282

### DIFF
--- a/src/Php80/Php80.php
+++ b/src/Php80/Php80.php
@@ -88,17 +88,22 @@ final class Php80
         }
     }
 
-    public static function str_contains(string $haystack, string $needle): bool
+    public static function str_contains(string $haystack = null, string $needle = null): bool
     {
+        // Cast to string to avoid deprecation notices
+        if (null === $needle) {
+            $needle = '';
+        }
+
         return '' === $needle || false !== strpos($haystack, $needle);
     }
 
-    public static function str_starts_with(string $haystack, string $needle): bool
+    public static function str_starts_with(string $haystack = null, string $needle = null): bool
     {
         return 0 === \strncmp($haystack, $needle, \strlen($needle));
     }
 
-    public static function str_ends_with(string $haystack, string $needle): bool
+    public static function str_ends_with(string $haystack = null, string $needle = null): bool
     {
         return '' === $needle || ('' !== $haystack && 0 === \substr_compare($haystack, $needle, -\strlen($needle)));
     }
--- a/src/Php80/Php80.php
+++ b/src/Php80/Php80.php
@@ -91,9 +91,7 @@ final class Php80
     public static function str_contains(string $haystack = null, string $needle = null): bool
     {
         // Cast to string to avoid deprecation notices
-        if (null === $needle) {
-            $needle = '';
-        }
+        $needle = (string)$needle;
 
         return '' === $needle || false !== strpos($haystack, $needle);
     }
--- a/src/Php80/Php80.php
+++ b/src/Php80/Php80.php
@@ -103,6 +103,10 @@ final class Php80
 
     public static function str_ends_with(string $haystack = null, string $needle = null): bool
     {
+        // Cast both to string
+        $needle = (string)$needle;
+        $haystack = (string)$haystack;
+
         return '' === $needle || ('' !== $haystack && 0 === \substr_compare($haystack, $needle, -\strlen($needle)));
     }
 }
--- a/src/Php80/bootstrap.php
+++ b/src/Php80/bootstrap.php
@@ -26,13 +26,13 @@ if (!function_exists('preg_last_error_msg')) {
     function preg_last_error_msg(): string { return p\Php80::preg_last_error_msg(); }
 }
 if (!function_exists('str_contains')) {
-    function str_contains(string $haystack, string $needle): bool { return p\Php80::str_contains($haystack, $needle); }
+    function str_contains(string $haystack = null, string $needle = null): bool { return p\Php80::str_contains($haystack, $needle); }
 }
 if (!function_exists('str_starts_with')) {
-    function str_starts_with(string $haystack, string $needle): bool { return p\Php80::str_starts_with($haystack, $needle); }
+    function str_starts_with(string $haystack = null, string $needle = null): bool { return p\Php80::str_starts_with($haystack, $needle); }
 }
 if (!function_exists('str_ends_with')) {
-    function str_ends_with(string $haystack, string $needle): bool { return p\Php80::str_ends_with($haystack, $needle); }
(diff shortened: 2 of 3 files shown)


src/Php80/Php80.php
```diff
@@ -88,17 +88,23 @@ public static function preg_last_error_msg(): string
         }
     }
 
-    public static function str_contains(string $haystack, string $needle): bool
+    public static function str_contains(?string $haystack, ?string $needle): bool
     {
+        // Cast to string to avoid deprecation notices
+        if (null === $needle)
+        {
```
PatrickRose, line 95: This needs fixing - I'll sort after the tests have run in case there's another error I don't know about.

src/Php80/bootstrap.php
```diff
@@ -26,13 +26,13 @@ function fdiv(float $dividend, float $divisor): float { return p\Php80::fdiv($di
     function preg_last_error_msg(): string { return p\Php80::preg_last_error_msg(); }
 }
 if (!function_exists('str_contains')) {
-    function str_contains(string $haystack, string $needle): bool { return p\Php80::str_contains($haystack, $needle); }
+    function str_contains(string $haystack = null, string $needle = null): bool { return p\Php80::str_contains($haystack, $needle); }
```
IonBazan, line 29: How about we omit the type hint completely here. This way the parameter will be forced mandatory and the type check will occur in Php80 class only.

src/Php80/Php80.php
```diff
@@ -88,17 +88,22 @@ public static function preg_last_error_msg(): string
         }
     }
 
-    public static function str_contains(string $haystack, string $needle): bool
+    public static function str_contains(string $haystack = null, string $needle = null): bool
     {
+        // Cast to string to avoid deprecation notices
+        if (null === $needle) {
```
IonBazan, line 94: I think it can be simply casted to string.

tests/Php80/Php80Test.php
```diff
@@ -177,6 +186,11 @@ public function testStrEndsWith()
         $testEmoji = '🙌🎉✨🚀'; // 0xf0 0x9f 0x99 0x8c 0xf0 0x9f 0x8e 0x89 0xe2 0x9c 0xa8 0xf0 0x9f 0x9a 0x80
         $this->assertTrue(str_ends_with($testEmoji, "🚀")); // 0xf0 0x9f 0x9a 0x80
         $this->assertFalse(str_ends_with($testEmoji, "✨")); // 0xe2 0x9c 0xa8
+
+        $this->assertTrue(str_starts_with(null, ''));
+        $this->assertTrue(str_starts_with('', null));
+        $this->assertFalse(str_starts_with(null, 'test'));
+        $this->assertTrue(str_starts_with('test', null));
```
IonBazan, line 193: Please add some tests for type errors, when passing a bool or other types there.

src/Php80/bootstrap.php
PatrickRose: Sounds good to me!

stof: that's not good. Having no typehint makes the behavior different for code running with strict_types and calling the polyfill (the typehint in the PHP80 class won't reproduce this, as they won't be called with strict types from this file)

PatrickRose: @stof raises a good point - [I've just checked what the behaviour of `str_contains` with `strict_types` is on PHP80](https://3v4l.org/B5eA1) and it would cause a `TypeError` if you pass null.

I don't think we resolve that in userland because we don't know if `strict_types` is on or not, either:

 * We allow the typehint to be nullable in the polyfill, and accept that the behaviour of `null` with `strict_types` differs from the PHP8 implementation
 * We don't allow the typehint to be nullable in the polyfill, and accept that the behaviour of `null` without `strict_types` differs from the PHP8 implementation

stof: Well, I'm OK if the behavior for `null` differs with strict_types (`null` is a special case for types of internal functions, and its behavior cannot be replicated in userland).
But removing the typehints here and keeping them only in the PHP80 class would mean that the behavior would *also* be different when passing other types (try passing an integer)

PatrickRose:
I don't really understand the latest failures. It appears that the failing test is  `$this->assertFalse(str_ends_with(null, 'test'))`, which fails with `Polyfills enabled, substr_compare(): The start position cannot exceed initial string length` on PHP 7.0 / 7.1 - but I don't understand why `$this->assertFalse(str_ends_with($testStr, $testStr.$testStr));` would not fail for the same reason

tests/Php80/Php80Test.php
PatrickRose: I've added these, are you happy @IonBazan?

src/Php80/Php80.php
PatrickRose: Changed, thank @IonBazan!

src/Php80/bootstrap.php
PatrickRose: Thanks @stof  - I've added unit tests for the various types, and kept the typehint. Are you happy with me resolving this?

(Sorry, I don't know the etiquette for resolving review conversations on Symfony PRs!)

src/Php80/Php80.php
```diff
@@ -88,18 +88,25 @@ public static function preg_last_error_msg(): string
         }
     }
 
-    public static function str_contains(string $haystack, string $needle): bool
+    public static function str_contains(string $haystack = null, string $needle = null): bool
     {
+        // Cast to string to avoid deprecation notices
+        $needle = (string)$needle;
+
         return '' === $needle || false !== strpos($haystack, $needle);
```
IonBazan, line 96: Shouldn't we cast the `$haystack` too?

PatrickRose: According to the unit tests it's not required, though it is for `$needle`

PatrickRose:
Just pinging this - are the latest changes okay?

PatrickRose:
As discussed on the linked PR, we have to either match `strict_types=1` or not and it's preferable to match.

PR #284, opened 2020-08-12
Fix php8 error passing too many arguments to strrchr

TysonAndre:
https://www.php.net/strrchr has the signature
`strrchr ( string $haystack , mixed $needle ) : string`

Passing the unused 3rd argument becomes an ArgumentCountError in php 8.0

Detected via static analysis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 1.18.2
 
   * Add a polyfill for the `Attribute` class
+  * Fix polyfill for `mb_strrchr()`
 
 # 1.18.1
 
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -629,10 +629,11 @@ final class Mbstring
     {
         $encoding = self::getEncoding($encoding);
         if ('CP850' === $encoding || 'ASCII' === $encoding) {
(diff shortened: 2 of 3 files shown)


stof:
doesn't this mean that the behavior of `$part` is not properly polyfilled ?

TysonAndre:
> Determines which portion of haystack this function returns. If set to TRUE, it returns all of haystack from the beginning to the last occurrence of needle. If set to FALSE, it returns all of haystack from the last occurrence of needle to the end,

It looks like the polyfill was wrong and no tests covered it with true and those encodings. Will work on tests and a fix later

Ayesh:
Related: #247 
This exception is one of the many exceptions that are thrown in PHP 8. See [PHP 8.0: Internal function warnings now throw TypeError and ValueError exceptions](https://php.watch/versions/8.0/internal-function-exceptions).

All warnings raised by the original mbstring extension are now promoted to exceptions as part of the warnings promotion efforts in PHP 8.0. This among dozens of other components that already throw exceptions, and I believe we will need to scrutinize all functions in preparation for this change. 

[`ValueError` exception class](https://php.watch/versions/8.0/ValueError) is polyfilled already in #219

nicolas-grekas (approved):

nicolas-grekas:
Thank you @TysonAndre.

Issue #235, opened 2020-03-03
PHP 8.0 polyfill compatibility


IonBazan:
I'm opening this issue to discuss the ways we could assure compatibility of `symfony/polyfill-php80` with PHP versions < 7.0.
The only reason we require PHP >= 7.0 for this polyfill is `fdiv()` function that I found impossible to port to older versions and recently added type hints. However, all other features we polyfilled so far:
 - `Stringable` interface
 - `ValueError` class
 - `FILTER_VALIDATE_BOOL` constant
 - `preg_last_error_msg` function
 - `str_contains` function
 - `get_debug_type` function

would work just fine on PHP 5.3+! Maybe it would be worth considering enabling support for older versions of PHP and just conditionally disabling `fdiv` polyfill on older versions. I know it would add an extra condition in `bootstrap.php` but we should decide if we want more compatibility or more  consistency of polyfilled features.

nicolas-grekas:
We can split the class in two Php80 and Php80Fdiv. Up for a PR?

nicolas-grekas:
But we could also let things as is. PHP 5 is getting old, why spend time on it now?

GrahamCampbell:
Do we wanna polyfill https://wiki.php.net/rfc/token_as_object too.

nicolas-grekas:
Only if one needs it, so it should come by demand IMHO, not by anticipation. And the one demanding should be the implementer ideally also.

charrondev:
What about `str_starts_with` and `str_ends_with`? Looks like these were added as well. 

https://php.watch/versions/8.0/str_starts_with-str_ends_with

nicolas-grekas:
@charrondev check #250

llaville:
@GrahamCampbell to follow your comment, I agree with you, and until it will include in `polyfill-php80`, an alternative exists at https://github.com/PHPWatch/phptoken-polyfill

TysonAndre:
There's also `class Attribute` - I didn't see any issues or PRs mentioning this.

E.g. in this example snippet, it would throw an Error in php 7 and older if there was no definition for Attribute

```php
<?php

// This line crashes without a polyfill for Attribute.
const FUNCTIONLIKE_ATTRIBUTES = Attribute::TARGET_FUNCTION | Attribute::TARGET_METHOD; 

#[Attribute(FUNCTIONLIKE_ATTRIBUTES)]
function example() {
}
```

[...]

PR #286, opened 2020-09-12
Add a polyfill for the Attribute class

TysonAndre:
Code which works with both php 7 and 8 may need to use the class constants or
instances of the attribute class outside of attribute declarations. E.g.

    // This line throws without a polyfill for Attribute.
    const FUNCTIONLIKE_ATTRIBUTES = Attribute::TARGET_FUNCTION | Attribute::TARGET_METHOD;

    #[Attribute(FUNCTIONLIKE_ATTRIBUTES)]
    function example() { }

For #235

Attribute had real types in the actual constructor, is final, etc. This polyfill is currently targeting php 7+, so the real type `int $flags` was used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.18.2
+
+  * Add a polyfill for the `Attribute` class
+
 # 1.18.1
 
   * Don't force labels containing URL delimiters to stay in their Unicode form when `using idn_to_ascii`
--- a/src/Php80/Resources/stubs/Attribute.php
+++ b/src/Php80/Resources/stubs/Attribute.php
@@ -1,0 +1,22 @@
+<?php
+
+#[Attribute(Attribute::TARGET_CLASS)]
+final class Attribute
+{
+    const TARGET_CLASS = 1;
+    const TARGET_FUNCTION = 2;
+    const TARGET_METHOD = 4;
+    const TARGET_PROPERTY = 8;
+    const TARGET_CLASS_CONSTANT = 16;
+    const TARGET_PARAMETER = 32;
+    const TARGET_ALL = 63;
+    const IS_REPEATABLE = 64;
+
+    /** @var int */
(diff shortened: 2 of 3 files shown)


src/Php80/Resources/stubs/Attribute.php
```diff
@@ -0,0 +1,20 @@
...
+    const TARGET_FUNCTION = 2;
+    const TARGET_METHOD = 4;
+    const TARGET_PROPERTY = 8;
+    const TARGET_CLASS_CONSTANT = 16;
+    const TARGET_PARAMETER = 32;
+    const TARGET_ALL = 63;
+    const IS_REPEATABLE = 64;
+
+    /** @var int */
+    public $flags;
```
TysonAndre, line 15: This has a real type signature in php 8.0, but that's impractical to polyfill while supporting php versions older than php 7.4 and probably not going to matter for working php 8 applications

```
php > $x = new Attribute();
php > $x->flags = 'test';

Warning: Uncaught TypeError: Cannot assign string to property Attribute::$flags of type int in php shell code:1
Stack trace:
#0 {main}
```

nicolas-grekas (approved):

nicolas-grekas:
Thank you @TysonAndre.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.18.2
```
xabbuh, line 1: Shouldn't we bump to `1.19.0` instead?

jaswrks:
@TysonAndre @nicolas-grekas Hi! :)
Curious as to why this didn't get a PHP version check wrapped around it like the other polyfill classes?

TysonAndre:
> Curious as to why this didn't get a PHP version check wrapped around it like the other polyfill classes?

This was merged in 2020, the PR adding them was in 2021 https://github.com/symfony/polyfill/pull/359 

I expect it to be safe?

PR #287, opened 2020-10-01
Fix the name of arguments for PHP 8

azjezz:
With PHP 8 coming, we need to have function/methods parameters names consistent between polyfills and the original implementations.

I'm not sure if this is all of it yet.

see: 
https://github.com/php/php-tasks/issues/23
https://github.com/php/php-tasks/issues/16

### DIFF
--- a/src/Mbstring/bootstrap.php
+++ b/src/Mbstring/bootstrap.php
@@ -12,28 +12,28 @@
 use Symfony\Polyfill\Mbstring as p;
 
 if (!function_exists('mb_convert_encoding')) {
-    function mb_convert_encoding($s, $to, $from = null) { return p\Mbstring::mb_convert_encoding($s, $to, $from); }
+    function mb_convert_encoding($string, $to_encoding, $from_encoding = null) { return p\Mbstring::mb_convert_encoding($string, $to_encoding, $from_encoding); }
 }
 if (!function_exists('mb_decode_mimeheader')) {
-    function mb_decode_mimeheader($s) { return p\Mbstring::mb_decode_mimeheader($s); }
+    function mb_decode_mimeheader($string) { return p\Mbstring::mb_decode_mimeheader($string); }
 }
 if (!function_exists('mb_encode_mimeheader')) {
-    function mb_encode_mimeheader($s, $charset = null, $transferEnc = null, $lf = null, $indent = null) { return p\Mbstring::mb_encode_mimeheader($s, $charset, $transferEnc, $lf, $indent); }
+    function mb_encode_mimeheader($string, $charset = null, $transfer_encoding = null, $newline = null, $indent = null) { return p\Mbstring::mb_encode_mimeheader($string, $charset, $transfer_encoding, $newline, $indent); }
 }
 if (!function_exists('mb_decode_numericentity')) {
-    function mb_decode_numericentity($s, $convmap, $enc = null) { return p\Mbstring::mb_decode_numericentity($s, $convmap, $enc); }
+    function mb_decode_numericentity($string, $map, $encoding = null) { return p\Mbstring::mb_decode_numericentity($string, $map, $encoding); }
 }
 if (!function_exists('mb_encode_numericentity')) {
-    function mb_encode_numericentity($s, $convmap, $enc = null, $is_hex = false) { return p\Mbstring::mb_encode_numericentity($s, $convmap, $enc, $is_hex); }
+    function mb_encode_numericentity($string, $map, $encoding = null, $hex = false) { return p\Mbstring::mb_encode_numericentity($string, $map, $encoding, $hex); }
 }
 if (!function_exists('mb_convert_case')) {
-    function mb_convert_case($s, $mode, $enc = null) { return p\Mbstring::mb_convert_case($s, $mode, $enc); }
+    function mb_convert_case($string, $mode, $encoding = null) { return p\Mbstring::mb_convert_case($string, $mode, $encoding); }
 }
 if (!function_exists('mb_internal_encoding')) {
-    function mb_internal_encoding($enc = null) { return p\Mbstring::mb_internal_encoding($enc); }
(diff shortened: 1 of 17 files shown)


src/Php70/bootstrap.php
```diff
@@ -20,10 +20,13 @@
...
 if (!function_exists('intdiv')) {
-    function intdiv($dividend, $divisor) { return p\Php70::intdiv($dividend, $divisor); }
+    function intdiv($num1, $num2) { return p\Php70::intdiv($num1, $num2); }
 }
 if (!function_exists('preg_replace_callback_array')) {
-    function preg_replace_callback_array(array $patterns, $subject, $limit = -1, &$count = 0) { return p\Php70::preg_replace_callback_array($patterns, $subject, $limit, $count); }
+    /**
+     * $flags is not supported.
+     */
+    function preg_replace_callback_array(array $pattern, $subject, $limit = -1, &$count = 0, $flags = null) { return p\Php70::preg_replace_callback_array($pattern, $subject, $limit, $count); }
```
azjezz, line 29: PHPs `preg_replace_callback_array` has a $flags argument, but the polyfill doesn't. so I added it. this would work the same.

Adding support for $flags is unrelated to this PR.

src/Mbstring/bootstrap.php
```diff
@@ -42,88 +42,88 @@ function mb_list_encodings() { return p\Mbstring::mb_list_encodings(); }
...
 if (!function_exists('mb_output_handler')) {
-    function mb_output_handler($contents, $status) { return p\Mbstring::mb_output_handler($contents, $status); }
+    function mb_output_handler($string, $status) { return p\Mbstring::mb_output_handler($string, $status); }
 }
 if (!function_exists('mb_http_input')) {
     function mb_http_input($type = '') { return p\Mbstring::mb_http_input($type); }
 }
 if (!function_exists('mb_convert_variables')) {
-    function mb_convert_variables($toEncoding, $fromEncoding, &$a = null, &$b = null, &$c = null, &$d = null, &$e = null, &$f = null) { return p\Mbstring::mb_convert_variables($toEncoding, $fromEncoding, $a, $b, $c, $d, $e, $f); }
+    function mb_convert_variables($to_encoding, $from_encoding, &$var, &...$vars) { return p\Mbstring::mb_convert_variables($to_encoding, $from_encoding, $var, ...$vars); }
```
azjezz, line 114: this is how PHP defines `mb_convert_variables`, but this is not supported in PHP 5.3 🤔 what should we do here?

nicolas-grekas: Conditional include

azjezz: should we remove it from `Mbstring::mb_convert_variables` and implement the function alone i seperate files ( e.g: `mb_convert_variables.5.php` / `mb_convert_variables.8.php` ) ?

azjezz: &... is supported since 5.6, so the new syntax is loaded in PHP >= 5.6, this allows passing more arguments, which matches the original implementation signature.

azjezz: hm [&$var, ...$vars] is only supported in >= 7.4 ( the only way to prepend an item while keeping the reference that I know of, when using `array_unshift($vars, $var)` we lose the reference to `$var` ), so the new signature is loaded only on >= 7.4

azjezz:
we would probably have to do another round once PHP 8 is released, as more functions are being updated ATM.

src/Php70/bootstrap.php
```diff
@@ -20,10 +20,13 @@
 }
 
 if (!function_exists('intdiv')) {
-    function intdiv($dividend, $divisor) { return p\Php70::intdiv($dividend, $divisor); }
+    function intdiv($num1, $num2) { return p\Php70::intdiv($num1, $num2); }
 }
 if (!function_exists('preg_replace_callback_array')) {
-    function preg_replace_callback_array(array $patterns, $subject, $limit = -1, &$count = 0) { return p\Php70::preg_replace_callback_array($patterns, $subject, $limit, $count); }
+    /**
+     * $flags is not supported.
```
nicolas-grekas, line 27: let's remove this comment

src/Mbstring/mb_convert_variables.70400.php
```diff
@@ -0,0 +1,31 @@
...
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Mbstring as p;
+
+if (!function_exists('mb_convert_variables')) {
+    /**
+     * Convert character code in variable(s)
+     */
+    function mb_convert_variables($to_encoding, $from_encoding, &$var, &...$vars)
```
nicolas-grekas, line 18: is this the exact signature as reported by reflection since PHP 74?
let's put these extra files in the Resources folder.

src/Mbstring/bootstrap.php
```diff
@@ -42,88 +42,92 @@ function mb_list_encodings() { return p\Mbstring::mb_list_encodings(); }
...
 if (!function_exists('mb_http_input')) {
     function mb_http_input($type = '') { return p\Mbstring::mb_http_input($type); }
 }
-if (!function_exists('mb_convert_variables')) {
-    function mb_convert_variables($toEncoding, $fromEncoding, &$a = null, &$b = null, &$c = null, &$d = null, &$e = null, &$f = null) { return p\Mbstring::mb_convert_variables($toEncoding, $fromEncoding, $a, $b, $c, $d, $e, $f); }
+
+if (PHP_VERSION_ID >= 70400) {
+    require_once __DIR__ . '/mb_convert_variables.70400.php';
+} else {
+    require_once __DIR__ . '/mb_convert_variables.50400.php';
```
nicolas-grekas, line 117: no spaces around the dots

nicolas-grekas:
Thank you @azjezz.

PR #290, opened 2020-10-09
Include a php 8 bootstrap file

selcukcukur:


stof:
this change is wrong. This is covered by `src/bootstrap.php`, because the PHP80 polyfill is not compatible with PHP 5.x

selcukcukur:
> this change is wrong. This is covered by `src/bootstrap.php`, because the PHP80 polyfill is not compatible with PHP 5.x

I saw it now, thanks.

PR #291, opened 2020-10-13
Fix passing $length=null to grapheme_substr()

nicolas-grekas:


tests/Intl/Grapheme/GraphemeTest.php
```diff
@@ -89,6 +89,10 @@ public function testGraphemeSubstr()
     {
         $c = 'déjà';
 
+        if (\PHP_VERSION_ID >= 70000) {
```
stof, line 92: why only on PHP 7.0+ when the polyfill does not have such condition ?

src/Intl/Grapheme/Grapheme.php
```diff
@@ -105,8 +105,12 @@ public static function grapheme_strlen($s)
         return 0 === $len && '' !== $s ? null : $len;
     }
 
-    public static function grapheme_substr($s, $start, $len = 2147483647)
+    public static function grapheme_substr($s, $start, $len = null)
```
stof, line 108: do we even need an optional argument here ? the polyfill function always provides it.

nicolas-grekas: actually, this is required for testing.

tests/Intl/Grapheme/GraphemeTest.php
nicolas-grekas: reverted: the behavior depends also on the version of the intl extensions. The implementation is now unconditional (but tests are for practical reasons)

PR #296, opened 2020-10-20
Update branch-alias for main + v1.19

nicolas-grekas:
Waiting for the substree-splitter to sync the "main" branch before merging.


stof:
@nicolas-grekas to avoid closing existing PRs, I suggest retargeting them to `main` rather than `master` *before* deleting the master branch (either through the API, or manually as there's only a few PRs)

Issue #288, opened 2020-10-04
`iconv_substr` and `grapheme_substr` changes in PHP 8


Ayesh:
PHP 8 returns an empty string (`""`) on `iconv_substr` and `grapheme_substr` functions if the `$start` parameter exceeds the size of the provided string (`$s`). https://php.watch/versions/8.0/substr-out-of-bounds

PHP 5 through 7 native functions and Symfony polyfills for iconv (`Symfony\Polyfill\Iconv\Iconv::iconv_substr`) and Intl (`Symfony\Polyfill\Intl\Grapheme\Grapheme::grapheme_substr`) currently return `false` in this case. This is the correct behavior for PHP < 8.0, but I think it will be more complete if we were to mimic this behavior for `iconv_substr` and `grapheme_substr` if the runtime is PHP 8 or later.

I can submit PRs, but I thought it will be better to ask maintainers thoughts on this. Should we always return an empty string (which makes the polyfill not exactly compatible), or return an empty string only on PHP 8?

## Tasks

 - [x] #289 
 - [ ] #297

nicolas-grekas:
Let's return conditionally based on the runtime version, this will reduce the BC breaking potential.

Ayesh:
Thank you @nicolas-grekas. I updated the issue so we can track `iconv_substr`, `mb_susbtr`, and `grapheme_substr` functions all from one issue. I sent #289 for `iconv`, and will send another for `grapheme`. 

There are some changes in `sub(i)str` and `str(i)str` too, which I will open another issue to track.

PR #289, opened 2020-10-04
iconv: Adjust `iconv_substr` PHP 8 compatibility

Ayesh:
`iconv_substr` is changed in [PHP 8.0 to return an empty string on out-of-bound offsets](https://php.watch/versions/8.0/substr-out-of-bounds).
This commit updates the tests with `@requires` annotatios and polyfill updates to make sure the return results are in line with native function behavior.

Related: #288

PR #297, opened 2020-10-21
Intl Grapheme: Adjust `grapheme_substr` PHP 8 compatibility

Ayesh:
`grapheme_substr` is changed in [PHP 8.0 to return an empty string on out-of-bound offsets](https://php.watch/versions/8.0/substr-out-of-bounds#grapheme_substr).

This commit updates the tests with `@requires` annotations and polyfill updates to make sure the return results are in line with native function behavior. In PHP 8, an empty string (`""`) is returned if the length parameter is out of bounds of the given string. Prior to PHP 8.0, boolean `false` is returned.

Related: #288

### DIFF
--- a/src/Intl/Grapheme/Grapheme.php
+++ b/src/Intl/Grapheme/Grapheme.php
@@ -120,10 +120,14 @@ final class Grapheme
             $start += $slen;
         }
         if (0 > $start) {
-            return false;
+            if (\PHP_VERSION_ID < 80000) {
+                return false;
+            }
+
+            $start = 0;
         }
         if ($start >= $slen) {
-            return false;
+            return \PHP_VERSION_ID >= 80000 ? '' : false;
         }
 
         $rem = $slen - $start;
@@ -135,7 +139,7 @@ final class Grapheme
             return '';
         }
         if (0 > $len) {
-            return false;
+            return \PHP_VERSION_ID >= 80000 ? '' : false;
         }
         if ($len > $rem) {
             $len = $rem;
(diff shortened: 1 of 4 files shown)


Ayesh:
Apologies about the force-push noise, some tests were failing because older PHPUnit versions apparently ignore `@requires PHP >= 8.0` style annotations, so I had to use `@requires PHP 8.0` instead.

src/Intl/Grapheme/Grapheme.php
```diff
@@ -123,7 +123,7 @@ public static function grapheme_substr($s, $start, $len = null)
             return false;
         }
         if ($start >= $slen) {
-            return false;
+            return \PHP_VERSION_ID >= 80000 ? '' : false;
```
nicolas-grekas, line 126: what about L123 & L138?

Ayesh: Thanks a lot for the review @nicolas-grekas, you are right, those two `false` statements need to be changed to conditionally return empty strings on PHP 8 too. 

There was not enough test coverage for those edge cases, so the new commit extends the tests with them too: 
 - `grapheme_substr($c, -42, 0)`
 - `grapheme_substr($c, 42, 5)`

Ayesh:
I sent a PR for `iconv` before, #289, that is unfortunately incomplete. I will send a new PR to cover them too. I apologize for the time wasted and the excessive noise with those PRs, and I will be more careful to pay attention to edge cases too.

nicolas-grekas:
No worries. Please add to this very PR, that ok.

Ayesh:
Thank you, I cherry-picked the commit from #302 and closed it.

nicolas-grekas:
Thank you @Ayesh.

PR #298, opened 2020-10-21
Improve performances of array_key_last

### LINKED ISSUE #293: [improvement] array_key_last
I've note some minor performance issues on array_key_last.
A good work has been done on this pull request #150 

I think we could do a little better with array_slice instead of using both end and key function.

Here I come with [benchmarks](http://sandbox.onlinephpfunctions.com/code/5984e7ea158cbb306a9d36066967ca09e5b77ce3)

```php
function array_key_last(array $array) {
    return (!empty($array)) ? key(array_slice($array, -1, 1, true)) : null;
}
```

jderusse:
Suggested by @donaldinou   (fix #293 )

up to 6 times faster on big arrays like `array_fill(0, 1024*1024, 'hello');`

https://blackfire.io/profiles/05de3d1f-0193-462d-839f-361c4c206a0d/graph

### DIFF
--- a/src/Php73/bootstrap.php
+++ b/src/Php73/bootstrap.php
@@ -27,5 +27,5 @@ if (!function_exists('array_key_first')) {
     function array_key_first(array $array) { foreach ($array as $key => $value) { return $key; } }
 }
 if (!function_exists('array_key_last')) {
-    function array_key_last(array $array) { end($array); return key($array); }
+    function array_key_last(array $array) { key(array_slice($array, -1, 1, true)); }
 }


nicolas-grekas:
Thank you @jderusse.

src/Php73/bootstrap.php
```diff
@@ -27,5 +27,5 @@ function hrtime($asNum = false) { return p\Php73::hrtime($asNum); }
     function array_key_first(array $array) { foreach ($array as $key => $value) { return $key; } }
 }
 if (!function_exists('array_key_last')) {
-    function array_key_last(array $array) { end($array); return key($array); }
+    function array_key_last(array $array) { key(array_slice($array, -1, 1, true)); }
```
nicolas-grekas, line 30: funny that we missed the return statement :)

nicolas-grekas: fixed in df3c863

stof: isn't the testsuite catching this kind of mistakes ?

nicolas-grekas: yes it is, that's how I spotted it

stof: then we have an issue in our process if we merge PRs when the CI tells us it's broken...

jderusse: Sorry about that!

CI were broken at that time https://travis-ci.org/github/symfony/polyfill/jobs/737662800 

Did 2 mistake:
I didn't ran test locally (I often count on CI for that)
I didn't marked the PR as draft

truly sorry

PR #299, opened 2020-10-21
Update Mbstring.php

WinterSilence:
`mb_list_encodings()` add `ASCII` encoding
`mb_encoding_aliases()` add `ASCII` encoding aliases
`mb_str_split()` fix validation

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -383,7 +383,7 @@ final class Mbstring
 
     public static function mb_list_encodings()
     {
-        return array('UTF-8');
+        return array('ASCII', 'UTF-8');
     }
 
     public static function mb_encoding_aliases($encoding)
@@ -392,6 +392,8 @@ final class Mbstring
             case 'UTF8':
             case 'UTF-8':
                 return array('utf8');
+            case 'ASCII':
+                return array('cp367', 'csASCII', 'iso-ir-6', 'IBM367', 'IBM-367');
         }
 
         return false;
@@ -533,18 +535,20 @@ final class Mbstring
 
             return null;
         }
-
-        if (1 > $split_length = (int) $split_length) {
-            trigger_error('The length of each segment must be greater than zero', E_USER_WARNING);
+        if (null !== $split_length && !\is_bool($split_length) && !\is_numeric($split_length)) {
+            trigger_error('mb_str_split() expects parameter 2 to be int, '.\gettype($split_length).' given', E_USER_WARNING);
+            
+            return null;
+        }
+        $split_length = (int) $split_length;
+        if (1 > $split_length) {
+            trigger_error('mb_str_split(): The length of each segment must be greater than zero', E_USER_WARNING);
 
             return false;
         }
 
-        if (null === $encoding) {
-            $encoding = mb_internal_encoding();
-        }
-
-        if ('UTF-8' === $encoding = self::getEncoding($encoding)) {
+        $encoding = self::getEncoding($encoding);
+        if ('UTF-8' === $encoding) {
             $rx = '/(';
             while (65535 < $split_length) {
                 $rx .= '.{65535}';
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -540,15 +540,14 @@ final class Mbstring
             
             return null;
         }
-        $split_length = (int) $split_length;
-        if (1 > $split_length) {
+
+        if (1 > $split_length = (int) $split_length) {
             trigger_error('mb_str_split(): The length of each segment must be greater than zero', E_USER_WARNING);
 
             return false;
         }
 
-        $encoding = self::getEncoding($encoding);
-        if ('UTF-8' === $encoding) {
+        if ('UTF-8' === $encoding = self::getEncoding($encoding)) {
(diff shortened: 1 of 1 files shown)


src/Mbstring/Mbstring.php
```diff
@@ -392,6 +392,8 @@ public static function mb_encoding_aliases($encoding)
             case 'UTF8':
             case 'UTF-8':
                 return array('utf8');
+            case 'ASCII':
+                return array('cp367', 'csASCII', 'iso-ir-6', 'IBM367', 'IBM-367');
```
nicolas-grekas, line 396: Is that really true? I'm not sure providing any of these would make the polyfill act as if 'ascii' was provided. This would deserve some tests.

src/Mbstring/Mbstring.php
```diff
@@ -533,18 +535,20 @@ public static function mb_str_split($string, $split_length = 1, $encoding = null
...
+            trigger_error('mb_str_split(): The length of each segment must be greater than zero', E_USER_WARNING);
 
-        if (null === $encoding) {
-            $encoding = mb_internal_encoding();
+            return false;
         }
 
-        if ('UTF-8' === $encoding = self::getEncoding($encoding)) {
+        $encoding = self::getEncoding($encoding);
+        if ('UTF-8' === $encoding) {
```
nicolas-grekas, line 551: this change is not needed

src/Mbstring/Mbstring.php
```diff
@@ -533,18 +535,20 @@ public static function mb_str_split($string, $split_length = 1, $encoding = null
...
-
-            return false;
+        if (null !== $split_length && !\is_bool($split_length) && !\is_numeric($split_length)) {
+            trigger_error('mb_str_split() expects parameter 2 to be int, '.\gettype($split_length).' given', E_USER_WARNING);
+            
+            return null;
         }
+        $split_length = (int) $split_length;
+        if (1 > $split_length) {
+            trigger_error('mb_str_split(): The length of each segment must be greater than zero', E_USER_WARNING);
```
nicolas-grekas, line 545: can you add tests for the two new warnings please?

src/Mbstring/Mbstring.php
```diff
@@ -533,18 +535,20 @@ public static function mb_str_split($string, $split_length = 1, $encoding = null
...
 
-        if (null === $encoding) {
```
nicolas-grekas, line 543: where did this go? please add a test if possible

src/Mbstring/Mbstring.php
```diff
@@ -533,18 +535,20 @@ public static function mb_str_split($string, $split_length = 1, $encoding = null
...
-            trigger_error('The length of each segment must be greater than zero', E_USER_WARNING);
-
-            return false;
+        if (null !== $split_length && !\is_bool($split_length) && !\is_numeric($split_length)) {
+            trigger_error('mb_str_split() expects parameter 2 to be int, '.\gettype($split_length).' given', E_USER_WARNING);
+            
+            return null;
         }
+        $split_length = (int) $split_length;
+        if (1 > $split_length) {
```
nicolas-grekas, line 544: cs change should be reverted

nicolas-grekas (commented):
can you provide some background as to why this should be needed?

src/Mbstring/Mbstring.php
WinterSilence: @nicolas-grekas more readable

nicolas-grekas: unrelated and subjective, complicates the review

src/Mbstring/Mbstring.php
nicolas-grekas: but this has a technical meaning, we don't want to list aliases that don't work
these don't work, so we don't list them.

src/Mbstring/Mbstring.php
derrabus: Whether this is more readable is probably debatable, but that question is out of scope for this PR.

If the actual change is hidden among CS fixes, a PR is really hard to review.

src/Mbstring/Mbstring.php
WinterSilence: @nicolas-grekas
~~~php
$split_length = 1.2;
var_dump(1.1 > $split_length = (int) $split_length); // bool(true) 1.1>1
~~~

nicolas-grekas: Since this is a polyfill, the correct behavior is easy to decide: how does the native function behave? To verify and prevent regressions, the correct behavior should be added as a test case, if one is missing.

WinterSilence: @nicolas-grekas need fix ci configs before update tests

nicolas-grekas:
@WinterSilence could you please tell us about the status of this PR? For which reason did you submit it? Which issue does it solve? Also, do you plan to add tests to it?

WinterSilence:
@nicolas-grekas work in progress

WinterSilence:
@nicolas-grekas 
> Also, do you plan to add tests to it?

As I have already said:

> "All checks have failed" - need fix CI configs before update tests, but I can't do it

nicolas-grekas:
I'm closing for now, please submit again when you're ready!

PR #300, opened 2020-10-21
Fix Travis build for the main branch

derrabus:
Our Travis configuration does not reflect the new main branch. Let's fix this.


nicolas-grekas:
Thank you @derrabus.

PR #303, opened 2020-10-23
Bump minimum PHP version to 7.1

nicolas-grekas:
I propose to bump the minimum PHP version to ~7.2~ 7.1
~We would keep `main` aliased to `1.19.x-dev` and we would only provide bug fixes for it.~


stof:
Why making PHP 7.2 the min version for the 1.x version of polyfills ? This won't help with the need to change all error handling in PHP 8 AFAICT.

nicolas-grekas:
> Why making PHP 7.2 the min version for the 1.x version of polyfills ?

That's a bit arbitrary. 7.2 matches the current min of Symfony 5.

> This won't help with the need to change all error handling in PHP 8 AFAICT.

Nope indeed, that's loosely related (it will help prepare the CI and cleanup the codebase a bit, nothing more)

stof:
> That's a bit arbitrary. 7.2 matches the current min of Symfony 5.

But Symfony 4.4 is still maintained for years to come, and relies on polyfills. Shouldn't we at least take it into account ?

nicolas-grekas:
I think we do, by continuing to maintain `1.19.x-dev`.

stof:
@nicolas-grekas in the past, we have added usage of new polyfills in patch releases of LTS versions when adding support of new PHP versions. If `1.19.x-dev` receives only bugfixes, this pattern won't work anymore as the new polyfills won't exist in it.

nicolas-grekas:
> in the past, we have added usage of new polyfills in patch releases of LTS versions when adding support of new PHP versions. If 1.19.x-dev receives only bugfixes, this pattern won't work anymore as the new polyfills won't exist in it.

I understand, I think that's fine for 4.4: I don't anticipate that a polyfill for 8.1+ will be absolutely required for 4.4. We'll just skip using that polyfill if one is created. At least I think that the benefit of empty packages for older PHP versions outweigh the drawback for 4.4.

stof:
@nicolas-grekas do we win a lot in terms of code cleanup for the polyfill codebase by going 7.2+ compared to going 7.1.3+ (the min version of Symfony 4) ?

stof:
> I understand, I think that's fine for 4.4: I don't anticipate that a polyfill for 8.1+ will be absolutely required for 4.4. We'll just skip using that polyfill if one is created.

The reason we use polyfills in LTS versions is that we migrate to new functions (like `get_debug_type` for instance) in all maintained branches to reduce on-going merge conflicts. So to me, forbidding to use new polyfills in the LTS version would actually hurt our maintenance, unless we also avoid using it in the newest branch...

fabpot:
I'm ok to drop all PHP versions (up to 7.2), but we don't need to "maintain" 1.19 IMHO. Why not support 7.1 for 4.4? First, because the code is already old and stable (no known bugs), and if any, people should really update their PHP version instead (I don't see why we would need to maintain code for apps using obsolete PHP versions).

stof:
@fabpot if we make a new patch release of 4.4 depend on a `symfony/polyfill-php81` when adding support for PHP 8.1 in the 4.4 branch, and that `symfony/polyfill-php81` does not have a version supporting PHP 7.1, it means we are effectively bumping the min version of the 4.4 LTS as composer won't be able to resolve dependencies anymore on PHP 7.1. And the Symfony LTS policy does not allow us to bump the min PHP version AFAIK.

nicolas-grekas:
Let me try downgrading to 7.1

composer.json
```diff
@@ -16,15 +16,15 @@
...
     "require": {
-        "php": ">=5.3.3",
-        "ircmaxell/password-compat": "~1.0",
-        "paragonie/random_compat": "~1.0|~2.0|~9.99",
+        "php": ">=7.2",
         "symfony/intl": "~2.3|~3.0|~4.0|~5.0"
     },
     "require-dev": {
-        "symfony/phpunit-bridge": "~3.2"
+        "symfony/phpunit-bridge": "^4.4"
```
stof, line 23: this should even be `^4.4 || ^5.0` (or maybe only `^5.0` if that's fine for the min PHP version), to match the way we use the bridge in Symfony itself: we use the latest version even when testing older branches (which is why `symfony/phpunit-bridge` has a separate policy for the PHP support)

src/Intl/Idn/composer.json
```diff
@@ -20,10 +20,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3",
-        "symfony/polyfill-intl-normalizer": "^1.10",
-        "symfony/polyfill-php70": "^1.10",
-        "symfony/polyfill-php72": "^1.10"
+        "php": ">=7.2",
+        "symfony/polyfill-intl-normalizer": "^1.10|^2"
```
stof: why `^2` if we don't create a 2.x version ?

src/Php55/composer.json
```diff
@@ -16,18 +16,13 @@
...
     "require": {
-        "php": ">=5.3.3",
-        "ircmaxell/password-compat": "~1.0"
+        "php": ">=7.2"
     },
-    "autoload": {
-        "psr-4": { "Symfony\\Polyfill\\Php55\\": "" },
-        "files": [ "bootstrap.php" ]
+    "replace": {
+        "ircmaxell/password-compat": "*"
```
stof, line 22: I don't think we need to replace that package. We should just stop requiring it

stof (commented):
Old polyfill packages which have no code anymore could use `type: metapackage` so composer does not download a useless zip containing only a composer.json file.

composer.json
nicolas-grekas: updated

src/Php55/composer.json
nicolas-grekas: Since we guarantee the minimum php version, we can and should replace it to skip the package IMHO.

src/Intl/Idn/composer.json
nicolas-grekas: reverted

nicolas-grekas:
> Old polyfill packages which have no code anymore could use type: metapackage

updated 

ready for 7.1

nicolas-grekas:
PR is ready.

src/Php55/composer.json
derrabus: Neither do I think that this is a good idea. This setting prevents the installation of `ircmaxell/password-compat` with no way around it. We don't own `ircmaxell/password-compat` and it shouldn't be up to us to decide if an application is allowed to install it or not.

Or let's put it that way:

* We just remove the dependency. Worst case: Some other library still requires it and we end up with an unnecessary package. Damage: zero.

* We declare that we replace the package. Worst case: An app or some other library relies on internals of that package (it shouldn't, but who am I to judge) and the app breaks.

I'd prefer to just remove `ircmaxell/password-compat` from the dependencies as @stof suggested.

src/Php55/composer.json
```diff
@@ -16,17 +16,15 @@
         }
     ],
     "require": {
-        "php": ">=5.3.3",
-        "ircmaxell/password-compat": "~1.0"
+        "php": ">=7.1"
```
derrabus, line 19: On the metapackage we could even go lower:

```suggestion
        "php": ">=5.6"
```

But the current constraint is fine as well.

tests/Php72/Php72Test.php
```diff
@@ -109,8 +109,14 @@ public function testSapiWindowsVt100SupportWarnsOnInvalidInputType()
...
-        $this->setExpectedException('PHPUnit\Framework\Error\Warning', 'expects parameter 1 to be resource');
-        sapi_windows_vt100_support('foo', true);
+        try {
+            sapi_windows_vt100_support('foo', true);
+        } catch (\PHPUnit\Framework\Error\Warning $e) {
+            $this->expectException('PHPUnit\Framework\Error\Warning');
+            $this->expectExceptionMessage('expects parameter 1 to be resource');
+
+            throw $e;
+        }
```
derrabus, line 119: I don't quite understand these changes. Are we excepting a warning or are we not?

.travis.yml
```diff
@@ -36,9 +23,8 @@ before_install:
     - echo session.gc_probability = 0 >> $INI
     - echo apc.enable_cli = 1 >> $INI
     - phpenv config-rm xdebug.ini || echo "xdebug not available"
-    - if [[ $TRAVIS_PHP_VERSION = 5.* ]]; then echo yes | pecl install -f apcu-4.0.11; fi
-    - if [[ $TRAVIS_PHP_VERSION = 7.* ]]; then echo yes | pecl install -f apcu_bc-1.0.4; fi
-    - if [[ $TRAVIS_PHP_VERSION = 7.* ]]; then echo yes | pecl install -f apcu-5.1.17; fi
+    - yes | pecl install -f apcu_bc-1.0.4
+    - yes | pecl install -f apcu-5.1.17
```
derrabus, line 27:
```suggestion
    - yes | pecl install -f apcu-5.1.19
```

If we want this to work on nightly as well.

nicolas-grekas: not needed for this PR

tests/Php72/Php72Test.php
nicolas-grekas: We are. But we also want skips to work, and this is how.

src/Php55/composer.json
nicolas-grekas: Nope: that would conflict with what symfony/polyfill tells.

src/Php55/composer.json
nicolas-grekas: Fine, updated.

src/Php54/README.md
```diff
@@ -1,17 +0,0 @@
-Symfony Polyfill / Php54
```
derrabus, line 1: We should keep a nice README even for the now empty packages. The README is rendered on the split repos on GitHub as well as on the packagist page of the package.

src/Php55/composer.json
derrabus: All right.

src/Php54/README.md
nicolas-grekas: readme files are back

.travis.yml
derrabus: Well, the current Travis config won't install APCu at all on nightly. Your version would try to install an incompatible APCu. imho, we should either bump APCu or restore the `if`.

nicolas-grekas: good call, bumped

tests/Php72/Php72Test.php
derrabus: As long as `markTestSkipped()` is called before `expectException()`, we should be good, shouldn't we?

nicolas-grekas: As long as yes, that's not the case here :)
The test listener creates a dummy implementation of the polyfilled function when the native one is not available, that throws. It used to work seamlessly, but doesn't anymore apparently after bumping phpunit.

derrabus: That's odd. Well okay, let's leave it ad it is for now and investigate later. 🙃

derrabus (approved):
Travis is still sleeping. But the tests passed on my machine.

jrfnl:
As the changes from this PR were merged before PR #308, the changes made here were still applied to the stand-alone packages, including tagging a release, basically making the stand-alone packages for PHP < 7.2 useless and uninstallable unless the version is fixed at `1.1.9`.

derrabus:
@jrfnl Could you elaborate on what happens if you try to install the polyfill packages on php 7.0 and below? My expectation is that the packes should be installed and locked at 1.19.0. This is not what happens?

derrabus:
@jrfnl I've tested this locally now with the following composer.json file:

```JSON
{
    "name": "derrabus/php56-test",
    "type": "project",
    "license": "MIT",
    "require": {
        "php": ">=5.6",
        "symfony/polyfill-php70": "^1"
    },
    "config": {
        "platform": {
            "php": "5.6.40"
        }
    }
}
```

Running `composer update` on this file gives me the following output:

```
Updating dependencies
Lock file operations: 2 installs, 0 updates, 0 removals
  - Locking paragonie/random_compat (v2.0.19)
  - Locking symfony/polyfill-php70 (v1.19.0)
Writing lock file
```

This looks okay to me.

jrfnl:
@derrabus Hi Alexander, thanks for taking a look. Not sure what happened, might have been a project with a loose version constraint like `*`.

Either way, even if the install is ok, it still seems very weird that the repos are now empty and have tagged releases on the empty versions. I imagine that will also make future maintenance of those polyfills less intuitive if ever they will need an update.

derrabus:
Thank you for your feedback, Juliette. Yes, we would basically have to tag 1.19.x releases and work on the split repo directly. But I don't expect many changes to those polyfills in the future. They should be pretty much feature complete.

PR #304, opened 2020-10-23
Code cleanup: Remove redundant else/elseif blocks

Ayesh:
A very minor improvement in code clarity. Removed `else` and `elseif` blocks where the previous code block has `return`, `continue`, or `throw` calls, where the execution flaw is halted anyway. 

Thank you.


nicolas-grekas:
Thank you @Ayesh.

PR #301, opened 2020-10-21
Fix CI for php 8

derrabus:
Our tests currently don't work on php 8. I'd like to give it a try to fix this.

PR #305, opened 2020-10-23
GitHub Actions

derrabus:
This PR ports our CI to GitHub Actions. While Travis still considers if it should boot the VMs for our CI, the run on GH Actions has been completed already.

* This PR includes #303, so please ignore the first commit when reviewing
* The PHP 8 run fails (as it does on Travis). I'll try to resolve this in #301.


derrabus:
GH Actions seem to be disabled on this repo. Here's the latest run on my fork: https://github.com/derrabus/polyfill/runs/1298646165

.github/workflows/tests.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+
+    test:
+        name: PHPUnit Tests
+        runs-on: ubuntu-latest
+
+        env:
+            COMPOSER_ROOT_VERSION: dev-main
+            SYMFONY_PHPUNIT_REMOVE_RETURN_TYPEHINT: 1
```
derrabus, line 15: Oh, that shouldn't be necessary anymore. I'll remove it.

nicolas-grekas:
Dunno why they didn't run. Settings tell they are enabled.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+
+        steps:
+            -   name: Checkout
+                uses: actions/checkout@v2
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
+                with:
+                    coverage: "none"
+                    extensions: "apcu, uuid"
```
derrabus, line 34: 💡 We could play with this setting in the matrix, so run the tests with and without the polyfilled extensions.

nicolas-grekas: Not needed, that wouldn't provide anything

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+                    - '8.0'
+            fail-fast: false
+
+        steps:
+            -   name: Checkout
+                uses: actions/checkout@v2
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
+                with:
```
jderusse, line 32: consider `tools: composer:v2` until release of composer for a faster build

stof:
> Dunno why they didn't run. Settings tell they are enabled.

IIRC, the workflow files need to be present in the target branch to allow the job to run. A PR *adding* the CI workflow won't trigger it.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+
+        strategy:
+            matrix:
+                php:
+                    - '7.1'
+                    - '7.2'
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
+            fail-fast: false
```
jderusse, line 24: this is opposite to previous conf in travis

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+        steps:
+            -   name: Checkout
+                uses: actions/checkout@v2
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
+                with:
+                    coverage: "none"
+                    extensions: "apcu, uuid"
+                    ini-values: "memory_limit=-1, apc.enable_cli=1"
```
jderusse, line 35: is `session.gc_probability = 0` not needed anymore?

derrabus:
> IIRC, the workflow files need to be present in the target branch to allow the job to run. A PR _adding_ the CI workflow won't trigger it.

Oh my. Well then, you can click on `derrabus:improvement/gh-actions` at the top of this page to see the latest run.

stof:
The failure on PHP 8 is because it installs PHPUnit 9.3 which is not yet supported by the bridge (I think 5.2-dev supports it but I'm not sure). Travis was forcing `SYMFONY_PHPUNIT_VERSION=7.2`

.github/workflows/tests.yml
derrabus: The `fast_finish` setting on Travis works slightly different than `fail-fast`, IIRC.

If I did not disable `fail-fast`, GH actions would cancel the whole matrix immediately if one of the matrix builds fail. That way, we would never know if we deal with a general failure or a version-specific one.

Right now this would mean: The php 8 build fails and the others are cancelled.

.github/workflows/tests.yml
derrabus: Good catch, added.

.github/workflows/tests.yml
derrabus: You're right. This should be safe now.

derrabus:
> The failure on PHP 8 is because it installs PHPUnit 9.3 which is not yet supported by the bridge (I think 5.2-dev supports it but I'm not sure).

PhpUnitBridge 5.2-dev is actually being used. That's why we get PHPUnit 9.3 in the first place.

The listener is the problem, see https://github.com/symfony/symfony/issues/33331#issuecomment-713703230

> Travis was forcing `SYMFONY_PHPUNIT_VERSION=7.2`

Yes, but Nicolas' PR removes that already. PHPUnit 7.2 won't work either because it is not compatible with PHP 8.

nicolas-grekas:
Thank you @derrabus.

PR #306, opened 2020-10-24
GitHub Actions - Windows

jderusse:
Follow  #305 to use Github Action for Windows.

Having everything in a single place should simplify maintenance on build scripts and ease fixing the CI


nicolas-grekas:
Note a fundamental difference between appveyor and gha here: appveyor runs a 32b version of php. This is on purpose.
Now, the gha jobs fails, that should be fixed :)

jderusse:
> Note a fundamental difference between appveyor and gha here: appveyor runs a 32b version of php. This is on purpose.

Fixed. By downloading the win 32 binary. But I'face a strange issue: When running `php ./vendor/bin/simple-phpunit install` the step is stuck. (same when I run `php ./vendor/bin/simple-phpunit`. Tests a re played (and failed) bu the step never finish.) Looks like the process is still running.

jderusse:
I think we are good on this.

3 edge cases:
- when APCU is enabled, running `php.exe a.php` which containes `<?php exec('php.exe b.php');` stuck GHA job.
- microtime in Windows tests suite is broken!
- phpunit does not take `@require` into account when counting number of tests. That's why it performs only  5% of tests)

tests/Php73/Php73Test.php
```diff
@@ -53,6 +53,16 @@ public function testHardwareTimeAsNumType()
...
     public function testHardwareTimeAsNum()
     {
+        // time is not reliable in tests on Windows.
+        // reproduce case:
+        //   $a = 1E6 * (float) microtime(false);
+        //   usleep(1000);
+        //   $b = 1E6 * (float) microtime(false);
+        //   $this->assertGreaterThanOrEqual(1000, $b-$a); // randomly false on windows!
+        if ('\\' === DIRECTORY_SEPARATOR) {
+            $this->markTestSkipped('Test base on time are not reliable on windows');
```
nicolas-grekas, line 63: `Tests based on time are not reliable on Windows.` (same below)

would `@group time-sensitive` help?

.github/workflows/tests.yml
```diff
@@ -41,3 +41,113 @@ jobs:
...
+                    echo extension_dir=ext >> php.ini
+                    echo extension=php_openssl.dll >> php.ini
+                    echo extension=php_apcu.dll >> php.ini
+                    echo extension=php_intl.dll >> php.ini
+                    echo extension=php_mbstring.dll >> php.ini
+                    echo extension=php_ldap.dll >> php.ini
+                    echo ::endgroup::
+
+                    echo ::group::Download-composer
+                    curl.exe -fsL https://github.com/composer/composer/releases/download/2.0.0/composer.phar -o composer.phar
```
stof, line 101: why hardcoding 2.0.0 by taking it from github release instead of taking `https://getcomposer.org/composer-stable.phar` which gets the latest stable release ? 2.0.0 is already outdated by 2 releases.

tests/Intl/Idn/IdnTest.php
```diff
@@ -76,7 +76,7 @@ public function getData()
         $tests = array();
 
         while (false !== ($line = fgets($h))) {
-            if ("\n" === $line || '#' === $line[0]) {
+            if (PHP_EOL === $line || '#' === $line[0]) {
```
stof, line 79: this looks wrong to me. If you don't alter the newlines when cloning the code, the file will *not* use Windows line endings and this check will break.

jderusse: well, In fact this change fix the tests. The line ends with `\r\n` in windows (tested on a Windows 10 laptop and Github Action)

```
while (false !== ($line = fgets($h))) {
  var_dump(json_encode($line));
```
linux: `string(20) ""# IdnaTestV2.txt\n""`
windows: `string(22) ""# IdnaTestV2.txt\r\n""`

I can add `if ("\n" === $line || "\r" === $line || "\r\n" === $line || '#' === $line[0]) {` if you are not confortable.

.github/workflows/tests.yml
jderusse: thanks

tests/Intl/Idn/IdnTest.php
stof: maybe Github Actions configures git with `core.auto_crlf=true` while Appveyor does not.

But in any case, our tests should keep working if someone uses `core.auto_crlf=input` or `core.auto_crlf=false` (`core.auto_crlf=input` is IMO a much saner value on Windows as it makes you work on the original source code, which works fine in all Windows editor except old versions of Notepad.exe).

I would go with `"\n" === $line || PHP_EOL === $line`, or maybe `"\n" === $line || "\r\n" === $line`. no need to care about standalone `\r`. this is dead since at least a decade.

jderusse: good catch for `core.auto_crlf`.. Fixed

tests/Php73/Php73Test.php
```diff
@@ -51,6 +51,28 @@ public function testHardwareTimeAsNumType()
...
+    public function testHardwareTimeAsNum73()
+    {
+        // time is not reliable in tests on Windows.
+        // reproduce case:
+        //   $a = 1E6 * (float) microtime(false);
+        //   usleep(1000);
+        //   $b = 1E6 * (float) microtime(false);
+        //   $this->assertGreaterThanOrEqual(1000, $b-$a); // randomly false on windows!
+        if ('\\' === DIRECTORY_SEPARATOR) {
+            $this->markTestSkipped('Test based on time are not reliable on Windows');
```
nicolas-grekas, line 66: Tests (missing "s", same below)

tests/Php73/Php73Test.php
```diff
@@ -51,6 +51,28 @@ public function testHardwareTimeAsNumType()
...
+     */
+    public function testHardwareTimeAsNum73()
+    {
+        // time is not reliable in tests on Windows.
+        // reproduce case:
+        //   $a = 1E6 * (float) microtime(false);
+        //   usleep(1000);
+        //   $b = 1E6 * (float) microtime(false);
+        //   $this->assertGreaterThanOrEqual(1000, $b-$a); // randomly false on windows!
+        if ('\\' === DIRECTORY_SEPARATOR) {
```
nicolas-grekas, line 65: missing `\` `DIRECTORY_SEPARATOR` (same below)

jderusse: fixed

nicolas-grekas:
I'm closing because this is not critical, appveyor works well.

PR #310, opened 2020-10-31
Import symfony/intl polyfills

nicolas-grekas:
Fixes https://github.com/symfony/symfony/issues/37758

This is currently blocked by `NumberFormatter`, which uses the `Currencies` to read currency data.

TODO:
- [x] update the namespace
- [x] move tests to the `tests/` folder of symfony/polyfill


nicolas-grekas:
Except for a few red tests on appveyor, this PR is green.
Only the last commit should be reviewed as all the other ones are extracted from the git history of symfony/intl

/cc @stof et al.

src/Intl/Icu/composer.json
```diff
@@ -0,0 +1,42 @@
...
+    "autoload": {
+        "files": [ "bootstrap.php" ],
+        "psr-4": { "Symfony\\Polyfill\\Intl\\Icu\\": "" },
+        "classmap": [ "Resources/stubs" ],
+        "exclude-from-classmap": [
+            "/Tests/"
+        ]
+    },
+    "suggest": {
+        "ext-intl": "For best performance"
```
stof, line 30: I would say `For best performance and support of other locales than "en"`.

tests/Intl/Icu/CurrenciesTest.php
```diff
@@ -0,0 +1,37 @@
...
+            $data[$code] = [$symbol];
+        }
+
+        foreach ($meta['Meta'] as $code => [$fractionDigit, $roundingIncrement]) {
+            $data[$code] = ($data[$code] ?? []) + [1 => $fractionDigit, $roundingIncrement];
+        }
+
+        $data = "<?php\n\nreturn ".var_export($data, true).";\n";
+
+        $this->assertStringEqualsFile(\dirname(__DIR__, 3).'/src/Intl/Icu/Resources/currencies.php', $data);
```
stof, line 35: should we assert the exact content of the file or the content of the array ? The current assertion might make the test fragile if PHP changes the exact formatting used by `var_export` between versions.

src/Intl/Icu/composer.json
nicolas-grekas: updated

tests/Intl/Icu/CurrenciesTest.php
nicolas-grekas: I'm going to keep it that way because that will ease synching the file. `var_export()` is pretty stable also.

nicolas-grekas:
appveyor is now green, PR ready.

stof (approved):

PR #314, opened 2020-11-13
Optimize grapheme polyfill

nikic:
grapheme_position() currently uses iconv/mbstring strpos implementations to look for the needle. However, this is not necessary: For UTF-8, we can use the normal strpos functions (as long as we also do the following substr using the normal functions).

This cuts off 25% of the runtime from symfony_demo /en/blog without intl. iconv_strpos is *extremely* slow.

### DIFF
--- a/src/Intl/Grapheme/Grapheme.php
+++ b/src/Intl/Grapheme/Grapheme.php
@@ -205,13 +205,25 @@ final class Grapheme
             }
         }
 
-        switch ($mode) {
-            case 0: $needle = iconv_strpos($s, $needle, 0, 'UTF-8'); break;
-            case 1: $needle = mb_stripos($s, $needle, 0, 'UTF-8'); break;
-            case 2: $needle = iconv_strrpos($s, $needle, 'UTF-8'); break;
-            default: $needle = mb_strripos($s, $needle, 0, 'UTF-8'); break;
+        // As UTF-8 is self-synchronizing, and we have ensured the strings are valid UTF-8,
+        // we can use normal binary string functions here. For case-insensitive searches,
+        // case fold the strings first.
+        $caseInsensitive = $mode & 1;
+        $reverse = $mode & 2;
+        if ($caseInsensitive) {
+            // Use the same case folding mode as mbstring does for mb_stripos().
+            // Stick to SIMPLE case folding to avoid changing the length of the string, which
+            // might result in offsets being shifted.
+            $mode = \defined('MB_CASE_FOLD_SIMPLE') ? \MB_CASE_FOLD_SIMPLE : \MB_CASE_UPPER;
+            $s = mb_convert_case($s, $mode, 'UTF-8');
+            $needle = mb_convert_case($needle, $mode, 'UTF-8');
+        }
+        if ($reverse) {
+            $needlePos = strrpos($s, $needle);
+        } else {
+            $needlePos = strpos($s, $needle);
         }
 
-        return false !== $needle ? self::grapheme_strlen(iconv_substr($s, 0, $needle, 'UTF-8')) + $offset : false;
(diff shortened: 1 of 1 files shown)


stof:
Another alternative could be to use `mb_substr`

@nikic would the mbstring polyfill benefit from a similar optimization when using the UTF-8 encoding (which might be the common use case for mbstring) ?

nikic:
> Another alternative could be to use `mb_substr`

Yeah, that would also be much better than iconv_substr (as long as mb_substr is not also polyfilled by iconv_substr). In the end, mb_substr effectively also does an internal conversion to UTF-8 and a normal substr, plus adjustment of offsets. If there's no need to have a character offset, as in this case, it's better to just directly use substr.

> @nikic would the mbstring polyfill benefit from a similar optimization when using the UTF-8 encoding (which might be the common use case for mbstring) ?

Possibly. I've had mbstring enabled when testing, so didn't notice anything there. substr can be used to find the position of the string, but one would still have to convert the returned byte offset into a character offset. That part would still have to use some inefficient method (though only if the string is actually found).

stof:
I'm fine using `substr` directly in this polyfill. But I indeed suggest optimizing the UTF-8 case in the mbstring polyfill (which uses iconv for the fallback).

Side note, why is `iconv_substr` so slow compared to others ?

src/Intl/Grapheme/Grapheme.php
```diff
@@ -205,13 +205,24 @@ private static function grapheme_position($s, $needle, $offset, $mode)
...
-            case 2: $needle = iconv_strrpos($s, $needle, 'UTF-8'); break;
-            default: $needle = mb_strripos($s, $needle, 0, 'UTF-8'); break;
+        // As UTF-8 is self-synchronizing, and we have ensured the strings are valid UTF-8,
+        // we can use normal binary string functions here. For case-insensitive searches,
+        // case fold the strings first.
+        $caseInsensitive = $mode & 1;
+        $reverse = $mode & 2;
+        if ($caseInsensitive) {
+            // Use the same case folding mode as mbstring does for mb_stripos().
+            // We could also use MB_CASE_FOLD (without _SIMPLE) for better results.
```
nicolas-grekas, line 215: why don't we do it if it yields better results?

src/Intl/Grapheme/Grapheme.php
```diff
@@ -205,13 +205,24 @@ private static function grapheme_position($s, $needle, $offset, $mode)
...
+            $mode = \defined('MB_CASE_FOLD_SIMPLE') ? \MB_CASE_FOLD_SIMPLE : \MB_CASE_UPPER;
+            $s = mb_convert_case($s, $mode, 'UTF-8');
+            $needle = mb_convert_case($needle, $mode, 'UTF-8');
+        }
+        if ($reverse) {
+            $needle_pos = strrpos($s, $needle);
+        } else {
+            $needle_pos = strpos($s, $needle);
```
nicolas-grekas, line 223: $needlePos please :)

src/Intl/Grapheme/Grapheme.php
nikic: I'm not sure that would actually be okay in this context ... For example, let's say we have a string `XßYß` and search for `SS` from the end. Then non-simple case-folding would convert the string into something like `XSSYSS` and we'd find it. But at that point the grapheme offset to the second SS would be 4, rather than the 3 you would expect based on the original string.

So it seems safer to stick with SIMPLE folding, which does not change the length of the string.

nicolas-grekas: makes sense, thanks!

nicolas-grekas:
Thank you @nikic.

PR #316, opened 2020-12-26
Verify the signature of polyfills on PHP >= 8

nicolas-grekas:


nicolas-grekas:
PR is ready. I would appreciate help to fix the failures on PHP 8 once it's merged :pray:

Ayesh:
> Errors: 54, Failures: 18, Warnings: 2

This is much more manageable and miles ahead of what we had before. Thank you for this amazing work @nicolas-grekas 🙏🏼. I will try to work on further fixes in coming days.

src/Apcu/bootstrap.php
```diff
@@ -15,37 +15,41 @@
     return;
 }
 
+if (PHP_VERSION_ID >= 80000) {
+    return require __DIR__.'/bootstrap.php8';
```
derrabus, line 19: Encoding the PHP version in the file suffix went out of fashion since the good old php3 days. This file extension is a bit surprising and has already caused interesting side effects (FriendsOfPHP/PHP-CS-Fixer#5324). I would appreciate if we would find a better naming convention and use the standard `.php` file extension for PHP code.

derrabus:
> I would appreciate help to fix [the failures](https://github.com/symfony/polyfill/pull/316/checks?check_run_id=1612105563) on PHP 8 once it's merged 🙏

🙋🏻

I'll pick (*rolls dice*) the UUID polyfill.

src/Apcu/bootstrap.php
nicolas-grekas: updated

PR #319, opened 2020-12-27
Skip tests of the 7.2/7.4 polyfills that cause Errors on PHP 8

derrabus:
Some functions that were added with PHP 7.2/7.4 have been made a bit stricter with PHP 8: They now trigger `TypeError`s and `ValueError`s on invalid input.

I think the PHP 7.2 polyfill should implement the polyfilled functions as they were shipped with PHP 7.2. Because of that, I think we can skip the affected tests on PHP 8.


nicolas-grekas:
Thank you @derrabus.

PR #321, opened 2021-01-01
Adjust intl polyfills for PHP 8

derrabus:


src/Intl/Grapheme/Grapheme.php
```diff
@@ -61,7 +61,15 @@ public static function grapheme_extract($s, $size, $type = GRAPHEME_EXTR_COUNT,
         $type = (int) $type;
         $start = (int) $start;
 
-        if (!isset($s[0]) || 0 > $size || 0 > $start || 0 > $type || 2 < $type) {
+        if ($type !== GRAPHEME_EXTR_COUNT && $type !== GRAPHEME_EXTR_MAXBYTES && $type !== GRAPHEME_EXTR_MAXCHARS) {
```
nicolas-grekas, line 64: yoda :)

derrabus: Correct you are.

nicolas-grekas:
Thank you @derrabus.

src/Intl/Grapheme/Grapheme.php
```diff
@@ -181,7 +189,7 @@ public static function grapheme_strstr($s, $needle, $beforeNeedle = false)
     private static function grapheme_position($s, $needle, $offset, $mode)
     {
         $needle = (string) $needle;
-        if (!preg_match('/./us', $needle)) {
+        if (80000 > \PHP_VERSION_ID && !preg_match('/./us', $needle)) {
```
stof, line 192: what about the similar regexp 4 lines later on the string ?

And what happens in PHP 8 for non-unicode strings ?

derrabus: The corresponding change in PHP was that an empty needle is now allowed which caused this check to fail. Now that you bring it up, I think I should have solved this differently. I'll submit a PR in the evening.

glensc:
I think the checks are flipped:
- https://github.com/symfony/polyfill/pull/347

or at least must be modified to include 8.0.0

since the code has no comments and commits don't explain the reasoning behind the changes, I'm not certain what this is supposed to do or fix.

PR #324, opened 2021-01-06
Add polyfill-php81

nicolas-grekas:
https://wiki.php.net/rfc/is_list

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,6 +30,7 @@
         "symfony/polyfill-php73": "self.version",
         "symfony/polyfill-php74": "self.version",
         "symfony/polyfill-php80": "self.version",
+        "symfony/polyfill-php81": "self.version",
         "symfony/polyfill-iconv": "self.version",
         "symfony/polyfill-intl-grapheme": "self.version",
         "symfony/polyfill-intl-icu": "self.version",
@@ -47,9 +48,6 @@
             "src/bootstrap.php",
             "src/Apcu/bootstrap.php",
             "src/Ctype/bootstrap.php",
-            "src/Php72/bootstrap.php",
-            "src/Php73/bootstrap.php",
-            "src/Php74/bootstrap.php",
             "src/Uuid/bootstrap.php",
             "src/Iconv/bootstrap.php",
             "src/Intl/Grapheme/bootstrap.php",
@@ -70,7 +68,7 @@
     "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
-            "dev-main": "1.21-dev"
+            "dev-main": "1.22-dev"
         }
     }
 }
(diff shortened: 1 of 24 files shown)


src/Php81/composer.json
```diff
@@ -0,0 +1,39 @@
...
+    "type": "library",
+    "description": "Symfony polyfill backporting some PHP 8.1+ features to lower PHP versions",
+    "keywords": ["polyfill", "shim", "compatibility", "portable"],
+    "homepage": "https://symfony.com",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Ion Bazan",
+            "email": "ion.bazan@gmail.com"
+        },
```
lyrixx, line 12: Is it legit?

nicolas-grekas: removed thanks

lyrixx (approved):

tests/Php81/Php81Test.php
```diff
@@ -0,0 +1,28 @@
...
+
+class Php81Test extends TestCase
+{
+    /**
+     * @covers \Symfony\Polyfill\Php81\Php81::array_is_list
+     */
+    public function testArrayIsList()
+    {
+        $this->assertTrue(array_is_list([]));
+        $this->assertTrue(array_is_list([\NAN]));
```
stof, line 24: this misses a case of an array with holes, and potentially also of a sequential array not starting from 0 (see the RFC describing this case for rejecting the `is_sequential_array` name).

Maybe we could take test cases from the PHP PR implementing the RFC.

composer.json
```diff
@@ -30,6 +30,7 @@
         "symfony/polyfill-php73": "self.version",
         "symfony/polyfill-php74": "self.version",
         "symfony/polyfill-php80": "self.version",
+        "symfony/polyfill-php81": "self.version",
```
stof, line 33: this is broken because you forgot to update the autoloading config of this root package to make it load the new code.

tests/Php81/Php81Test.php
nicolas-grekas: updated

composer.json
nicolas-grekas: apparently, we also missed loading the php80 bootstrap file here?
fixed!

nicolas-grekas:
@fabpot could you please configure the subtree splitter to split the `src/Php81` folder into https://github.com/symfony/polyfill-php81 ? (I already created the repo)

composer.json
stof: PHP 8 is loaded in `src/bootstrap.php` due to the different PHP requirement

derrabus (approved):

Issue #326, opened 2021-01-07
v1.22.0 is not compatible with PHP <7.4


gmessier:
This pull request: #322 use the ??= operator that is available only since PHP 7.4.
The code should be replaced using "$encoding  = $encoding ?? mb_internal_encoding();" or the minimal version in the composer.json must be bumped to ">=7.4"

derrabus:
Closing as explained in https://github.com/symfony/polyfill/pull/327#issuecomment-756400063

PR #327, opened 2021-01-07
Fix compatibility with PHP <7.4

gmessier:
To fix issue #326

### DIFF
--- a/src/Mbstring/bootstrap80.php
+++ b/src/Mbstring/bootstrap80.php
@@ -122,7 +122,7 @@ if (!function_exists('mb_chr')) {
     function mb_chr(int $codepoint, string $encoding = null): string|false { return p\Mbstring::mb_chr($codepoint, $encoding); }
 }
 if (!function_exists('mb_scrub')) {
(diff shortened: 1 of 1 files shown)


derrabus:
The file you are fixing in meant for PHP 8.0 and above. It also contains union types that won't work on PHP 7.

gmessier:
Oh. I didn't check things carefully.
In fact the problem is that we have this when we run Psalm:

"ERROR: ParseError - vendor/symfony/polyfill-mbstring/bootstrap80.php:125:86 - Syntax error, unexpected '=' on line 125 (see https://psalm.dev/173)
    function mb_scrub(string $string, string $encoding = null): string { $encoding ??= mb_internal_encoding(); return mb_convert_encoding($string, $encoding, $encoding); }"

derrabus:
Closing then.

gmessier:
But, we should still to find a way to prevent static analyzers to breaks because some files contains stuff that are not compatible with the PHP version used... manually add exclusions seems not the better way (not directly related to this package thought, but if you have ideas...).

derrabus:
I suggest you open an issue for Psalm because that's where the issue is. This package needs to ship files for PHP 8 because the polyfilled extension behaves differently on PHP 8.

PR #328, opened 2021-01-09
Use null-coalesce instead of null-coalesce-equals

dshafik:
The use of the PHP 8.0 only null-coalesce-equals (`??=`) means that this file causes Psalm to see it as a parse error for projects not yet migrated to PHP 8. There is currently no way to avoid this, as psalm considers a parse error an unignorable error, even though in this case the file would never be included unless it actually is running on PHP 8.0.

The only way to avoid this is to set the entire project as PHP 8.0 but then you wouldn't be notified of legitimate issues when deploying to PHP 7.x.

It's a simple change from `$a ??= $b` to `$a = $a ?? $b` form.

### DIFF
--- a/src/Mbstring/bootstrap80.php
+++ b/src/Mbstring/bootstrap80.php
@@ -122,7 +122,7 @@ if (!function_exists('mb_chr')) {
     function mb_chr(int $codepoint, string $encoding = null): string|false { return p\Mbstring::mb_chr($codepoint, $encoding); }
 }
 if (!function_exists('mb_scrub')) {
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
See https://github.com/vimeo/psalm/issues/4961

dshafik:
@nicolas-grekas I'm sorry but why not just make the code compatible? Nobody is using .php8 and this doesn't change the meaning, performance, or anything measurable about the code in any way — just to use a cool new operator that is less understood and barely documented.

nicolas-grekas:
This file also uses union types, which are not just a new cool operator. Changing the operator won't make the file more parseable on php 7.2.

PR #334, opened 2021-02-17
[PHP 8.1] Add `CURLStringFile` polyfill

### LINKED ISSUE #333: PHP 8.1 - CURLStringFile
PHP 8.1 adds a new `CURLStringFile`, that makes it possible to upload a file directly from a variable. With existing `CURLFile`, it is required to specify a path or a URI (that allows `data://` fortunately).

 - php/php-src#6456
 - [PHP 8.1: Curl: File uploads from strings with CURLStringFile](https://php.watch/versions/8.1/CURLStringFile)

There is a [polyfill](https://php.watch/versions/8.1/CURLStringFile#CURLStringFile-polyfill) that I created and tested manually. It works by encoding the passed data into a `data://` URI within the constructor, and setting them into properties of a `CURLFile` subclass.

I would like to ask if a `CURLStringFile` polyfill would fit Symfony Polyfill project. If it does, I'd happily send a PR :)

Thank you.

Ayesh:
PHP 8.1 adds a [new class `CURLStringFile`](https://php.watch/versions/8.1/CURLStringFile), that works similar to `CURLFile`, but does not require a _path_ to a file, and accepts a string of file contents instead. This can be polyfilled with a `CURLFile` sub class that uses a `data://` URI inside a constructure and calling `CURLFile::__construct` with that data URI.

Closes #333.

I will leave this as a draft because I'm not sure how to write tests for this. I'm thinking a PHP CLI server that echoes `$_FILES`, and a test that checks the returned content for file(s) that it uploaded.

### DIFF
--- a/src/Php81/README.md
+++ b/src/Php81/README.md
@@ -7,6 +7,7 @@ This component provides features added to PHP 8.1 core:
 - [`enum_exists`](https://php.net/enum-exists)
 - [`MYSQLI_REFRESH_REPLICA`](https://php.net/mysqli.constants#constantmysqli-refresh-replica) constant
 - [`ReturnTypeWillChange`](https://wiki.php.net/rfc/internal_method_return_types)
+- [`CURLStringFile`](https://php.net/CURLStringFile) (but only if PHP >= 7.4 is used)
 
 More information can be found in the
 [main Polyfill README](https://github.com/symfony/polyfill/blob/main/README.md).
--- a/src/Php81/Resources/stubs/CURLStringFile.php
+++ b/src/Php81/Resources/stubs/CURLStringFile.php
@@ -1,0 +1,51 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+if (\PHP_VERSION_ID >= 70400 && extension_loaded('curl')) {
+    /**
+     * @property string $data
+     */
+    class CURLStringFile extends CURLFile
+    {
+        private $data;
+
+        public function __construct(string $data, string $postname, string $mime = 'application/octet-stream')
+        {
+            $this->data = $data;
+            parent::__construct('data://application/octet-stream;base64,'.base64_encode($data), $mime, $postname);
+        }
+
+        public function __set(string $name, $value): void
+        {
+            if ('data' !== $name) {
+                $this->$name = $value;
+
+                return;
+            }
+
+            if (is_object($value) ? !method_exists($value, '__toString') : !is_scalar($value)) {
+                throw new \TypeError('Cannot assign '.gettype($value).' to property CURLStringFile::$data of type string');
+            }
+
+            $this->name = 'data://application/octet-stream;base64,'.base64_encode($value);
+        }
+
+        public function __isset(string $name): bool
+        {
+            return isset($this->$name);
+        }
+
+        public function &__get(string $name)
+        {
+            return $this->$name;
+        }
+    }
+}
--- a/src/Php81/composer.json
+++ b/src/Php81/composer.json
@@ -26,7 +26,7 @@
     "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
-            "dev-main": "1.27-dev"
+            "dev-main": "1.28-dev"
         },
         "thanks": {
             "name": "symfony/polyfill",
--- a/tests/Php81/CURLStringFileTest.php
+++ b/tests/Php81/CURLStringFileTest.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Tests\Php81;
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @requires extension curl
+ * @requires PHP 7.4
+ */
+class CURLStringFileTest extends TestCase
+{
+    private static $server;
+
+    public static function setUpBeforeClass(): void
+    {
+        $spec = [
+            1 => ['file', '/dev/null', 'w'],
+            2 => ['file', '/dev/null', 'w'],
+        ];
+        if (!self::$server = @proc_open(('\\' === \DIRECTORY_SEPARATOR ? '' : 'exec ').\PHP_BINARY.' -S localhost:8086', $spec, $pipes, __DIR__.'/fixtures')) {
+            self::markTestSkipped('Unable to start PHP server.');
+        }
+        sleep(1);
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        if (self::$server) {
+            proc_terminate(self::$server);
+            proc_close(self::$server);
+        }
+    }
+    
+    public function testCurlFileShowsContents(): void
+    {
+        $file = new \CURLStringFile('Hello', 'symfony.txt', 'text/plain');
+        $data = ['test_file' => $file];
+
+        $ch = curl_init('http://localhost:8086/curl-echo-server.php');
+
+        curl_setopt($ch, CURLOPT_POST,1);
+        curl_setopt($ch, CURLOPT_POSTFIELDS, $data);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
+        $response = curl_exec($ch);
+
+        $this->assertEquals(200, curl_getinfo($ch, CURLINFO_HTTP_CODE));
+        $this->assertSame('{"test_file":{"name":"symfony.txt","type":"text\/plain","error":0,"size":5}}', $response);
+    }
+}
--- a/tests/Php81/fixtures/curl-echo-server.php
+++ b/tests/Php81/fixtures/curl-echo-server.php
@@ -1,0 +1,10 @@
+<?php
+
+$output = $_FILES;
+foreach ($_FILES as $name => $file) {
+    if (\is_string($file['tmp_name'] ?? null)) {
+        unset($file['tmp_name'], $file['full_path']);
+        $output[$name] = $file;
+    }
+}
+echo json_encode($output);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.28.0
+
+  * Add `CURLStringFile` class introduced in PHP 8.1 (but only if PHP >= 7.4 is used)
+
 # 1.27.0
 
   * Add PHP 8.3 polyfill for `json_validate()`
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Polyfills are provided for:
 - the `enum_exists` function introduced in PHP 8.1;
 - the `MYSQLI_REFRESH_REPLICA` constant introduced in PHP 8.1;
 - the `ReturnTypeWillChange` attribute introduced in PHP 8.1;
+- the `CURLStringFile` class introduced in PHP 8.1 (but only if PHP >= 7.4 is used);
 - the `AllowDynamicProperties` attribute introduced in PHP 8.2;
 - the `SensitiveParameter` attribute introduced in PHP 8.2;
 - the `SensitiveParameterValue` class introduced in PHP 8.2;
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
     "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
-            "dev-main": "1.27-dev"
+            "dev-main": "1.28-dev"
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -64,6 +64,7 @@
             "src/Intl/MessageFormatter/Resources/stubs",
             "src/Intl/Normalizer/Resources/stubs",
             "src/Php82/Resources/stubs",
+            "src/Php81/Resources/stubs",
             "src/Php80/Resources/stubs",
             "src/Php73/Resources/stubs"
         ]
--- a/src/Apcu/composer.json
+++ b/src/Apcu/composer.json
@@ -25,7 +25,7 @@
     "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
-            "dev-main": "1.27-dev"
+            "dev-main": "1.28-dev"
         },
         "thanks": {
             "name": "symfony/polyfill",
--- a/src/Ctype/composer.json
+++ b/src/Ctype/composer.json
@@ -31,7 +31,7 @@
     "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
-            "dev-main": "1.27-dev"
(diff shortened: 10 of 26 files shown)


src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+// https://php.watch/versions/8.1/CURLStringFile#CURLStringFile-polyfill
+class CURLStringFile extends CURLFile
+{
+    public function __construct(string $data, string $postname, string $mime = "application/octet-stream")
```
OskarStark, line 6:
```suggestion
    public function __construct(string $data, string $postname, string $mime = 'application/octet-stream')
```

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+// https://php.watch/versions/8.1/CURLStringFile#CURLStringFile-polyfill
+class CURLStringFile extends CURLFile
```
nicolas-grekas, line 4: I think we should wrap this declaration in an `extension_loaded('curl')` check

nicolas-grekas:
If we add a test case (as you, I don't know which one exactly), it should with the native one and with the polyfilled one (when curl is not installed), so that we can prove that this is really a polyfill.

derrabus:
> I'm thinking a PHP CLI server that echoes `$_FILES`, and a test that checks the returned content for file(s) that it uploaded.

Sounds like a plan. You can find an example of using the CLI webserver in tests here:

https://github.com/symfony/symfony/blob/6ae59a9a171695ddef5294898681c060e8c1f3fe/src/Symfony/Component/HttpFoundation/Tests/ResponseFunctionalTest.php#L22-L29

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,10 @@
...
+{
```
stof, line 5: This polyfill is incomplete, because it does not expose the properties. CURLStringFile is a mutable class in PHP.

stof (changes requested):

src/Php81/Resources/stubs/CURLStringFile.php
nicolas-grekas: alternatively, we'd need to polyfill `CURLFile` also

stof: well, there is no reason to define CURLStringFile if curl is not available IMO.

and btw, we cannot polyfill CURLFile if you have an older curl version without CURLFile: the core `curl` functions would not know how to handle it (here, extending `CURLFile` is precisely the hack to make `curl_*` functions do the right thing with our `CURLStringFile` when receiving it as argument).

The proper guard is probably even `class_exists(CURLFile::class, false)`, because the `symfony/polyfill-php81` package supports PHP 7.1+ but curl does not have `CURLFile` in these older versions.

nicolas-grekas: I agree. I'd add both conditions to benefit from opcache optimizations:
`if (extension_loaded('curl') && class_exists(CURLFile::class, false)) {`

src/Php81/Resources/stubs/CURLStringFile.php
nicolas-grekas: You mean that mutating the `data` property should be taken into account, isn't it? Anything else?

stof: Well, we need to expose the 3 public properties, and support mutating them, but also reading them.

nicolas-grekas: Since we extend `CURLFile`, which already has `$mime` and `$postname`, they're covered, right?
`$name` would need to be kept in sync with `$data` though.
I suppose this can be done with some magic methods

stof: the mime type also impacts the data when using a base64 URL. Also, are they public properties in CURLFile ?

nicolas-grekas: > the mime type also impacts the data when using a base64 URL.

Oh, then maybe we can use a generic mime type or no mime type in the data URL, if curl supports that @Ayesh?

> Also, are they public properties in CURLFile ?

Yes they are: https://www.php.net/CURLFile

stof: I fear that passing a generic mime type might change how curl sends the data (as CURLFile would then have 2 conflicting values for the mime type)

Ayesh:
Thanks a lot for the feedback. 

I totally forgot about the `CURLStringFile::$data` mutability, and like @nicolas-grekas said, we might (unfortunately) need to use magic methods here. I updated with `__get`, `__set` and `__isset` methods to update the `CURLStringFile::$name` if `$data` was modified. It feels a bit... yucky... to do this, but I'm looking for any suggestions.

src/Php81/Resources/stubs/CURLStringFile.php
Ayesh: Thank you - I added the `if` block. I don't think it will prevent Composer autoloader from _trying_ to load it, but I think the condition is necessary from the polyfill to not polyfill if `curl` extension isn't enabled.

src/Php81/Resources/stubs/CURLStringFile.php
Ayesh: Thank you, I changed it like you suggested.

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+if (extension_loaded('curl') && class_exists(CURLFile::class, false)) {
+    // https://php.watch/versions/8.1/CURLStringFile#CURLStringFile-polyfill
```
nicolas-grekas, line 4: should be removed
and replaced by an `@property mixed $data` annotation

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,43 @@
...
+    class CURLStringFile extends CURLFile
+    {
+        private $rawData;
```
nicolas-grekas, line 7: should be named `$data`

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public function __isset(string $name): bool
+        {
+            return $name === 'data';
+        }
+
+        public function __get(string $name)
+        {
+            if ($name === 'data') {
```
nicolas-grekas, line 36: `return $this->$name;` would make it

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,43 @@
...
+                throw new \TypeError('Cannot assign null to property CURLStringFile::$data of type string');
+            }
+
+            $this->name = 'data://' . $this->mime . ';base64,' . base64_encode($value);
+            $this->rawData = $value;
+        }
+
+        public function __isset(string $name): bool
+        {
+            return $name === 'data';
```
nicolas-grekas, line 31: `return isset($this->$name);`

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,43 @@
...
+if (extension_loaded('curl') && class_exists(CURLFile::class, false)) {
+    // https://php.watch/versions/8.1/CURLStringFile#CURLStringFile-polyfill
+    class CURLStringFile extends CURLFile
+    {
+        private $rawData;
+
+        public function __construct(string $data, string $postname, string $mime = 'application/octet-stream')
+        {
+            $this->rawData = $data;
+            parent::__construct('data://' . $mime . ';base64,' . base64_encode($data), $mime, $postname);
```
nicolas-grekas, line 12: can you try putting always the same mime-type here (`application/octet-stream`)? Does Curl complain if this is out of sync with `$mime`?
because if it does, we'll also have to sync $nale when $mime changes - I hope we won't have to.

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public function __set(string $name, $value): void
+        {
+            if ($name !== 'data') {
```
nicolas-grekas, line 17:
```php
if ('data' === $name) {
    $this->name = 'data://application/octet-stream;base64,' . base64_encode($value);
}

$this->$name = $value;
```

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,43 @@
...
+                return;
+            }
+
+            if (!is_string($value)) {
```
nicolas-grekas, line 21: I suppose native PHP also accepts stringable, or integers, etc?
Maybe we should just not do this type check.
Alternatively, the closes we could do is adding the `string` type to the `$value` argument, in the signature.

nicolas-grekas (commented):
can you please also add a line in the changelog?

src/Php81/composer.json
```diff
@@ -20,7 +20,8 @@
     },
     "autoload": {
         "psr-4": { "Symfony\\Polyfill\\Php81\\": "" },
-        "files": [ "bootstrap.php" ]
+        "files": [ "bootstrap.php" ],
+        "classmap": [ "Resources/stubs" ]
```
nicolas-grekas, line 24: should be listed in the main composer.json also

Ayesh:
Thank you so much for the help @nicolas-grekas - I have made the changes. 
I will also push a test. It passes on PHP >= 7.4, and I'm still pulling my hair on why it fails on older PHP versions.

stof:
> I will also push a test. It passes on PHP >= 7.4, and I'm still pulling my hair on why it fails on older PHP versions.

that's because your implementation relies on the fact that curl knows how to deal with a `CurlFile` object. But that `CurlFile` won't exist there.

Ayesh:
Thank you @stof. Looking at [CURLFile doc](https://www.php.net/manual/en/class.curlfile.php), it says CURLFile should available in PHP >= 5.5. My direction is that the CLI server failed to start at all in older PHP versions.

tests/Php81/CURLStringFileTest.php
```diff
@@ -0,0 +1,56 @@
...
+    public function testCurlFileShowsContents(): void {
+        $file = new \CURLStringFile('Hello', 'symfony.txt', 'text/plain');
+        $data = ['test_file' => $file];
+
+        $ch = curl_init('http://localhost:8086/curl-echo-server.php');
+
+        curl_setopt($ch, CURLOPT_POST,1);
+        curl_setopt($ch, CURLOPT_POSTFIELDS, $data);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
+        $reponse = curl_exec($ch);
```
stof, line 52: typo in the variable name

tests/Php81/fixtures/curl-echo-server.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+$output = $_FILES;
+foreach ($_FILES as $name => $file) {
+    unset($file['tmp_name']);
+    $output[$name] = $file;
+}
+print_r($output);
```
stof, line 8: I'd rather use `json_encode` than `print_r`. I'm not sure `print_r` is guaranteed to keep the exact same output formatting forever.

tests/Php81/CURLStringFileTest.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $this->assertStringEqualsFile(__DIR__ . '/fixtures/CURLSTringFileTest-expect.txt', $reponse);
```
stof, line 54: before asserting the content of the response, you should first add a check that the HTTP request was successful IMO

tests/Php81/fixtures/curl-echo-server.php
Ayesh: Thank you; You are right, `json_encode` is more resilient. Updating now.

tests/Php81/CURLStringFileTest.php
Ayesh: Thank you - Adding an assertion for response code === 200.

Ayesh:
I tested with a PHP 7.3 setup, it indeed turns out that the problem is that CURLFile only supports streams in PHP >= 7.4, just like @stof mentioned. 
I updated the tests and `CURLStringFile` to be only available in PHP >= 7.4, so I can mark this PR for review. All tests pass. 

For PHP 7.0 through 7.3, I think we _could_ get it working by temporarily writing the contents to a file, but I'm not sure if the effort is worth it. I will open the PR for review either way, and will happily work on support for older PHP versions if you think it's a worthwhile idea. 🤞🏼

Thank you.

tests/Php81/fixtures/curl-echo-server.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+$output = $_FILES;
+foreach ($_FILES as $name => $file) {
+    unset($file['tmp_name']);
```
nicolas-grekas, line 5: Just in case, we should also check that the array has the proper structure.
Right now, if one submits `foo[]` as a file name, this file will leak the `tmp_name`.
Better not.

nicolas-grekas:
Can you please rebase?

tests/Php81/fixtures/curl-echo-server.php
Ayesh: I will work on this. I think we need to recursively iterate and remove the `tmp_name`. Probably worth adding an extra test too, for nested files.

src/Php81/Resources/stubs/CURLStringFile.php
nicolas-grekas: what about this?

nicolas-grekas (approved):
(I fixed my last comments)

src/Php81/Resources/stubs/CURLStringFile.php
```diff
@@ -0,0 +1,36 @@
...
+        private $data;
+
+        public function __construct(string $data, string $postname, string $mime = 'application/octet-stream')
+        {
+            $this->data = $data;
+            parent::__construct('data://application/octet-stream;base64,' . base64_encode($data), $mime, $postname);
+        }
+
+        public function __set(string $name, $value): void
+        {
```
stof, line 18: CURLStringFile has typed properties in PHP. So you probably need to validate the type of `$value` and trigger a TypeError when it is not a string

src/Php81/README.md
```diff
@@ -5,6 +5,7 @@ This component provides features added to PHP 8.1 core:
 
 - [`array_is_list`](https://php.net/array_is_list)
 - [`MYSQLI_REFRESH_REPLICA`](https://www.php.net/manual/en/mysqli.constants.php#constantmysqli-refresh-replica) constant
+- `CURLStringFile` class
```
stof, line 8: I would mention that this is only polyfilled for PHP 7.4+, not for older versions.

nicolas-grekas (approved):
I just rebased and resolved @stof's comments.

nicolas-grekas:
Thank you @Ayesh.

PR #336, opened 2021-02-18
PHP 7.3 fix

glushchenko:
For PHP 7.3 default variant is INTL_IDNA_VARIANT_2003:

```
if (\PHP_VERSION_ID < 70400) {
        function idn_to_ascii($domain, $flags = 0, $variant = \INTL_IDNA_VARIANT_2003, &$idna_info = null) { return p\Idn::idn_to_ascii($domain, $flags, $variant, $idna_info); }
    }
        function idn_to_utf8($domain, $flags = 0, $variant = \INTL_IDNA_VARIANT_2003, &$idna_info = null) { return p\Idn::idn_to_utf8($domain, $flags, $variant, $idna_info); }
    }
}
```

But in Idn.php script executes error for INTL_IDNA_VARIANT_2003.

```
        if (\PHP_VERSION_ID >= 70200 && self::INTL_IDNA_VARIANT_2003 === $variant) {
            @trigger_error('idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated', \E_USER_DEPRECATED);
        }
```

In result all default functions like idn_to_ascii('demo.com') and idn_to_utf8('demo.com') are not working.

### DIFF
--- a/src/Intl/Idn/Idn.php
+++ b/src/Intl/Idn/Idn.php
@@ -142,12 +142,13 @@ final class Idn
      * @param int    $options
      * @param int    $variant
      * @param array  $idna_info
+     * @param int    $phpVersion
      *
      * @return string|false
      */
-    public static function idn_to_ascii($domainName, $options = self::IDNA_DEFAULT, $variant = self::INTL_IDNA_VARIANT_UTS46, &$idna_info = [])
+    public static function idn_to_ascii($domainName, $options = self::IDNA_DEFAULT, $variant = self::INTL_IDNA_VARIANT_UTS46, &$idna_info = [], $phpVersion = \PHP_VERSION_ID)
     {
-        if (\PHP_VERSION_ID >= 70200 && self::INTL_IDNA_VARIANT_2003 === $variant) {
+        if ($phpVersion >= 70400 && self::INTL_IDNA_VARIANT_2003 === $variant) {
             @trigger_error('idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated', \E_USER_DEPRECATED);
         }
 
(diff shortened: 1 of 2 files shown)


src/Intl/Idn/Idn.php
```diff
@@ -142,12 +142,13 @@ final class Idn
...
      * @param array  $idna_info
+     * @param int    $phpVersion
      *
      * @return string|false
      */
-    public static function idn_to_ascii($domainName, $options = self::IDNA_DEFAULT, $variant = self::INTL_IDNA_VARIANT_UTS46, &$idna_info = [])
+    public static function idn_to_ascii($domainName, $options = self::IDNA_DEFAULT, $variant = self::INTL_IDNA_VARIANT_UTS46, &$idna_info = [], $phpVersion = \PHP_VERSION_ID)
     {
-        if (\PHP_VERSION_ID >= 70200 && self::INTL_IDNA_VARIANT_2003 === $variant) {
+        if ($phpVersion >= 70400 && self::INTL_IDNA_VARIANT_2003 === $variant) {
```
stof, line 151: `\PHP_VERSION_ID` should stay inline so that OPcache can perform dead code elimination.

stof (changes requested):

tests/Intl/Idn/IdnTest.php
```diff
@@ -288,6 +288,15 @@ public function testEncodePhp53($decoded, $encoded)
         $this->assertSame($encoded, $result);
     }
 
+    /**
+     * @dataProvider domainNamesProvider
+     */
+    public function testEncodeDefaultPhp73($decoded, $encoded)
```
stof, line 294: this test does not make sense. It is not relying on a default encoding, and is overriding a PHP version check which does not make sense either.

src/Intl/Idn/Idn.php
glushchenko: I can make changes, then code will be non testable.

tests/Intl/Idn/IdnTest.php
glushchenko: Problem related to php version checking, encode/decode not working with PHP 7.3 only. Of course i can rewrite it for only for check one item.

stof: the right way to check that the polyfill works on PHP 7.3 is to run the testsuite on PHP 7.3, not to fake a version check (which does not guarantee anything)

nicolas-grekas:
Can you please also remove the `INTL_IDNA_VARIANT_2003` constant from https://github.com/symfony/polyfill/blob/main/src/Intl/Idn/bootstrap80.php#L78

src/Intl/Idn/Idn.php
```diff
@@ -147,7 +147,7 @@ final class Idn
      */
     public static function idn_to_ascii($domainName, $options = self::IDNA_DEFAULT, $variant = self::INTL_IDNA_VARIANT_UTS46, &$idna_info = [])
     {
-        if (\PHP_VERSION_ID >= 70200 && self::INTL_IDNA_VARIANT_2003 === $variant) {
+        if (\PHP_VERSION_ID >= 70400 && self::INTL_IDNA_VARIANT_2003 === $variant) {
```
nicolas-grekas, line 150: But PHP itself triggers the deprecation on PHP 7.3: https://3v4l.org/ghvgU
So I'm not sure this is correct.

glushchenko:
Seems like invalid errors handling in my project. Thank you guys.

PR #338, opened 2021-02-19
[PHP 8.1] add enum_exists()

### LINKED ISSUE #335: PHP 8.1 - enum_exists
In order to simplify code checking for class existence for any kind of PHP classes (actual classes, interfaces, traits and now enums), it might be great to provide a polyfill for the `enum_exists` function.
The implementation would be quite straightforward: `return false;` as enums cannot exist on older versions.

nicolas-grekas:
Fix #335

### DIFF
--- a/src/Php81/bootstrap.php
+++ b/src/Php81/bootstrap.php
@@ -18,3 +18,7 @@ if (\PHP_VERSION_ID >= 80100) {
 if (!function_exists('array_is_list')) {
     function array_is_list(array $array): bool { return p\Php81::array_is_list($array); }
 }
+
+if (!function_exists('enum_exists')) {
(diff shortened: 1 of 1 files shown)


src/Php81/bootstrap.php
```diff
@@ -18,3 +18,7 @@
 if (!function_exists('array_is_list')) {
     function array_is_list(array $array): bool { return p\Php81::array_is_list($array); }
 }
+
+if (!function_exists('enum_exists')) {
+    function enum_exists(string $enum, bool $autoload = true): bool { return $autoload && class_exists($enum, $autoload) && false; }
```
stof, line 23: you could omit the `$autoload` argument of `class_exist` as `false` will never reach that point (or remove the `$autoload && ` before it)

nicolas-grekas: true, updated.

stof (approved):

PR #343, opened 2021-03-17
[PHP 8] Add IntlDateFormatter::RELATIVE_* constants

sebpacz:
This PR adds polyfill for new constants introduced for dateType in the `IntlDateFormatter` class (see https://www.php.net/manual/en/class.intldateformatter.php#intl.intldateformatter-constants). The new constants work the same as `IntlDateFormatter::FULL`, `IntlDateFormatter::LONG`, `IntlDateFormatter::MEDIUM`, `IntlDateFormatter::SHORT`, but yesterday, today, and tomorrow is displayed as yesterday, today, and tomorrow, respectively. Moreover, I discovered that the pattern is ignored if the date, that will be formatted is the today, yesterday, or tomorrow date.


src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -571,4 +613,32 @@ protected function getDefaultPattern()
...
     }
+
+    /**
+     * Returns a relative date format for the specified timestamp.
+     *
+     * @param int $timestamp
+     *
+     * @return string
+     */
+    protected function getRelativeDateFormat(int $timestamp)
```
stof, line 624: should be private

src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -67,10 +67,23 @@ abstract class IntlDateFormatter
...
+    public const RELATIVE_FULL = 128;
+    public const RELATIVE_LONG = 129;
+    public const RELATIVE_MEDIUM = 130;
+    public const RELATIVE_SHORT = 131;
+
     /* calendar formats */
     public const TRADITIONAL = 0;
     public const GREGORIAN = 1;
 
+    protected const RELATIVE_DATE_TYPES = [
```
nicolas-grekas, line 80: should be removed to me

src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -146,6 +168,10 @@ public function __construct(?string $locale, ?int $dateType, ?int $timeType, $ti
 
         $this->setPattern($pattern);
         $this->setTimeZone($timezone);
+
+        if (\in_array($this->dateType, self::RELATIVE_DATE_TYPES)) {
```
nicolas-grekas, line 172: `$this->isRelativeDateType = \in_array($this->dateType, [...]);`

src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -207,8 +233,28 @@ public function format($datetime)
...
 
-        $transformer = new FullTransformer($this->getPattern(), $this->getTimeZoneId());
-        $formatted = $transformer->format($this->createDateTime($datetime));
+        $pattern = $this->getPattern();
+        $formatted = '';
+
+        if ($this->isRelativeDateType) {
+            $formatted = $this->getRelativeDateFormat($datetime);
+
+            if (!empty($formatted)) {
```
nicolas-grekas, line 242: we should avoid using `empty()`

src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -571,4 +617,32 @@ protected function getDefaultPattern()
 
         return $pattern;
     }
+
+    /**
+     * Returns a relative date format for the specified timestamp.
+     *
+     * @param int $timestamp
```
nicolas-grekas, line 624: should be removed

src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -571,4 +617,32 @@ protected function getDefaultPattern()
...
+     *
+     * @return string
```
nicolas-grekas, line 626: should be removed and replaced by a real return type

src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -571,4 +617,32 @@ protected function getDefaultPattern()
...
+     */
+    private function getRelativeDateFormat(int $timestamp)
+    {
+        $today = \DateTime::createFromFormat('U', time(), $this->dateTimeZone);
```
nicolas-grekas, line 630: this should use `createDateTime()` instead

src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -571,4 +617,32 @@ protected function getDefaultPattern()
...
+
+        $datetime = $this->createDateTime($timestamp);
+        $datetime->setTime(0, 0, 0);
+
+        $interval = $today->diff($datetime);
+
+        if (false !== $interval) {
+            if (0 === $interval->days) {
+                return 'today';
+            } elseif (1 === $interval->days) {
```
nicolas-grekas, line 641: "elseif" can be replaced by an "if", thanks to the "return"

tests/Intl/Icu/AbstractIntlDateFormatterTest.php
```diff
@@ -479,6 +479,16 @@ public function testFormatWithTimezoneFromPhp()
...
     }
 
+    public function testFormatIgnoresPatternForRelativeDateType()
+    {
+        $formatter = $this->getDateFormatter('en', IntlDateFormatter::RELATIVE_FULL, IntlDateFormatter::FULL, new \DateTimeZone('GMT'), IntlDateFormatter::GREGORIAN, 'zzzz');
+
+        $datetime = \DateTime::createFromFormat('U', time(), new \DateTimeZone('GMT'));
+        $datetime->setTime(0, 0, 0);
+
+        $this->assertEquals('today at 12:00:00 AM Greenwich Mean Time', $formatter->format($datetime));
```
nicolas-grekas, line 489: let's use assertSame when possible

tests/Intl/Icu/Verification/IntlDateFormatterTest.php
```diff
@@ -71,6 +71,35 @@ public function testDateAndTimeType($timestamp, $datetype, $timetype, $expected)
         parent::testDateAndTimeType($timestamp, $datetype, $timetype, $expected);
     }
 
+    /**
+     * @dataProvider relativeDateTypeProvider
+     */
+    public function testRelativeDateType($timestamp, $datetype, $timetype, $expected)
+    {
+        if (\PHP_VERSION_ID < 80000) {
```
nicolas-grekas, line 79: should be replaced by `@requires PHP 8`

tests/Intl/Icu/Verification/IntlDateFormatterTest.php
```diff
@@ -71,6 +71,35 @@ public function testDateAndTimeType($timestamp, $datetype, $timetype, $expected)
...
+        if (version_compare(\INTL_ICU_VERSION, '59.1', '<')) {
+            $this->markTestSkipped('ICU version 59.1 is required.');
+        }
+
+        parent::testRelativeDateType($timestamp, $datetype, $timetype, $expected);
+    }
+
+    public function testFormatIgnoresPatternForRelativeDateType()
+    {
+        if (\PHP_VERSION_ID < 80000) {
```
nicolas-grekas, line 92: `@requires PHP 8`

nicolas-grekas (commented):
Thanks for submitting, here are some comments.

sebpacz:
@nicolas-grekas thank you for your review. I have updated the code according to your suggestions.

stof:
Be aware that this won't allow you to use those constants on PHP <8 if the native implementation of intl formatters is loaded though (as we cannot patch existing classes).
So this polyfilling is useful for code requiring PHP 8+ but wanting to make the dependency on ext-intl optional. But projects supporting PHP 7 will still need to check for existence of the feature.

src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -207,8 +226,26 @@ public function format($datetime)
             $datetime = $datetime->getTimestamp();
         }
 
-        $transformer = new FullTransformer($this->getPattern(), $this->getTimeZoneId());
-        $formatted = $transformer->format($this->createDateTime($datetime));
+        $pattern = $this->getPattern();
+        $formatted = '';
+
+        if ($this->isRelativeDateType && $formatted = $this->getRelativeDateFormat($datetime)) {
+            $formatted = $this->getRelativeDateFormat($datetime);
```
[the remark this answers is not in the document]
sebpacz: you're right, thank you

derrabus (approved):

PR #345, opened 2021-03-22
[PHP 8.1] Fiber exceptions

Ayesh:
The Fiber [RFC](https://wiki.php.net/rfc/fibers) is concluding today with 48:14 in favor. So it looks like it will pass and we will get Fibers in PHP 8.1 🎉🍾.  I'm getting ahead of myself, but this RFC is just as good as passed now :)

While we cannot bring the main `Fiber` class, I think bringing [`FiberError`](https://php.watch/versions/8.1/fibers#FiberError) (extends `\Error`) and [`FiberExit`](https://php.watch/versions/8.1/fibers#FiberExit) (extends `\Exception`) can help bring cross-version compatibility for error logs (that serialize exceptions), and elsewhere Exceptions are dealt with. 

We previously added Exception classes (such as `ValueError` and `match` expressions exceptions) to PHP 8.0 polyfill on similar use cases.

### DIFF
--- a/src/Php81/Resources/stubs/FiberExit.php
+++ b/src/Php81/Resources/stubs/FiberExit.php
@@ -1,0 +1,15 @@
+<?php
+
+/**
+ * Exception thrown when destroying a fiber. This exception cannot be caught by user code.
+ */
+final class FiberExit extends Exception
+{
+    /**
+     * Constructor throws to prevent user code from throwing FiberExit.
+     */
+    public function __construct()
+    {
+        throw new \Error('The "FiberExit" class is reserved for internal use and cannot be manually instantiated');
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -62,6 +62,7 @@
             "src/Intl/MessageFormatter/Resources/stubs",
             "src/Intl/Normalizer/Resources/stubs",
             "src/Php80/Resources/stubs",
+            "src/Php81/Resources/stubs",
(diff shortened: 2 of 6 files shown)


nicolas-grekas:
Are you sure we can't polyfill the Fiber class, in a synchronous style? I didn't think about it, but that'd be great if yes
/cc @kelunik @trowski do you think it's possible?

kelunik:
@nicolas-grekas if it was possible, we wouldn't have needed the RFC 😄

Ayesh:
I was amazed at the cleverness of Amphp with `Generator`, but i don't think they come close to Fiber's suspend/resume features. Awesome stuff from the RFC!

We could technically accept a callable and prohibit it from running `suspend` method, but I think that will be a lot less useful even if not troublesome.

src/Php81/Resources/stubs/FiberExit.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+/**
+ * Exception thrown when destroying a fiber. This exception cannot be caught by user code.
+ */
+final class FiberExit extends Exception
```
nikic, line 6: This class doesn't actually exist. It's an internal marker class that is not (or should not be) visible from user code.

nicolas-grekas:
Closing as I don't see how that could be useful in practice.
Also after @nikic's comment.
Thanks for submitting.

PR #347, opened 2021-04-05
Grapheme, include 8.0.0 being php 8.0 syntax

glensc:


glensc:
actually, I'm not sure what the code supposed to do, but comparing `PHP_VERSION_ID > 80000` seems wrong.

now on closer look, it should be rather `PHP_VERSION_ID < 80000` then return, i.e do not throw about type errors on PHP 8.0?

glensc:
the changes were introduced in https://github.com/symfony/polyfill/commit/436c20f0ba9481ed7e38a887a877b750d294735b

they are all wrong? i.e needs flipping?

src/Intl/Grapheme/Grapheme.php
```diff
@@ -62,7 +62,7 @@ public static function grapheme_extract($s, $size, $type = \GRAPHEME_EXTR_COUNT,
         $start = (int) $start;
 
         if (\GRAPHEME_EXTR_COUNT !== $type && \GRAPHEME_EXTR_MAXBYTES !== $type && \GRAPHEME_EXTR_MAXCHARS !== $type) {
-            if (80000 > \PHP_VERSION_ID) {
+            if (80000 >= \PHP_VERSION_ID) {
```
derrabus, line 65: I'm pretty sure the current version is correct. `80000 > \PHP_VERSION_ID` matches PHP releases before 8.0. For those releases, we return `false` because that's how `ext-mbstring` would behave there.

The `ValueError` that we throw otherwise has been introduced with PHP 8.0.0.

derrabus (changes requested):

src/Intl/Grapheme/Grapheme.php
glensc: After flipping the Yoda mode, you can see that `PHP_VERSION_ID > 80000` maches versions afer 8.0.0, no including 8.0.0 iself!

glensc: Nevermind, when flipping Yoda mode, it should have to change the sign too, so it would be `PHP_VERSION_ID < 80000`

Blame Yoda for brain freeze here!

and here it's not even justified (to have Yoda mode), you can't assign values to constants with assign statement! but makes code difficult to understand if you don't do Yoda.

glensc: making a comparison "is number x less than than 8" with `8 > x` makes my brain hurt!

derrabus: I feel you. But that's the CS that is enforced on this repo. 🤷🏻

Issue #342, opened 2021-03-17
Mbstring polyfill calls undefined iconv


kontur:
Hey,

I'm authoring a Wordpress plugin and use Twig in it. I've recently updated my Twig library I use in it and this recently installed this polyfill package (polyfill-mbstring v1.22.1). I got the following report / error log from a user using PHP 7.3.27:

```
An error of type E_ERROR was caused in line 285 of the file /home/---/public_html/wp-content/plugins/fontsampler/vendor/symfony/polyfill-mbstring/Mbstring.php. Error message: Uncaught Error: Call to undefined function Symfony\Polyfill\Mbstring\iconv() in /home/---/public_html/wp-content/plugins/fontsampler/vendor/symfony/polyfill-mbstring/Mbstring.php:285
Stack trace:
#0 /home/---/public_html/wp-content/plugins/fontsampler/vendor/symfony/polyfill-mbstring/Mbstring.php(585): Symfony\Polyfill\Mbstring\Mbstring::mb_convert_case('lerySlideshowTh...', 1, 'ASCII')
#1 /home/---/public_html/wp-content/plugins/fontsampler/vendor/symfony/polyfill-mbstring/bootstrap.php(67): Symfony\Polyfill\Mbstring\Mbstring::mb_strtolower('lerySlideshowTh...', 'ASCII')
#2 /home/---/public_html/wp-content/plugins/redux-framework/redux-core/inc/classes/class-redux-autoloader.php(59): mb_strtolower('lerySlideshowTh...', 'ASCII')
#3 [internal function]: Redux_Autoloader->load('lerySlideshowTh...')
#4 [internal function]: spl_autoload_call('JSJGallerySlide...')
#5 /home/---/public_html/wp-content/plugins/jsj-gallery-slideshow/themes/default/theme-main.php(12): class_exists('JSJGallerySlide..
```

Looking at L285 the code seems like a simple call to `iconv` which should be available in that PHP version. Yet the error log above reads like it is trying to call iconv as method of Mbstring — or am I misinterpreting this?

Note that a error of this gravity will actually shut down that users entire Wordpress site, caused by my plugin — so quite a nasty surprise just from having updated Twig.

Any idea what's going on here?

nicolas-grekas:
I suppose ext-iconv is not installed on this setup. Try installing the iconv polyfill next to the mbstring one.

Alternatively, try installing ext-iconv.

kontur:
[...]

PR #349, opened 2021-04-19
Use fully-qualified iconv calls in the mbstring polyfill

stof:
The mbstring polyfill does not require the iconv polyfill due to the size of the package and the fact that most systems will have the native ext-iconv (it is enabled by default in PHP).
Using fully qualified calls gives a better error message when the iconv extension is missing and the polyfill is not installed, which makes it easier to solve the issue.

Refs #342


nicolas-grekas (approved):

derrabus (approved):

nicolas-grekas:
Thank you @stof.

PR #350, opened 2021-04-19
Update intl currencies data

nicolas-grekas:


tests/Intl/Icu/CurrenciesTest.php
```diff
@@ -18,10 +18,20 @@
  */
 class CurrenciesTest extends TestCase
 {
+    /**
+     * @requires PHP 7.2
```
stof, line 22: why this requirement ?

nicolas-grekas: because on 7.1 symfony/intl 4.4 is installed, which doesn't have the updated currencies

jderusse: Side thinking: Should we add a constant with the version in our packages?

I know that we already talked about that and we prefer using behavioal check (ie. `if class_exists...`)

but in some cases this does not work (ie. this case, or change hard to detect like method's argument change/deprecated). And we have to add a check on something that is totally unrelated (like this php version check)

nicolas-grekas: Unless we *have to*, we should try hard not to add this constant IMHO.

stof: @jderusse no constant. Use `Composer\InstalledVersions('symfony/intl')` if you need the info (or one of the dedicated composer plugins if you need support for composer 1)

stof: @nicolas-grekas a check on `Intl::getIcuDataVersion()` would make more sense IMO (and be more reliable as there is nothing forbidding to install the 4.4 package on 7.2+, for instance in a `--prefer-lowest` job)

Issue #351, opened 2021-04-23
Missing return in mb_parse_str


webber2k6:
The function of mb_parse_str in bootstrap80.php requires the return value to be of type bool, but nothing is returned.

This gives a PHP TypeError:
TypeError: mb_parse_str(): Return value must be of type bool, none returned in ~\vendor\symfony\polyfill-mbstring\bootstrap80.php:54

PR #352, opened 2021-04-23
[mbstring] add return value to mb_parse_str (#351)

webber2k6:
This pull request fixes the TypeError from #351.

### DIFF
--- a/src/Mbstring/bootstrap.php
+++ b/src/Mbstring/bootstrap.php
@@ -55,7 +55,7 @@ if (!function_exists('mb_detect_order')) {
     function mb_detect_order($encoding = null) { return p\Mbstring::mb_detect_order($encoding); }
 }
 if (!function_exists('mb_parse_str')) {
-    function mb_parse_str($string, &$result = []) { parse_str($string, $result); }
+    function mb_parse_str($string, &$result = []) { parse_str($string, $result); return (bool) $result; }
 }
 if (!function_exists('mb_strlen')) {
     function mb_strlen($string, $encoding = null) { return p\Mbstring::mb_strlen($string, $encoding); }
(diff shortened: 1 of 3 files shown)


src/Mbstring/bootstrap80.php
```diff
@@ -51,7 +51,7 @@ function mb_detect_encoding(?string $string, array|string|null $encodings = null
     function mb_detect_order(array|string|null $encoding = null): array|bool { return p\Mbstring::mb_detect_order((string) $encoding); }
 }
 if (!function_exists('mb_parse_str')) {
-    function mb_parse_str(?string $string, &$result = []): bool { parse_str((string) $string, $result); }
+    function mb_parse_str(?string $string, &$result = []): bool { parse_str((string) $string, $result); return !empty($result); }
```
nicolas-grekas, line 54: `return (bool) $result;` would be enough

nicolas-grekas (commented):
Thanks for the PR.
Can you please add a test case in `tests/Mbstring/MbstringTest.php`?

nicolas-grekas:
`src/Mbstring/bootstrap.php` has the same issue btw (the test case would spot it ;) )

webber2k6:
> `src/Mbstring/bootstrap.php` has the same issue btw (the test case would spot it ;) )

The function in `src/Mbstring/bootstrap.php` has no return type annotation.

nicolas-grekas:
> The function in src/Mbstring/bootstrap.php has no return type annotation.

sure, but the native function returns a bool nonetheless, so we have to align the behavior of the polyfil.

src/Mbstring/bootstrap.php
```diff
@@ -55,7 +55,7 @@ function mb_detect_encoding($string, $encodings = null, $strict = false) { retur
     function mb_detect_order($encoding = null) { return p\Mbstring::mb_detect_order($encoding); }
 }
 if (!function_exists('mb_parse_str')) {
-    function mb_parse_str($string, &$result = []) { parse_str($string, $result); }
+    function mb_parse_str($string, &$result = []): bool { parse_str($string, $result); return (bool) $result; }
```
nicolas-grekas, line 58: the return type should be removed: the native function doesn't have one until PHP8, see https://3v4l.org/m49P0

nicolas-grekas (approved):

nicolas-grekas:
Thank you @webber2k6.

PR #354, opened 2021-05-20
Add ReturnTypeWillChange attribute

derrabus:
I'd like to add this nice little class that I've found in my PHP 8.1 nightly build.

```php
dump(new ReflectionClass(\ReturnTypeWillChange::class));
```

Bildschirmfoto 2021-05-21 um 15 23 19

### DIFF
--- a/src/Php81/Resources/stubs/ReturnTypeWillChange.php
+++ b/src/Php81/Resources/stubs/ReturnTypeWillChange.php
@@ -1,0 +1,9 @@
+<?php
+
+#[Attribute(Attribute::TARGET_METHOD)]
+final class ReturnTypeWillChange
+{
+    public function __construct()
+    {
+    }
+}
--- a/src/Php81/README.md
+++ b/src/Php81/README.md
@@ -5,6 +5,7 @@ This component provides features added to PHP 8.1 core:
 
 - [`array_is_list`](https://php.net/array_is_list)
 - [`MYSQLI_REFRESH_REPLICA`](https://www.php.net/manual/en/mysqli.constants.php#constantmysqli-refresh-replica) constant
+- [`ReturnTypeWillChange`](https://wiki.php.net/rfc/internal_method_return_types)
 
 More information can be found in the
 [main Polyfill README](https://github.com/symfony/polyfill/blob/master/README.md).
--- a/src/Php81/composer.json
+++ b/src/Php81/composer.json
@@ -20,7 +20,8 @@
     },
     "autoload": {
         "psr-4": { "Symfony\\Polyfill\\Php81\\": "" },
-        "files": [ "bootstrap.php" ]
+        "files": [ "bootstrap.php" ],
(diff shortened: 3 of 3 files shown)


nicolas-grekas:
To you plan to use that in DebugClassLoader? We don't have to, since attributes do *have* to be instantiated ;)

derrabus:
`DebugClassLoader` should not instantiate the attribute and thus does not need to depend on the polyfill. I thought about adding the polyfill to components where we want to annotate methods with that attribute though.

Either way, providing a polyfill for the attribute and giving app developers the opertunity to `->newInstance()` without shooting themselves in the foot is a good thing, imho. 🙂

stof (approved):

src/Php81/Resources/stubs/ReturnTypeWillChange.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+#[Attribute]
```
nicolas-grekas, line 3: doesn't it have any flag? eg TARGET_METHOD?

nicolas-grekas: /cc @kocsismate should we add it to php-src?

derrabus: I actually doesn't, see my dumper output above.

stof: We should do the same than php-src. But it might make sense for php-src to configure the target there.

derrabus: Yes, not having a target on this attribute looks like a mistake.

stof: https://github.com/php/php-src/pull/6971/files#diff-7fa10e4571ac44e902ec9885a4854f676b2e7f5ccc21711a2b626ac9d6e3f289R289 makes me think that the attribute is actually already restricted to method targets, but that this is not properly exposed to Reflection
/cc @kocsismate @nikic

nikic: Thanks for the report, I'll look into it.

nikic: Should be fixed by https://github.com/php/php-src/commit/db6e60e744ab0d98d10b74cff3e897d5ff2f7574.

derrabus: I can confirm that reflection now reports `4` as argument which is the value of `Attribute::TARGET_METHOD`.

nicolas-grekas (approved):

src/Php81/Resources/stubs/ReturnTypeWillChange.php
derrabus: I've updated the screenshot.

stof (approved):

PR #355, opened 2021-05-20
[Php81] Optimize `array_is_list`

boesing:
Hey there,

I've used this function quite a lot and I see a massive performance gap between `webmozart/assert` and this polyfill.
https://3v4l.org/ruYVL

With this PR I synchronize the logic with that from `webmozart/assert`.
https://github.com/webmozarts/assert/blob/f07851c5b43e4cb502c24068620e9af6cbdd953f/src/Assert.php#L1823

### DIFF
--- a/src/Php81/Php81.php
+++ b/src/Php81/Php81.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\Polyfill\Php81;
 
+use function array_values;
+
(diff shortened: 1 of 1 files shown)


boesing:
So, php does not treat `\NAN` by reference as it does with integers? 
Interesting.

PR #356, opened 2021-05-20
[Php81] Optimize `array_is_list`

boesing:
Hey there,

I've used this function quite a lot and I see a massive performance gap between [webmozart/assert](https://github.com/webmozarts/assert/blob/f07851c5b43e4cb502c24068620e9af6cbdd953f/src/Assert.php#L1823) and this polyfill.
https://3v4l.org/ruYVL

This polyfill covers a niche case with `\NAN` and thus, my previous attempt in #355 failed.

So I've worked out another solution which is still a little bit faster than the current one.
https://3v4l.org/KiCg2

### DIFF
--- a/src/Php81/Php81.php
+++ b/src/Php81/Php81.php
@@ -20,7 +20,7 @@ final class Php81
 {
     public static function array_is_list(array $array): bool
     {
-        if ([] === $array) {
+        if ([] === $array || $array === array_values($array)) {
             return true;
         }
 
--- a/tests/Php81/Php81Test.php
+++ b/tests/Php81/Php81Test.php
@@ -26,6 +26,10 @@ class Php81Test extends TestCase
         $this->assertFalse(array_is_list(['a' => 'b']));
         $this->assertFalse(array_is_list([0 => 'a', 2 => 'b']));
         $this->assertFalse(array_is_list([1 => 'a', 2 => 'b']));
+
+        $x = ['key' => 2, NAN];
+        unset($x['key']);
+        $this->assertTrue(array_is_list($x));
     }
 
     /**


derrabus:
Can you also include associative arrays in you comparison? My expectation would be that the current implementation performs significantly better on non-list arrays.

boesing:
That is indeed correct. 
https://3v4l.org/Ef2pN

nicolas-grekas:
This patch might be worth it, WDYT?
Note that our test suite is not correct. The patch contains a better special case from [the original RFC](https://wiki.php.net/rfc/is_list).
```patch
diff --git a/src/Php81/Php81.php b/src/Php81/Php81.php
index 709c20b..8921446 100644
--- a/src/Php81/Php81.php
+++ b/src/Php81/Php81.php
@@ -20,7 +20,7 @@ final class Php81
 {
     public static function array_is_list(array $array): bool
     {
-        if ([] === $array) {
+        if (!$array || array_values($array) === $array) {
             return true;
         }
 
diff --git a/tests/Php81/Php81Test.php b/tests/Php81/Php81Test.php
index 17f5864..4271d0b 100644
--- a/tests/Php81/Php81Test.php
+++ b/tests/Php81/Php81Test.php
@@ -21,11 +21,14 @@ class Php81Test extends TestCase
     public function testArrayIsList()
     {
         $this->assertTrue(array_is_list([]));
-        $this->assertTrue(array_is_list([\NAN, 'foo', 123]));
         $this->assertFalse(array_is_list([1 => 'a', 0 => 'b']));
         $this->assertFalse(array_is_list(['a' => 'b']));
         $this->assertFalse(array_is_list([0 => 'a', 2 => 'b']));
         $this->assertFalse(array_is_list([1 => 'a', 2 => 'b']));
+
+        $x = ['key' => 2, NAN];
+        unset($x['key']);
+        $this->assertTrue(array_is_list($x));
     }
```

boesing:
That would optimize the speed for list arrays while preserving old functionality. 👍🏻

boesing:
@nicolas-grekas I've applied your suggestion. Let me know if something is missing 😊👍🏻

boesing:
@derrabus @nicolas-grekas Is there any TODO left open? Please let me know if I can help with anything.

derrabus (approved):

Sanchous98:
> @derrabus @nicolas-grekas Is there any TODO left open? Please let me know if I can help with anything.

Could you check, if "foreach" loop could be replaced by "array_keys($array) === range(0, count($array) - 1)". I have checked and it works slightly faster.

nicolas-grekas:
@Sanchous98 I'm not convinced the proposals you make would be an improvement:

> "array_keys($array) === range(0, count($array) - 1)"

this is going to allocate two arrays

> array_keys($array) vs $value

this is also going to allocate an array to save a local variable which costs nothing.

Please provide extensive numbers to back your claims if you think we should reconsider. For now I think the attached patch is the one we should merge.

nicolas-grekas:
Thank you @boesing.

PR #359, opened 2021-06-05
[PHP80] Prevent redefinition of PHP 8.0 classes

Firtzberg:
Not redeclaring polyfilled classes when PHP version is 8.0 or above.
The same approach is used for the Stringable interface in [this commit](https://github.com/symfony/polyfill/commit/8a3e84999b869b30bc8688eb3784681028007d5e).

One of my scripts runs [`opcache_compile_file`](https://www.php.net/manual/en/function.opcache-compile-file.php) on vendor files, including symfony. I'm in the process of upgrading to PHP 8. Running `opcache_compile_file` on either of the modified files causes a fatal error `Fatal error: Cannot declare class ValueError/UnhandledMatchError, because the name is already in use`.
That's not the case on PHP versions below 8.

### DIFF
--- a/src/Php80/Resources/stubs/UnhandledMatchError.php
+++ b/src/Php80/Resources/stubs/UnhandledMatchError.php
@@ -1,5 +1,7 @@
 <?php
 
-class UnhandledMatchError extends Error
-{
+if (\PHP_VERSION_ID < 80000) {
+    class UnhandledMatchError extends Error
+    {
+    }
 }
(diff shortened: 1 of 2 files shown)


Firtzberg:
ping

derrabus:
> One of my scripts runs [`opcache_compile_file`](https://www.php.net/manual/en/function.opcache-compile-file.php) on vendor files, including symfony.

As mentioned elsewhere, you shouldn't do that. Symfony ships classes that might require optional dependencies or a higher PHP version. Even if we apply your fix here, your script will continue to explode elsewhere.

derrabus (approved):
I'm fine with the changes despite disagreeing with the motivation behind them. 😉

nicolas-grekas:
Thank you @Firtzberg.

PR #360, opened 2021-06-05
[PHP 8.1] [PHP 7.3] Preventing redeclaration of PHP version related class stubs.

Firtzberg:
Same as https://github.com/symfony/polyfill/pull/359, only for other PHP version related class stubs.


nicolas-grekas:
Thank you @Firtzberg.

PR #361, opened 2021-06-08
Fix mb_scrub to be compatible with php < 7.4

sserbin:
Null coalesce assignment operator was only added in php 7.4


derrabus:
Duplicate #326

PR #362, opened 2021-06-24
[PHP 7.4]: Add WeakReference class

Sanchous98:
Using refcount value, provided by "debug_zval_dump", we can implement smth. similar to WeakReference from PHP 7.4.

### DIFF
--- a/src/Php74/bootstrap.php
+++ b/src/Php74/bootstrap.php
@@ -24,6 +24,6 @@ if (!function_exists('mb_str_split') && function_exists('mb_substr')) {
 if (!function_exists('password_algos')) {
     function password_algos() { return p\Php74::password_algos(); }
 }
-if (!class_exists("WeakReference")) {
-    final class WeakReference extends p\Resources\WeakReference {}
+if (!class_exists('WeakReference')) {
+    final class WeakReference extends p\WeakReference {}
 }
--- a/src/Php74/bootstrap.php
+++ b/src/Php74/bootstrap.php
@@ -24,3 +24,6 @@ if (!function_exists('mb_str_split') && function_exists('mb_substr')) {
 if (!function_exists('password_algos')) {
     function password_algos() { return p\Php74::password_algos(); }
 }
+if (!class_exists("WeakReference")) {
+    final class WeakReference extends p\Resources\WeakReference {}
+}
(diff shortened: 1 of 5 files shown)


stof:
This reference is not weak. If you don't call `get()`, the reference is still kept strongly, which prevents collecting the object. This defeats the purpose of the polyfill IMO.

Sanchous98:
> This reference is not weak. If you don't call `get()`, the reference is still kept strongly, which prevents collecting the object. This defeats the purpose of the polyfill IMO.

Yes, that's right, object won't be collected, it's a sort of "hack" that will make PHP 7.4 code work properly in PHP <7.3 with a some loss of memory and performance. If we create a WeakReference in 7.4, we'll get "null" if object passed to WeakReference was unset. This implementation does the same

src/Php74/bootstrap.php
```diff
@@ -24,3 +24,6 @@ function mb_str_split($string, $length = 1, $encoding = null) { return p\Php74::
 if (!function_exists('password_algos')) {
     function password_algos() { return p\Php74::password_algos(); }
 }
+if (!class_exists("WeakReference")) {
```
OskarStark, line 27: Please use single quotes

stof:
I'm still not convinced about adding this hacky class in the polyfills. The main (only?) reason to use WeakReference is to avoid memory leaks. And that polyfill does not allow to implement that part. So the code would still not have the intended effect on older PHP versions.

keradus:
maybe polyfill would not behave as great as native implementation, but it would allow folks to write code that behaves great on the newest PHP version, while still executes on lower PHP version (maybe less performant, but at least not crashing on unknown classes)

Sanchous98:
> I'm still not convinced about adding this hacky class in the polyfills. The main (only?) reason to use WeakReference is to avoid memory leaks. And that polyfill does not allow to implement that part. So the code would still not have the intended effect on older PHP versions.

I'll try to use tick function to unset variable, but I'm not sure if it won't impact the performance too much.

jderusse:
Agreed with @keradus . This does not have to be an efficient weak reference. it's about allowing people writing the same code being compatible with PHP < 8

nicolas-grekas:
I'm closing as discussed, thanks for proposing.

PR #366, opened 2021-07-28
Allow null in str_*() functions

### LINKED ISSUE #365: Parameter types of PHPH 8.0 str_*() functions too strict
Our polyfills of the `str_*()` functions introduced with PHP 8.0 seem to be a bit too strict about typing. Our version throws a `TypeError` when `null` is passed to them while PHP's own implementation treats `null` like an empty string.

https://3v4l.org/0VShO

This has caused some issues on Symfony after we have migrated to the polyfilled functions, for example symfony/symfony#42290

derrabus:
Fixes #365

### DIFF
--- a/src/Php80/bootstrap.php
+++ b/src/Php80/bootstrap.php
@@ -26,13 +26,13 @@ if (!function_exists('preg_last_error_msg')) {
     function preg_last_error_msg(): string { return p\Php80::preg_last_error_msg(); }
 }
 if (!function_exists('str_contains')) {
-    function str_contains(string $haystack, string $needle): bool { return p\Php80::str_contains($haystack, $needle); }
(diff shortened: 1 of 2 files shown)


chalasr (approved):

chalasr:
Thank you @derrabus.

Tobion:
Passing null is deprecated in PHP 8.1. Doesn't this solution prevent people from seeing the deprecation when using the polyfill?
So I think the polyfill should also raise a deprecation in PHP 8.1.

derrabus:
@Tobion The polyfilled code is never used on PHP 8.1 because PHP has native implementations of those functions since 8.0.

Tobion:
Right... Was thinking these functions are like an extension that could be disabled.

PR #371, opened 2021-09-14
Mimic deprecation from PHP 8.1 in IntlDateFormatter

nicolas-grekas:


src/Intl/Icu/IntlDateFormatter.php
```diff
@@ -137,6 +137,16 @@ public function __construct(?string $locale, ?int $dateType, ?int $timeType, $ti
             throw new MethodArgumentValueNotImplementedException(__METHOD__, 'calendar', $calendar, 'Only the GREGORIAN calendar is supported');
         }
 
+        if (\PHP_VERSION_ID >= 80100) {
+            if (null === $dateType) {
+                @trigger_error('Passing null to parameter #2 ($dateType) of type int is deprecated', \E_USER_DEPRECATED);
```
stof, line 142: shouldn't it be a non-silenced one to actually mimic it ?

nicolas-grekas: We already merged some and they are silenced. Let's keep them silenced IMHO.

derrabus (approved):

PR #380, opened 2021-12-30
Prevent Cannot declare interface Stringable with multiple vendor folders

shyim:
Hey,

PhpStan uses this polyfill and inspects using reflection the project files and loads again a Stringable interface. This breaks. 
A simple interface check fixes it for me. Idk if this is a good way to fix this behaviour


shyim:
The root cause is the `container.dumper.inline_class_loader` option

PR #388, opened 2022-02-02
fix: don't define this class twice

sflx-matze:
In the detached Git-CI this class will be defined twice, for the PhpStan Check, if you are using the symfony package and the ObjectManager class for Doctrine parsing. This fix will define it only once.

### DIFF
--- a/src/Php80/Resources/stubs/Stringable.php
+++ b/src/Php80/Resources/stubs/Stringable.php
@@ -1,6 +1,6 @@
 <?php
 
-if (\PHP_VERSION_ID < 80000) {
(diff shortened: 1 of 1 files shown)


derrabus (approved):

nicolas-grekas:
When does this happen?
We fixed this in https://github.com/symfony/symfony/pull/45258, isn't that enough?
I'd prefer not doing conditional definition unless we understand the reason.

nicolas-grekas:
Closing in the hope that https://github.com/symfony/symfony/pull/45258 fixed the root of the issue.

PR #390, opened 2022-02-17
Fix outdated DTD in configuration

WinterSilence:
Update DTD to version declared in `required-dev` in `composer.json` to fix IDE's warnings.
Remove attribute `backupGlobals="false"` because it's default value.


nicolas-grekas:
Thank you @WinterSilence.

Issue #376, opened 2021-11-13
`tokenizer` extension


WinterSilence:
What's about add `tokenizer` extension containing class `PhpToken`?

keradus:
-1 for partial polyfill
if you claim that `tokenizer` extension is polyfilled, yet provide only `PhpToken` but not `token_get_all` function - PHP CS Fixer will fail to work

WinterSilence:
@keradus it's polyfill for php less 8.0 like as intl Normalizer, `tokenizer` extension still required.

nicolas-grekas:
> if you claim that tokenizer extension is polyfilled, yet provide only PhpToken but not token_get_all function - PHP CS Fixer will fail to work

I agree, but can't this be solved by *not* claiming that this would polyfill ext-tokenizer (since it doesn't)?

WinterSilence:
of course, we can implement `token_get_all()` (I saw few solutions), but this is extremely expensive at maintaining

nicolas-grekas:
We'd better *not* do this :)

nicolas-grekas:
It could be fine to add this class to the Php80 polyfill and define the class only if `\function_exists('token_get_all')`.
Can you please send a PR @WinterSilence? (with some test cases of course)

WinterSilence:
@nicolas-grekas send PR to current repo or [symfony/polyfill-php80](https://github.com/symfony/polyfill-php80)?

nicolas-grekas:
To the current repo please, the other is an automatic subtree split.

WinterSilence:
@nicolas-grekas As I'm see, package use custom legacy PhpUnit... I'm don't have time to explore this history museum and can write tests only for PhpUnit8+.

nicolas-grekas:
That's fine, just send the PR with what you have, we'll find a way.

WinterSilence:
@nicolas-grekas also, if `"symfony/phpunit-bridge": "^5.3|^6.0"` require same versions of phpunit, then need update `phpunit.xml.dist`: `xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/5.0/phpunit.xsd"` or `xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/6.0/phpunit.xsd"` to calm "barking" IDE (:

PR #392, opened 2022-02-25
[Php80] Add PhpToken polyfill

WinterSilence:
Initial issue #376 
Tests PHP 8

### DIFF
--- a/src/Php80/PhpToken.php
+++ b/src/Php80/PhpToken.php
@@ -1,0 +1,103 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Php80;
+
+/**
+ * @author Fedonyuk Anton <info@ensostudio.ru>
+ *
+ * @internal
+ */
+class PhpToken implements \Stringable
+{
+    /**
+     * @var int
+     */
+    public $id;
+
+    /**
+     * @var string
+     */
+    public $text;
+
+    /**
+     * @var int
+     */
+    public $line;
+
+    /**
+     * @var int
+     */
+    public $pos;
+
+    public function __construct(int $id, string $text, int $line = -1, int $position = -1)
+    {
+        $this->id = $id;
+        $this->text = $text;
+        $this->line = $line;
+        $this->pos = $position;
+    }
+
+    public function getTokenName(): ?string
+    {
+        if ('UNKNOWN' === $name = token_name($this->id)) {
+            $name = \strlen($this->text) > 1 || \ord($this->text) < 32 ? null : $this->text;
+        }
+
+        return $name;
+    }
+
+    /**
+     * @param int|string|array $kind
+     */
+    public function is($kind): bool
+    {
+        foreach ((array) $kind as $value) {
+            if (\in_array($value, [$this->id, $this->text], true)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public function isIgnorable(): bool
+    {
+        return \in_array($this->id, [\T_WHITESPACE, \T_COMMENT, \T_DOC_COMMENT, \T_OPEN_TAG], true);
+    }
+
+    public function __toString(): string
+    {
+        return (string) $this->text;
+    }
+
+    /**
+     * @return static[]
+     */
+    public static function tokenize(string $code, int $flags = 0): array
+    {
+        $line = 1;
+        $position = 0;
+        $tokens = token_get_all($code, $flags);
+        foreach ($tokens as $index => $token) {
+            if (\is_string($token)) {
+                $id = \ord($token);
+                $text = $token;
+            } else {
+                [$id, $text, $line] = $token;
+            }
+            $tokens[$index] = new static($id, $text, $line, $position);
+            $position += \strlen($text);
+        }
+
+        return $tokens;
+    }
+}
--- a/src/Php80/Resources/stubs/PhpToken.php
+++ b/src/Php80/Resources/stubs/PhpToken.php
@@ -1,0 +1,7 @@
+<?php
+
+if (\PHP_VERSION_ID < 80000 && \extension_loaded('tokenizer')) {
+    class PhpToken extends Symfony\Polyfill\Php80\PhpToken
+    {
+    }
+}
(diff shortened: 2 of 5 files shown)


src/Php80/PhpToken.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * @return static[]
+     */
+    public static function tokenize(string $code, int $flags = 0): array
+    {
+        $tokens = [];
+        $line = 1;
+        $position = 0;
+        foreach (\token_get_all($code, $flags) as $token) {
```
keradus, line 81: maybe array-map, so we don't need to extend $tokens each iteration?

WinterSilence: @keradus can you explain your idea? you mean `$tokens = \token_get_all($code, $flags);`?

src/Php80/Resources/stubs/PhpToken.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+if (PHP_VERSION_ID < 80000 && !class_exists('PhpToken') && function_exists('token_get_all')) {
```
nicolas-grekas, line 3: no need for the class_exists check, and better use extension_loaded which can be optimized by opcache:
`if (PHP_VERSION_ID < 80000 && extension_loaded('tokenizer')) {`

src/Php80/PhpToken.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+namespace Symfony\Polyfill\Php80;
```
nicolas-grekas, line 3: please add the license header

composer.json
```diff
@@ -65,6 +65,9 @@
             "src/Php73/Resources/stubs"
         ]
     },
+    "autoload-dev": {
+        "psr-4": { "Symfony\\Polyfill\\Tests\\": "tests/" }
```
nicolas-grekas, line 69: this is not needed

src/Php80/PhpToken.php
```diff
@@ -0,0 +1,110 @@
...
+use const T_COMMENT;
+use const T_DOC_COMMENT;
+use const T_OPEN_TAG;
+use const T_WHITESPACE;
+
+/**
+ * Tokenizer: PhpToken polyfill.
+ *
+ * @author Fedonyuk Anton <info@ensostudio.ru>
+ * @link https://www.php.net/manual/en/class.phptoken.php
```
nicolas-grekas, line 26: should be removed, we don't add such links in the codebase

src/Php80/PhpToken.php
```diff
@@ -0,0 +1,110 @@
...
+
+use function token_name;
+
+use const T_COMMENT;
+use const T_DOC_COMMENT;
+use const T_OPEN_TAG;
+use const T_WHITESPACE;
+
+/**
+ * Tokenizer: PhpToken polyfill.
```
nicolas-grekas, line 23: Can be removed

tests/Php80/PhpTokenTest.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Symfony\Polyfill\Tests\Php80;
```
nicolas-grekas, line 3: please add the license header here also

tests/Php80/PhpTokenTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+use PhpToken;
+use Symfony\Polyfill\Php80\PhpToken as PhpTokenPolyfill;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Tests of PhpToken polyfill, compares polyfill and native instance.
```
nicolas-grekas, line 10: can be removed also

tests/Php80/PhpTokenTest.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Symfony\Polyfill\Tests\Php80;
+
+use PhpToken;
```
nicolas-grekas, line 5: should be removed

src/Php80/PhpToken.php
```diff
@@ -0,0 +1,98 @@
...
+        return $name;
+    }
+
+    /**
+     * @param int|string|array $kind
+     */
+    public function is($kind): bool
+    {
+        $kind = (array) $kind;
+        if (\is_int(\current($kind))) {
```
nicolas-grekas, line 62: won't this fail when there are more that one item in $kind and not all of them are integers?
can you also add a test case that covers the `getTokenName` check?

tests/Php80/PhpTokenTest.php
```diff
@@ -0,0 +1,75 @@
...
+        $polyfillToken = new PhpTokenPolyfill(\T_ECHO, 'echo');
+        $this->assertEquals($token->getTokenName(), $polyfillToken->getTokenName());
+        // single char token
+        $token = new PhpToken(\ord(';'), ';');
+        $polyfillToken = new PhpTokenPolyfill(\ord(';'), ';');
+        $this->assertEquals($token->getTokenName(), $polyfillToken->getTokenName());
+        // unknown token
+        $token = new PhpToken(10000 , "\0");
+        $polyfillToken = new PhpTokenPolyfill(10000 , "\0");
+        $this->assertEquals($token->getTokenName(), $polyfillToken->getTokenName());
```
nicolas-grekas, line 44: let's use assertSame everywhere possible

WinterSilence: @nicolas-grekas what's reason? it's polyfill class and must be similar

nicolas-grekas: To ensure that types are also matching

src/Php80/PhpToken.php
WinterSilence: @nicolas-grekas yes, need rewrite and test every value

tests/Php80/PhpTokenTest.php
WinterSilence: @nicolas-grekas ok

src/Php80/PhpToken.php
WinterSilence: @nicolas-grekas fixed. In self defense: that such a possibility is undocumented. :)

WinterSilence: @nicolas-grekas 
 > can you also add a test case that covers the getTokenName check?

removed, [native class don't use it](https://3v4l.org/It16L#v8.1.3)

nicolas-grekas:
Thank you @WinterSilence.

WinterSilence:
@nicolas-grekas thank you too

PR #393, opened 2022-02-26
Fix Php80Test::testFdiv()

WinterSilence:
Need using `$this->assertEqualsWithDelta()` to compare float values.
Why not use `$this->assertEquals()` or `$this->assertSame()`? Because `(0.1 * 0.1) / 0.1` not equal 0.1, it's ~0.10000000000000002

### DIFF
--- a/tests/Php80/Php80Test.php
+++ b/tests/Php80/Php80Test.php
@@ -32,9 +32,9 @@ class Php80Test extends TestCase
         } catch (\DivisionByZeroError $e) {
             $result = $expected;
         }
-        $this->assertSame($expected, $result);
+        $this->assertEqualsWithDelta($expected, $result, 0.001);
         // Cast to string to detect negative zero "-0"
-        $this->assertSame((string) $expected, (string) $result);
+        $this->assertEquals(number_format($expected, 3), number_format($result, 3));
     }
 
     /**


nicolas-grekas:
This reminds me https://github.com/symfony/symfony/pull/45475
This happens since https://github.com/sebastianbergmann/phpunit/pull/4874

/cc @sebastianbergmann is that legit?

WinterSilence:
@nicolas-grekas your question to Sebastian is incorrect - this package use PhpUnit 5-6, https://github.com/sebastianbergmann/phpunit/pull/4874 is PhpUnit 8

derrabus (approved):

nicolas-grekas:
This PR is definitely motivated by https://github.com/sebastianbergmann/phpunit/pull/4874
The CI is currently green on versions that don't have that patch, and red on jobs that use recent PHP versions.

nicolas-grekas:
Issue fixed in a932445
Thanks for submitting.

WinterSilence:
@nicolas-grekas this PR motivated by https://github.com/symfony/polyfill/runs/5367868196?check_suite_focus=true
I'm don't see before #4874
It's not fixed it's f..n shame

nicolas-grekas:
>  It's not fixed it's f..n shame

what do you mean? yes it is, see newest CI runs as a proof.

WinterSilence:
@nicolas-grekas this PR too do this, but you close it and update test data instead using correct assertion.

nicolas-grekas:
This PR works around the test data, which was wrong in the first place. We want an exact check, not an approximation here.

WinterSilence:
@nicolas-grekas sorry, my fail

PR #395, opened 2022-03-04
[PHP 7.2] add json constants JSON_INVALID_UTF8_*

ginnerpeace:
Available as of PHP 7.2.0.
- JSON_INVALID_UTF8_IGNORE
- JSON_INVALID_UTF8_SUBSTITUTE

https://www.php.net/manual/en/json.constants.php


IonBazan:
I don't think it's a good idea to add consts that will not have any effect with`json_encode` and `json_decode` on PHP <7.2 as this could be misleading. Libraries might rely on existence of these constants to determine if invalid UTF characters would be ignored (or substituted), skipping some sanity or security checks afterwards.

nicolas-grekas:
Agreed with @IonBazan 
This is a policy in this repo, to not add stub polyfills.

PR #404, opened 2022-04-11
Polyfill SensitiveParameter and SensitiveParameterValue

TimWolla:


src/Php82/Php82.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+namespace Symfony\Polyfill\Php82;
+
+/**
+ * @author Tim Düsterhus <duesterhus@woltlab.com>
+ *
+ * @internal
+ */
+final class Php82
```
derrabus, line 19: Do we need this class?

src/Php82/SensitiveParameterValue.php
```diff
@@ -0,0 +1,57 @@
...
+    public function __serialize(): array
+    {
+        throw new \Exception("Serialization of 'SensitiveParameterValue' is not allowed");
+    }
+
+    public function __unserialize(array $data): void
+    {
+        throw new \Exception("Unserialization of 'SensitiveParameterValue' is not allowed");
+    }
+
```
derrabus, line 47:
```suggestion
```

Implementing `__sleep()` and `__wakeup()` should be enough.

src/Php82/bootstrap.php
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Php82 as p;
+
+if (\PHP_VERSION_ID >= 80200) {
+    return;
+}
```
derrabus, line 16: Do we need this file?

tests/Php82/SensitiveParameterValueTest.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        $implementations = [
+            [new SensitiveParameterValuePolyfill('secret')],
+        ];
+
+        if (\class_exists('SensitiveParameterValue')) {
+            $implementations[] = [new \SensitiveParameterValue('secret')];
+        }
+
+        return $implementations;
```
derrabus, line 32:
```suggestion
        yield [new SensitiveParameterValuePolyfill('secret')];

        if (\class_exists('SensitiveParameterValue')) {
            yield [new \SensitiveParameterValue('secret')];
        }
```

src/Php82/bootstrap.php
TimWolla: Not yet, but I've created it for consistency with the other stubs. It can be used for these constants (if you want to polyfill those / if it's possible to polyfill those): https://github.com/php/php-src/blob/b9f4fb8aefaeb3802d6f79f6aad7029b63e488a2/UPGRADING#L177-L181

src/Php82/Php82.php
TimWolla: See: https://github.com/symfony/polyfill/pull/404#discussion_r848428986

derrabus:
For reference, this PR adds polyfills introduced by this RFC: https://wiki.php.net/rfc/redact_parameters_in_back_traces

src/Php82/bootstrap.php
TimWolla: Apparently TestListenerTrait requires the existence of `bootstrap.php`: https://github.com/symfony/polyfill/blob/66dab8dc6c789d7a0b2010238ce55547b48a1941/src/Util/TestListenerTrait.php#L56. I don't think it makes sense to adjust that, because `bootstrap.php` will likely be required for PHP 8.2 at one point in the future.

I've removed the class file, but preserved the `bootstrap.php` due to the reasoning above.

TimWolla:
Hi Folks, are you currently waiting for me here? From my side I believe I answered all questions and consider this PR finished, unless you request some changes. The build failures appear to be unrelated and should likely be fixed in a separate PR.

tests/Php82/SensitiveParameterValueTest.php
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @author Tim Düsterhus <duesterhus@woltlab.com>
+ */
+class SensitiveParameterValueTest extends TestCase
+{
+    public function sensitiveParameterValueProvider()
+    {
+        yield [new SensitiveParameterValuePolyfill('secret')];
+
+        if (\class_exists('SensitiveParameterValue')) {
```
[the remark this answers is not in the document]
nicolas-grekas: Stof is right, these should not be needed. What is missing is an entry in the class-map section of the root composer.json file.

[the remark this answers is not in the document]
TimWolla: @nicolas-grekas Thanks, fixed. Shall I also squash the full series?

nicolas-grekas:
Thank you @TimWolla.

PR #405, opened 2022-05-09
Passing null to preg_split() throws deprecation on PHP 8.1

simPod:
> preg_split(): Passing null to parameter #3 ($limit) of type int is deprecated


simPod:
I did full text search to find those but I've failed. Thanks. Done.

nicolas-grekas (approved):

derrabus (approved):

nicolas-grekas:
Thank you @simPod.

PR #414, opened 2022-08-22
Support escaped unicode character sequences in idna test files

TRowbotham:
The IDNA [test file](https://www.unicode.org/Public/idna/15.0.0/IdnaTestV2.txt) for Unicode 15.0.0 has started using escaped
unicode character sequences. The [spec](https://www.unicode.org/reports/tr46/#Format) says that they can be in the form
of either \uXXXX or \x{XXXX}.

Additionally, remove a now irrelevant conditional that worked around a bug in previous test files.


nicolas-grekas:
Thank you @TRowbotham.

PR #416, opened 2022-10-28
PHP 8.3: json_validate polyfill

IonBazan:
This PR adds `json_validate()` polyfill - see: https://wiki.php.net/rfc/json_validate.
Tests examples were taken from the php-src implementation. 

Test failures are unrelated.
Float tests are fixed in #417, remaining PHP 8.2 mbstring test can be sorted out separately.

### DIFF
--- a/src/Php83/Php83.php
+++ b/src/Php83/Php83.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Php83;
+
+/**
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
+ * @internal
+ */
+final class Php83
+{
+    private const JSON_MAX_DEPTH = 0x7FFFFFFF; // see https://www.php.net/manual/en/function.json-decode.php
+
+    public static function json_validate(string $json, int $depth = 512, int $flags = 0): bool
+    {
+        if (0 !== $flags && \defined('JSON_INVALID_UTF8_IGNORE') && \JSON_INVALID_UTF8_IGNORE !== $flags) {
+            throw new \ValueError('json_validate(): Argument #3 ($flags) must be a valid flag (allowed flags: JSON_INVALID_UTF8_IGNORE)');
+        }
+
+        if ($depth <= 0) {
+            throw new \ValueError('json_validate(): Argument #2 ($depth) must be greater than 0');
+        }
+
+        if ($depth >= self::JSON_MAX_DEPTH) {
+            throw new \ValueError(sprintf('json_validate(): Argument #2 ($depth) must be less than %d', self::JSON_MAX_DEPTH));
+        }
+
+        json_decode($json, null, $depth, $flags);
+
+        return \JSON_ERROR_NONE === json_last_error();
+    }
+}
--- a/src/Php83/README.md
+++ b/src/Php83/README.md
@@ -1,0 +1,14 @@
+Symfony Polyfill / Php83
+========================
+
+This component provides features added to PHP 8.3 core:
+
+- [`json_validate`](https://wiki.php.net/rfc/json_validate)
+
+More information can be found in the
+[main Polyfill README](https://github.com/symfony/polyfill/blob/main/README.md).
+
+License
+=======
+
+This library is released under the [MIT license](LICENSE).
--- a/tests/Php83/Php83Test.php
+++ b/tests/Php83/Php83Test.php
@@ -1,0 +1,86 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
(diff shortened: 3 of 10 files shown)


src/Php83/Php83.php
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @author Ion Bazan <ion.bazan@gmail.com>
+ *
+ * @internal
+ */
+final class Php83
+{
+    public static function json_validate(string $json, int $depth = 512, int $flags = 0): bool
+    {
+        if ($flags !== 0 && $flags !== 1048576) {
```
IonBazan, line 23: This magic number is value of `JSON_INVALID_UTF8_IGNORE` constant which is available from PHP 7.2+. Should we provide it in `symfony/polyfill-php72` and require it in this polyfill instead?

src/Php83/README.md
```diff
@@ -0,0 +1,14 @@
+Symfony Polyfill / Php83
+========================
+
+This component provides features added to PHP 8.3 core:
+
+- [`array_is_list`](https://wiki.php.net/rfc/json_validate)
```
Ayesh, line 6:
```suggestion
- [`json_validate`](https://wiki.php.net/rfc/json_validate)
```

src/Php83/Php83.php
```diff
@@ -0,0 +1,39 @@
...
+            throw new \ValueError('json_validate(): Argument #3 ($flags) must be a valid flag (allowed flags: JSON_INVALID_UTF8_IGNORE)');
+        }
+
+        if ($depth <= 0 ) {
+            throw new \ValueError('json_validate(): Argument #2 ($depth) must be greater than 0');
+        }
+
+        if ($depth >= PHP_INT_MAX) {
```
Ayesh, line 31: Perhaps we can make `PHP_INT_MAX` (there is another in the next line) fully-qualified too?

tests/Php83/Php83Test.php
```diff
@@ -0,0 +1,77 @@
...
+namespace Symfony\Polyfill\Tests\Php83;
+
+use PHPUnit\Framework\TestCase;
+
+class Php83Test extends TestCase
+{
+    /**
+     * @covers \Symfony\Polyfill\Php83\Php83::json_validate
+     */
+    public function testJsonValidate()
```
OskarStark, line 21: What about using a dataProvider ?

src/Php83/Php83.php
IonBazan: Replacing this with hardcoded `2147483647` value which apparently is max depth in PHP json parser. See:
https://3v4l.org/gPIPk/rfc#vgit.master

I will be submitting a separate PHP issue for this.

src/Php83/Php83.php
derrabus: We decided to not polyfill constants for JSON flags because we cannot polyfill the functionality that is controlled via those constants.

In our case, you can use the real `JSON_INVALID_UTF8_IGNORE` constant if it's defined. If it isn't defined, passing that magic number is not supported either because the `json_decode()` function that you're calling with that flag wouldn't know what to do with it.

```suggestion
        if ($flags !== 0 && !(defined('JSON_INVALID_UTF8_IGNORE') && JSON_INVALID_UTF8_IGNORE === $flags)) {
```

derrabus (changes requested):

derrabus (approved):

src/Php83/Php83.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    private const JSON_MAX_DEPTH = (1 << 31) - 1; // see https://www.php.net/manual/en/function.json-decode.php
+    public static function json_validate(string $json, int $depth = 512, int $flags = 0): bool
+    {
+        if ($flags !== 0 && defined('JSON_INVALID_UTF8_IGNORE') && JSON_INVALID_UTF8_IGNORE !== $flags) {
+            throw new \ValueError('json_validate(): Argument #3 ($flags) must be a valid flag (allowed flags: JSON_INVALID_UTF8_IGNORE)');
+        }
+
+        if ($depth <= 0 ) {
+            throw new \ValueError('json_validate(): Argument #2 ($depth) must be greater than 0');
```
GromNaN, line 29: The `\ValueError` was introduced by PHP 8.0.

This polyfill should require `symfony/polyfill-php80` or `php>=8` ?

GromNaN (changes requested):

Ayesh (approved):
Nice, looks good to me.
I'm not sure what's the best way to tackle the tests that fail because of the lack of precision in the float value assertions.

Thank you.

src/Php83/Php83.php
derrabus: Indeed. Bumping to PHP 8 is not an option for now, I think. Let's depend on `symfony/polyfill-php80`.

IonBazan: Fixed and rebased - please have another look 👍🏻

derrabus (approved):

GromNaN (approved):

stof (approved):

nicolas-grekas:
Thank you @IonBazan.

PR #418, opened 2022-11-03
CS fix

nicolas-grekas:


derrabus (approved):

derrabus:
Needs a rebase.

src/Mbstring/Mbstring.php
```diff
@@ -80,7 +80,7 @@ final class Mbstring
 
     public static function mb_convert_encoding($s, $toEncoding, $fromEncoding = null)
     {
-        if (\is_array($fromEncoding) || ($fromEncoding !== null && false !== strpos($fromEncoding, ','))) {
+        if (\is_array($fromEncoding) || (null !== $fromEncoding && str_contains($fromEncoding, ','))) {
```
derrabus, line 83: As long as we don't require the php80 polyfill in the mbstring package, this change is dangerous. We should probably deactivate the corresponding fixer.

derrabus (changes requested):

PR #421, opened 2022-11-30
Fix json_validate rejecting valid maximum depth

Emix33:
Even though the error message for json_decode will say 'Argument #3 ($depth) must be less than 2147483647', the actual behavior of the method allows this (https://3v4l.org/QdfHt) and the official online documentation also confirms this behavior (https://www.php.net/manual/en/function.json-decode.php ' The value must be greater than 0, and less than or equal to 2147483647'). I have not been able to confirm whether 2147483647 is accepted by json_decode on a 32 bit system (where it is equal to PHP_MAX_INT) as I do not have access to one. (3v4l also uses 64 bit PHP)

### DIFF
--- a/tests/Php83/Php83Test.php
+++ b/tests/Php83/Php83Test.php
@@ -46,6 +46,7 @@ class Php83Test extends TestCase
         yield [true, '{ "test": {"foo": "bar"}, "test2": {"foo" : "bar" }, "test3": {"foo" : "bar" } }'];
         yield [false, '{"key1":"value1", "key2":"value2"}', 'Maximum stack depth exceeded', 1];
         yield [false, "\"a\xb0b\"", 'Malformed UTF-8 characters, possibly incorrectly encoded'];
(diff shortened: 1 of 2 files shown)


IonBazan (approved):
Good catch, thanks @Emix33!

tests/Php83/Php83Test.php
```diff
@@ -73,8 +74,9 @@ public function testInvalidOptionsProvided(int $depth, int $flags, string $expec
     public static function invalidOptionsProvider(): iterable
     {
         yield [0, 0, 'json_validate(): Argument #2 ($depth) must be greater than 0'];
-        yield [\PHP_INT_MAX, 0, 'json_validate(): Argument #2 ($depth) must be less than 2147483647'];
-
+        if(\PHP_INT_MAX > 2147483647){
```
IonBazan, line 77: Please just run a CS fixer here or apply this patch.
```suggestion
        if (\PHP_INT_MAX > 2147483647) {
```

nicolas-grekas:
Thank you @Emix33.

Issue #422, opened 2022-12-16
polyfill-mbstring: mb_check_encoding crashes for array input


ondrejmirtes:
Hi,
`mb_check_encoding` accepts an array of strings: https://3v4l.org/i5MBi

But `Symfony\Polyfill\Mbstring\Mbstring::mb_check_encoding` crashes with the same input:

```
Uncaught TypeError: preg_match(): Argument #2 ($subject) must be of type string, array given in /Users/ondrej/Development/phpstan/vendor/symfony/polyfill-mbstring/Mbstring.php:440
#0 /Users/ondrej/Development/phpstan/vendor/symfony/polyfill-mbstring/Mbstring.php(440): preg_match('//u', Array)
#1 /Users/ondrej/Development/phpstan/vendor/symfony/polyfill-mbstring/Mbstring.php(416): Symfony\Polyfill\Mbstring\Mbstring::mb_detect_encoding(Array, Array)
#2 /Users/ondrej/Development/phpstan/src/Type/Php/StrCaseFunctionsReturnTypeExtension.php(82): Symfony\Polyfill\Mbstring\Mbstring::mb_check_encoding(Array, 'UTF-8')
```

Tested with symfony/polyfill-mbstring v1.27.0.

Thanks :)

PR #426, opened 2023-01-27
add mb_check_encoding with array value

jeremyFreeAgent:
I tried to address the https://github.com/symfony/polyfill/issues/422 issue.

First time on that project. I'm not sure about what I've done because I'm afraid we cannot do anything for PHP <7.2.

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -406,6 +406,12 @@ final class Mbstring
 
     public static function mb_check_encoding($var = null, $encoding = null)
     {
+        if (PHP_VERSION_ID < 70200 && \is_array($var)) {
+            trigger_error('mb_check_encoding() expects parameter 1 to be string, array given', \E_USER_WARNING);
+
+            return null;
+        }
+
         if (null === $encoding) {
             if (null === $var) {
                 return false;
@@ -413,7 +419,21 @@ final class Mbstring
             $encoding = self::$internalEncoding;
         }
 
-        return self::mb_detect_encoding($var, [$encoding]) || false !== @iconv($encoding, $encoding, $var);
+        if (!\is_array($var)) {
+            return self::mb_detect_encoding($var, [$encoding]) || false !== @iconv($encoding, $encoding, $var);
+        }
+
+        foreach ($var as $key => $value) {
+            if (!self::mb_check_encoding($key, $encoding)) {
+                return false;
+            }
+            if (!self::mb_check_encoding($value, $encoding)) {
+                return false;
+            }
+        }
+
+        return true;
+
     }
 
     public static function mb_detect_encoding($str, $encodingList = null, $strict = false)
--- a/tests/Mbstring/MbstringTest.php
+++ b/tests/Mbstring/MbstringTest.php
@@ -467,6 +467,53 @@ class MbstringTest extends TestCase
     }
 
     /**
+     * @covers \Symfony\Polyfill\Mbstring\Mbstring::mb_check_encoding
+     *
+     * @requires PHP 7.2
+     */
+    public function testCheckEncodingWithArrayValue()
+    {
+        $this->assertTrue(mb_check_encoding(['aςσb'], 'UTF8'));
+        $this->assertTrue(mb_check_encoding(['abc'], 'ASCII'));
+        $this->assertTrue(mb_check_encoding(["\xE9"], 'Windows-1252'));
+
+        $this->assertTrue(mb_check_encoding(['aςσb', 'abc'], 'UTF8'));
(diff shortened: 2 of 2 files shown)


alexandre-daubois:
Caution, [the doc](https://www.php.net/manual/en/function.mb-check-encoding.php) tells that values passed as array are validated recursively 🙂

jeremyFreeAgent:
> Caution, [the doc](https://www.php.net/manual/en/function.mb-check-encoding.php) tells that values passed as array are validated recursively 🙂

Thanks!

derrabus:
Let's kill PHP 7.1 then, shall we? #428

src/Mbstring/Mbstring.php
```diff
@@ -413,6 +413,19 @@ public static function mb_check_encoding($var = null, $encoding = null)
             $encoding = self::$internalEncoding;
         }
 
+        if (is_array($var)) {
```
nicolas-grekas, line 416: let's use an early exit:
`if (!\is_array()) ... return;`

tests/Mbstring/MbstringTest.php
```diff
@@ -466,6 +466,38 @@ public function testCheckEncoding()
         $this->assertTrue(mb_check_encoding("\xE9", 'Windows-1252'));
     }
 
+    /**
+     * @covers \Symfony\Polyfill\Mbstring\Mbstring::mb_check_encoding
+     */
+    public function testCheckEncodingWithArrayValue()
+    {
+        if (70200 <= \PHP_VERSION_ID) {
```
nicolas-grekas, line 474: should be replaced with `@requires PHP 7.2`

stof:
For PHP 7.1, we need to emulate the PHP 7.1 behavior of mbstring, which is not supporting arrays (as we cannot make a project using PHP 7.1 with native mbstring support arrays by installing the polyfill as it would not replace the native functions).
That's totally possible to emulate (triggering the warning and returning `null`)

jeremyFreeAgent:
Thank you @stof and @nicolas-grekas for the PHP < 7.2 behaviour.

src/Mbstring/Mbstring.php
```diff
@@ -406,14 +406,34 @@ public static function mb_encoding_aliases($encoding)
 
     public static function mb_check_encoding($var = null, $encoding = null)
     {
+        if (PHP_VERSION_ID < 70200 && \is_array($var)) {
+            trigger_error('mb_check_encoding() expects parameter 1 to be string, array given');
```
nicolas-grekas, line 410: `, \E_USER_WARNING` as second arg (by default it's a notice)

tests/Mbstring/MbstringTest.php
```diff
@@ -466,6 +466,53 @@ public function testCheckEncoding()
...
+    /**
+     * @covers \Symfony\Polyfill\Mbstring\Mbstring::mb_check_encoding
+     *
+     * @requires PHP < 7.2
+     */
+    public function testCheckEncodingWithArrayValueForPhpLessThan72()
+    {
+        $errorMessage = null;
+        set_error_handler(function ($errno, $errstr, $errfile, $errline) use (&$errorMessage) { $errorMessage = $errstr; });
+        mb_check_encoding(['aςσb'], 'UTF8');
```
nicolas-grekas, line 510: this->assertNull

tests/Mbstring/MbstringTest.php
```diff
@@ -466,6 +466,53 @@ public function testCheckEncoding()
...
+
+    /**
+     * @covers \Symfony\Polyfill\Mbstring\Mbstring::mb_check_encoding
+     *
+     * @requires PHP < 7.2
+     */
+    public function testCheckEncodingWithArrayValueForPhpLessThan72()
+    {
+        $errorMessage = null;
+        set_error_handler(function ($errno, $errstr, $errfile, $errline) use (&$errorMessage) { $errorMessage = $errstr; });
```
nicolas-grekas, line 509: `set_error_handler(function ($type, $msg, $file, $line) use (&$errorMessage) { $errorMessage = \E_USER_WARNING === $type ? $msg : null; });

src/Mbstring/Mbstring.php
jeremyFreeAgent: Thanks, I saw it and fixed it!

tests/Mbstring/MbstringTest.php
jeremyFreeAgent: Good catch!

tests/Mbstring/MbstringTest.php
jeremyFreeAgent: Thanks

nicolas-grekas (approved):

nicolas-grekas:
Thank you @jeremyFreeAgent.

Issue #173, opened 2019-05-01
MB_CASE_FOLD doesn't handle ẞ character properly


colinodell:
## Summary

It seems that this polyfill incorrectly case-folds the `ẞ` character to `ß` instead of `ss`.

## Details

Running this code on PHP 7.3 with `mbstring` installed:

```php
echo mb_convert_case('ẞ', MB_CASE_FOLD);
```

Will produce this output: `ss`

However, when I run that same input through this polyfill (master branch):

```php
echo \Symfony\Polyfill\Mbstring\Mbstring::mb_convert_case('ẞ', \Symfony\Polyfill\Mbstring\Mbstring::MB_CASE_FOLD);
```

I instead get `ß`.  This does not match `mbstring`'s behavior on PHP 7.3.

colinodell:
It looks like this occurs with both `U+1E9E` LATIN CAPITAL LETTER SHARP S and `U+00DF` LATIN SMALL LETTER SHARP S.  Both of these should return `ss` when case-folded.

nicolas-grekas:
PHP 7.3 uses full case folding.
This is implemented in tchwork/utf8, so we could borrow [the map](https://github.com/tchwork/utf8/blob/master/src/Patchwork/Utf8/data/caseFolding_full.ser) from there:
https://github.com/tchwork/utf8/blob/30ec6451aec7d2536f0af8fe535f70c764f2c47a/src/Patchwork/Utf8.php#L168-L170

Up for a PR?

PR #427, opened 2023-01-28
Add case folding map

alexandre-daubois:
Fixes https://github.com/symfony/polyfill/issues/173

@nicolas-grekas, I've indeed borrowed the map you pointed at in the issue. Thanks for this!


src/Mbstring/Resources/unidata/caseFolding.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+return [
```
OskarStark, line 3: Shall we add some credits for tchwork/utf8's work?

alexandre-daubois: Definitely! Not sure which form should it take however. Maybe:

```php
// This map comes from tchwork/utf8 repository (https://github.com/tchwork/utf8).
```

?

src/Mbstring/Mbstring.php
```diff
@@ -299,11 +299,13 @@ public static function mb_convert_case($s, $mode, $encoding = null)
                     $upper = self::getData('upperCase');
                 }
                 $map = $upper;
-            } else {
-                if (self::MB_CASE_FOLD === $mode) {
-                    $s = str_replace(self::CASE_FOLD[0], self::CASE_FOLD[1], $s);
```
nicolas-grekas, line 304: this const is not used anymore, it can be removed

src/Mbstring/Resources/unidata/caseFolding.php
```diff
@@ -0,0 +1,108 @@
...
+    'ß' => 'ss',
```
nicolas-grekas, line 4: this map should be merged with self::CASE_FOLD

nicolas-grekas (commented):
The merged map is available in https://github.com/symfony/string/blob/6.2/AbstractUnicodeString.php (and it looks like there are some duplicates in there? to be fixed also)

src/Mbstring/Resources/unidata/caseFolding.php
nicolas-grekas: no need for credits about this, the map is just derived unicode data

nicolas-grekas:
Please have a look at failing tests, they might be related.

alexandre-daubois:
Updated the map 👍 Removed duplicates also in Symfony: https://github.com/symfony/symfony/pull/49159, thanks for pointing this out!

I'm checking the tests

src/Mbstring/Mbstring.php
```diff
@@ -301,7 +301,8 @@ public static function mb_convert_case($s, $mode, $encoding = null)
                 $map = $upper;
             } else {
                 if (self::MB_CASE_FOLD === $mode) {
-                    $s = str_replace(self::CASE_FOLD[0], self::CASE_FOLD[1], $s);
+                    $caseFolding = self::getData('caseFolding');
+                    $s = str_replace(\array_keys($caseFolding), \array_values($caseFolding), $s);
```
nicolas-grekas, line 305: This is sub-optimal: the map shouldn't require any processing to be used.
Please adjust the map so that this line can be written as:
`$s = str_replace($caseFolding[0], $caseFolding[1], $s);`

nicolas-grekas (commented):
Test failures were related, I fixed them then force-pushed, please fetch before the next iteration :)

src/Mbstring/Mbstring.php
nicolas-grekas: alternatively, we could try using strtr btw
if you can bench which variant would be fastest, that'd be great!

alexandre-daubois: Thank you for your suggestions! Here the benchmark code:

```php
$clock = new MonotonicClock();

$caseFolding = require 'map.php';
$keys = \array_keys($caseFolding);
$values = \array_values($caseFolding);

$now = $clock->now();
$s = str_replace($keys, $values, 'HI ßΐᾬﬗ');
$time = $clock->now()->diff($now);

$output->writeln($s.' str_replace µs :');
dump($time->f);

$now = $clock->now();
$s = strtr('HI ßΐᾬﬗ', $caseFolding);
$time = $clock->now()->diff($now);

$output->writeln($s.' strtr µs :');
dump($time->f);
```

Here are the results of a few executions:

| Attempt  | `str_replace` (µs) | `strtr` (µs) |
| --------- | ------------- | ------ |
| 1  | 5.8E-5  | 4.0E-6 |
| 2  | 5.5E-5  | 4.0E-6 |
| 3  | 1.1E-5  | 4.0E-6 |
| 4  | 5.4E-5  | 3.0E-6 |
| 5  | 1.1E-5  | 4.0E-6 |

Because of these results, I suggest to use `strtr`. I updated the PR accordingly! 👍

OskarStark (approved):

nicolas-grekas:
Thank you @alexandre-daubois.

PR #428, opened 2023-01-28
Drop support for PHP 7.1

derrabus:
With the sunset of Symfony 4, we don't need to support PHP 7.1 anymore.

This PR proposes to bump all polyfill packages to PHP 7.2. This also applies to `symfony/polyfill-php72` which basically becomes an empty package now and can be removed from the subtree split after its next release.


nicolas-grekas:
:thinking: I don't think we need to do this. It's expected for polyfills to support old versions to me. + looking at the diff, this look unneeded, from a technical pov.

Instead, we could decide that new polyfills shouldn't support PHP 7. That'd allow making them more accurate, thanks to better type declarations.

derrabus:
Sure, we don't need to. I just thought, it would make #426 and possibly other future contributions easier.

derrabus:
All right, let's not.

PR #433, opened 2023-06-29
Add odbc_connection_string_*() functions

derrabus:
This PR adds ODBC functions added in php/php-src#8307.

### DIFF
--- a/src/Php82/Php82.php
+++ b/src/Php82/Php82.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Php82;
+
+/**
+ * @author Alexander M. Turek <me@derrabus.de>
+ *
+ * @internal
+ */
+class Php82
+{
+    /**
+     * Determines if a string matches the ODBC quoting rules.
+     *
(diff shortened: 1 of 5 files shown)


src/Php82/bootstrap.php
```diff
@@ -9,6 +9,20 @@
  * file that was distributed with this source code.
  */
 
+use Symfony\Polyfill\Php82 as p;
+
 if (\PHP_VERSION_ID >= 80200) {
     return;
 }
+
+if (!function_exists('odbc_connection_string_is_quoted')) {
```
IonBazan, line 18: I think this should only be loaded if `odbc` extension is present.

stof: agreed

IonBazan (approved):

IonBazan:
Might be good to add `@requires extension odbc` annotations to the tests in case someone wants to run the test suite without that extension.

derrabus:
Makes sense, done.

src/Php82/Php82.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+namespace Symfony\Polyfill\Php82;
```
nicolas-grekas, line 3:
```suggestion
/*
 * This file is part of the Symfony package.
 *
 * (c) Fabien Potencier <fabien@symfony.com>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Symfony\Polyfill\Php82;
```

nicolas-grekas:
Thank you @derrabus.

PR #435, opened 2023-07-06
[PHP 8.3] Polyfill mb_str_pad()

IonBazan:
Polyfills the `mb_str_pad()` function added in PHP 8.3: https://wiki.php.net/rfc/mb_str_pad

Test cases were taken from the RFC implementation and adapted to PHPUnit.

### DIFF
--- a/tests/Php83/Php83Test.php
+++ b/tests/Php83/Php83Test.php
@@ -17,12 +17,109 @@ class Php83Test extends TestCase
 {
     /**
      * @covers \Symfony\Polyfill\Php83\Php83::json_validate
+     *
      * @dataProvider jsonDataProvider
      */
     public function testJsonValidate(bool $valid, string $json, string $errorMessage = 'No error', int $depth = 512, int $options = 0)
     {
         $this->assertSame($valid, json_validate($json, $depth, $options));
         $this->assertSame($errorMessage, json_last_error_msg());
+    }
+
+    /**
+     * @covers \Symfony\Polyfill\Php83\Php83::mb_str_pad
+     *
+     * @dataProvider paddingStringProvider
+     * @dataProvider paddingEmojiProvider
+     * @dataProvider paddingEncodingProvider
+     */
+    public function testMbStrPad(string $expectedResult, string $string, int $length, string $padString, int $padType, string $encoding = null): void
+    {
+        $this->assertSame($expectedResult, mb_convert_encoding(mb_str_pad($string, $length, $padString, $padType, $encoding), 'UTF-8', $encoding ?? mb_internal_encoding()));
+    }
+
+    /**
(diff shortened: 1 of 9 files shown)


Ayesh:
I'm sorry if I'm missing something documented elsewhere, but I wanted to ask if this should belong to Mbstring or Php83? Perhapas, before declaring polyfilled `mb_str_pad`, we should see if the `mbstring` exists either as the C extension or in the form of a polyfill by checking `function_exists('mb_substr')`?

IonBazan:
Good question - I wasn't sure because we don't check it for PHP 7.2 polyfill (`mb_chr` and `mb_ord`) but we do check it for PHP 7.4 (`mb_str_split`): https://github.com/symfony/polyfill/blob/05dfdffc0e834c216a677474fa05304db3384710/src/Php74/bootstrap.php#L21-L23

I guess we can add a check for `mb_substr` if needed here.

stof:
the new function should be added in 2 places

- in the php 8.3 polyfill _if other mbstring functions are available_
- in the mbstring polyfill (so that it is available for someone using PHP 8.3 natively without the native mbstring extension)

tests/Php83/Php83Test.php
```diff
@@ -25,6 +26,110 @@ public function testJsonValidate(bool $valid, string $json, string $errorMessage
...
+            $padStr = mb_convert_encoding('▶▶', $encoding, 'UTF-8');
+
+            yield ['Σὲ γνωρίζω ἀπὸ τὴν κόψη Зарегистрируйтесь▶▶▶', $input, 44, $padStr, \STR_PAD_RIGHT, $encoding];
+            yield ['▶▶▶Σὲ γνωρίζω ἀπὸ τὴν κόψη Зарегистрируйтесь', $input, 44, $padStr, \STR_PAD_LEFT, $encoding];
+            yield ['▶Σὲ γνωρίζω ἀπὸ τὴν κόψη Зарегистрируйтесь▶▶', $input, 44, $padStr, \STR_PAD_BOTH, $encoding];
+        }
+    }
+
+    /**
+     * @return iterable<array{0: string, 1: string, 2: int, 3: string, 4: int, 5?: string}>
```
nicolas-grekas, line 122: let's remove all those, they're not useful IMHO

IonBazan:
PHP 7.2 and 7.1 don't seem to handle UTF-32 strings correctly when mbstring is missing 🤔 
@nicolas-grekas Should I skip those tests?

nicolas-grekas:
> PHP 7.2 and 7.1 don't seem to handle UTF-32 strings correctly when mbstring is missing

that's quite possible, let's skip these yes (or fix the polyfill, but nobody uses utf32 so :shrug: )

nicolas-grekas:
Thank you @IonBazan.

PR #411, opened 2022-07-12
[PHP 8.2] Add `ini_parse_quantity` function

Ayesh:
Polyfill taken from [PHP.Watch - PHP 8.2: New `ini_parse_quantity` function](https://php.watch/versions/8.2/ini_parse_quantity#polyfill)

Tries to mimic the native warnings with `E_USER_WARNING` errors, and overflow/underflow conditions by inspecting the int to float type change.

PR #439, opened 2023-08-04
Add polyfill for ini_parse_quantity based on the C code in PHP

fisharebest:
See #411 for a previous attempt to polyfill this function.

The `ini_parse_quantity()` is difficult to polyfill as the error handling is non-trivial, and several invalid inputs are treated as valid.

This polyfill is a simple/direct implementation of the C code from PHP.
As such it uses `goto`s and pointer-based string access.
I am fully aware that this would fail most code reviews :-)

However it works and I hope it will be useful until a cleaner implementation can be provided.

The unit tests provide 100% code coverage.  I have also included (but commented out) a brute-force test which takes many hours to run. I found it was very helpful in finding edge cases.  It may be useful for anyone who tries to refactor the code.

The two overflow tests will probably fail on a 32bit build of PHP, however the code itself should work.
I just haven't been able to work out the expected values.

### DIFF
--- a/src/Php82/Php82.php
+++ b/src/Php82/Php82.php
@@ -13,6 +13,7 @@ namespace Symfony\Polyfill\Php82;
 
 /**
  * @author Alexander M. Turek <me@derrabus.de>
+ * @author Greg Roach <greg@subaqua.co.uk>
  *
  * @internal
  */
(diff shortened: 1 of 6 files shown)


src/Php82/Php82.php
```diff
@@ -11,8 +11,17 @@
 
 namespace Symfony\Polyfill\Php82;
 
+use function base_convert;
+use function sprintf;
+use function strpos;
+use function substr;
+use function trigger_error;
```
stof: this should not be used for 2 reasons:

- Polyfills are intended to be compatible with a broad range of PHP versions
- Symfony does not use `use function` or `use const` right now

src/Php82/Php82.php
```diff
@@ -71,4 +80,304 @@ public static function odbc_connection_string_quote(string $str): string
...
+            ++$digits;
+        } elseif ($digits < $str_end && $value[$digits] === '-') {
+            $is_negative = true;
+            $overflow = true;
+            ++$digits;
+        }
+
+        $digits_end = $digits;
+
+        // The native function treats 0x0x123 the same as 0x123.  This is a bug which we must replicate.
```
stof, line 200: have you reported it ?

tests/Php82/Php82Test.php
```diff
@@ -84,4 +91,363 @@ public static function provideMoreConnectionStringValues(): \Generator
     {
         yield 'double curly at the end' => ['foo}}', false, true, '{foo}}}}}'];
     }
+
+    /**
+     * @covers \Symfony\Polyfill\Php82\Php82::ini_parse_quantity
```
stof: we don't use `@cover`

tests/Php82/Php82Test.php
```diff
@@ -84,4 +91,363 @@ public static function provideMoreConnectionStringValues(): \Generator
...
+     */
+    public function testIniParseQuantity()
+    {
+        $this->assertSame(0, Php82::ini_parse_quantity(''));
```
stof: Tests should use `ini_parse_quantity` (but relying on namespace fallback, not throught a fully qualified class name), not the internal implementation method. This allows the TestListener to execute the test twice, using the native implementation or the polyfill.

tests/Php82/Php82Test.php
```diff
@@ -84,4 +91,363 @@ public static function provideMoreConnectionStringValues(): \Generator
...
+    /**
+     * This is a brute-force test that should only be used during local development.
+     * It takes approximately 4 hours to run.  It is good at picking up edge cases.
+     *
+     * @covers \Symfony\Polyfill\Php82\Php82::ini_parse_quantity
+     */
+    public function testIniParseQuantityUsingBruteForce()
+    {
+        if (PHP_VERSION_ID < 80200) {
+            // This test requires the PHP function as a reference.
```
stof: the tests should rather be marked as skipped than as passing as no-op.

stof (changes requested):

src/Php82/Php82.php
fisharebest: https://github.com/php/php-src/issues/11876

src/Php82/Php82.php
fisharebest: Done!

tests/Php82/Php82Test.php
fisharebest: Done!

tests/Php82/Php82Test.php
fisharebest: Done!

fisharebest:
I've done some more testing, and have one edge-case that I'm having difficulty in resolving.

When using a mutliplier (e.g. 9999999999G), the code checks whether 9999999999 * 1024^3 would cause an overflow.
If it does, it triggers a warning.

But it also does the multiplication anyway!
This overflows (using C arithmetic), and gives a meaningles answer.  e.g.

```
php > echo ini_parse_quantity('9999999999G');

Warning: Invalid quantity "9999999999G": value is out of range, using overflow result for backwards compatibility in php shell code on line 1

Call Stack:
   13.7692     389672   1. {main}() php shell code:0
   13.7696     389672   2. ini_parse_quantity($shorthand = '9999999999G') php shell code:1

-7709325834783293440
```

This is probably architecture-dependent, and I don't know what tools I can use.  e.g. can I have a dependency on the `gmp_` functions?

Should I continue with this PR - dispite this one (IMHO very minor) limitation - and maybe list it as a known bug.
Or do you need a complete 100% implementation before you can accept it?

TimWolla:
> This is probably architecture-dependent, and I don't know what tools I can use. e.g. can I have a dependency on the `gmp_` functions?

No, this is not arch dependent and trivial to resolve for the specific case of multiplying by a power of two. Multiplying by a power of two is equivalent to a left-shift by one. 1024 is 2^10, thus a shift of 10. `1024*1024*1024` is 2^30, thus a shift of 30:

```
var_dump(9999999999 << 30); // int(-7709325834783293440)
```

The result is specific to 64 Bit PHP, of course.

fisharebest:
Thanks @TimWolla - I was over-thinking it.

TimWolla:
Asking as a user: @fisharebest Is there some change pending from you or is this PR complete / ready for review?

fisharebest:
Sorry - this dropped off my radar.  I'll check where it is at later today.

tests/Php82/Php82Test.php
fisharebest: Done

fisharebest:
OK - I'm happy with the code now.  I've made all the requested changes, and it has 100% test coverage.

The CI tests are now all failing (they worked before) - but I don't think that's anything to do with me.  The tests run locally fine.

tests/Php82/Php82Test.php
```diff
@@ -13,6 +13,7 @@
 
 use PHPUnit\Framework\TestCase;
 
+
```
nicolas-grekas: should be removed

nicolas-grekas (commented):
Can you please also update the src/Php82/bootstrap.php file, the src/Php72/README one, and the root CHANGELOG one?

tests/Php82/Php82Test.php
fisharebest: done

fisharebest:
> Can you please also update the src/Php82/bootstrap.php file, the src/Php72/README one, and the root CHANGELOG one?

Done.

fisharebest:
Now that the CI tests are working, it looks like on PHP 7.1 - 7.2, the function is returning a `string` rather than an `int`.

Give me a few minutes...

src/Php82/bootstrap.php
```diff
@@ -30,3 +30,7 @@ function odbc_connection_string_should_quote(string $str): bool { return p\Php82
 if (!function_exists('odbc_connection_string_quote')) {
     function odbc_connection_string_quote(string $str): string { return p\Php82::odbc_connection_string_quote($str); }
 }
+
+if (!function_exists('ini_parse_quantity')) {
+    function ini_parse_quantity(string $shorthand): string { return p\Php82::ini_parse_quantity($shorthand); }
```
stof: the return type looks suspicious to me here

stof (changes requested):

src/Php82/bootstrap.php
fisharebest: Dang.  Copy/paste error.

fisharebest: Sorry for that.  I'm rushing as I'm likely to be unavailable for most of the next week.  Fixed.

src/Php82/Php82.php
```diff
@@ -71,4 +72,312 @@ public static function odbc_connection_string_quote(string $str): string
...
+                'Invalid quantity "%s": value is out of range, using overflow result for backwards compatibility',
+                self::escapeString($value)
+            );
+            trigger_error($message, E_USER_WARNING);
+        }
+
+        return $retval;
+    }
+
+    /*
```
TimWolla:
```suggestion
    /**
```

fisharebest: done

nicolas-grekas:
Thank you @fisharebest.

PR #440, opened 2023-08-14
[PHP 8.3] Add Exception/Error classes for `Date/Time` extension

Ayesh:
Adds nine classes for the new exceptions/errors declared in ext-date.

 - [RFC: PHP RFC: More Appropriate Date/Time Exceptions](https://wiki.php.net/rfc/datetime-exceptions)
 - [PHP.Watch: DateTime Exception Class Polyfill](https://php.watch/versions/8.3/datetime-exceptions#polyfill)

### DIFF
--- a/src/Php83/Resources/stubs/DateError.php
+++ b/src/Php83/Resources/stubs/DateError.php
@@ -1,0 +1,16 @@
+<?php
+
+/*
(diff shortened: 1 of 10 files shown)


nicolas-grekas:
Thank you @Ayesh.

PR #442, opened 2023-08-28
Fix the `ini_parse_quantity()` polyfill

TimWolla:
It previously was only defined (and tested) of the odbc extension is available.


stof (approved):

derrabus (approved):

nicolas-grekas:
Thank you @TimWolla.

PR #446, opened 2023-09-06
Use mb_* polyfill implementations in other functions

buismaarten:


derrabus:
Can you please elaborate the motivation behind this change?

buismaarten:
This library is a polyfill for mb_* functions, I assume we can't use the native functions when someone doesn't have the native extension enabled. @derrabus

stof:
This library polyfills the functions. So those functions will be available (provided by the polyfill)

stof:
I'm closing this as this PR does not solve any actual problem. The change was based on assumptions that were wrong.

PR #441, opened 2023-08-23
Fix the CI setup

stof:
As the default branch of the repository has been renamed, the COMPOSER_ROOT_VERSION cannot be set to dev-main relying on the branch alias as there is no such branch alias anymore.

PR #447, opened 2023-09-14
Fix the appveyor CI setup

stof:
this is the same fix than https://github.com/symfony/polyfill/pull/441 but for appveyor as the previous PR handled only the GitHub Actions config.


nicolas-grekas:
Thank you @stof.

PR #449, opened 2023-09-28
Fix return type for `Locale::parseLocale` - it can return `null`

andrew-demb:
https://www.php.net/manual/en/locale.parselocale.php

> Returns null when the length of locale exceeds INTL_MAX_LOCALE_LEN.

----

> Recreated PR from https://github.com/symfony/polyfill-intl-icu/pull/1


andrew-demb:
CI failure is unrelated

derrabus (approved):

nicolas-grekas:
Thank you @andrew-demb.

PR #454, opened 2023-11-06
Conform to IDNA version 15.1.0 revision 31

TRowbotham:
This adds the necessary changes to conform to [IDNA version 15.1.0 revision 31](https://www.unicode.org/reports/tr46/tr46-31.html#Modifications).

Notable Changes

* Transitional processing (the default in PHP) is now deprecated. No deprecation notices were added as PHP does not yet report a deprecation notice in this case.
* An error is no longer recorded for characters with a status of disallowed.
* When performing code point mapping and transitional processing is enabled the code point U+1E9E capital sharp s (ẞ), is replaced by the string “ss”
* A new internal option "IgnoreInvalidPunycode" was added, which is supposed to allow for an all-ASCII fast-path, however, the official tests do not test this configuration option.

### DIFF
--- a/src/Intl/Idn/Idn.php
+++ b/src/Intl/Idn/Idn.php
@@ -280,10 +280,6 @@ final class Idn
 
             switch ($data['status']) {
                 case 'disallowed':
-                    $info->errors |= self::ERROR_DISALLOWED;
-
-                    // no break.
-
                 case 'valid':
                     $str .= mb_chr($codePoint, 'utf-8');
 
@@ -294,7 +290,7 @@ final class Idn
                     break;
 
                 case 'mapped':
-                    $str .= $data['mapping'];
+                    $str .= $transitional && 0x1E9E === $codePoint ? 'ss' : $data['mapping'];
 
                     break;
 
@@ -346,6 +342,18 @@ final class Idn
             $validationOptions = $options;
 
             if ('xn--' === substr($label, 0, 4)) {
+                // Step 4.1. If the label contains any non-ASCII code point (i.e., a code point greater than U+007F),
+                // record that there was an error, and continue with the next label.
+                if (preg_match('/[^\x00-\x7F]/', $label)) {
+                    $info->errors |= self::ERROR_PUNYCODE;
+
+                    continue;
+                }
+
+                // Step 4.2. Attempt to convert the rest of the label to Unicode according to Punycode [RFC3492]. If
+                // that conversion fails, record that there was an error, and continue
+                // with the next label. Otherwise replace the original label in the string by the results of the
+                // conversion.
                 try {
                     $label = self::punycodeDecode(substr($label, 4));
                 } catch (\Exception $e) {
@@ -516,6 +524,8 @@ final class Idn
             if ('-' === substr($label, -1, 1)) {
                 $info->errors |= self::ERROR_TRAILING_HYPHEN;
             }
+        } elseif ('xn--' === substr($label, 0, 4)) {
+            $info->errors |= self::ERROR_PUNYCODE;
         }
 
         // Step 4. The label must not contain a U+002E (.) FULL STOP.
--- a/tests/Intl/Idn/IdnTest.php
+++ b/tests/Intl/Idn/IdnTest.php
@@ -284,6 +284,55 @@ class IdnTest extends TestCase
         $this->assertSame($encoded, $result);
     }
 
+    /**
+     * IDNA 15.1.0 revision 31
+     *
+     * This tests the change in "Section 4 Processing step 1. Map" which conditionally maps U+1E9E capital sharp s to
+     * "ss" if Transitional_Processing is used.
+     *
+     * @dataProvider captialSharpSProvider
+     */
+    public function testCapitalSharpSProcessing($input, $expected, $flags)
+    {
+        idn_to_utf8($input, $flags, \INTL_IDNA_VARIANT_UTS46, $info);
+        $this->assertSame($expected, $info['result']);
+    }
+
+    /**
+     * IDNA 15.1.0 revision 31
+     *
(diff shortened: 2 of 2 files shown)


derrabus:
Can we add a test?

stof:
I don't think we should add options that don't exist in PHP. This would defeat the purpose of this library.

TRowbotham:
I have decided to remove the "IgnoreInvalidPunycode" option as suggested since the option is not available in PHP.

src/Intl/Idn/Idn.php
```diff
@@ -346,6 +342,18 @@ private static function process($domain, array $options, Info $info)
...
+                // Step 4.1. If the label contains any non-ASCII code point (i.e., a code point greater than U+007F),
+                // record that there was an error, and continue with the next label.
+                if (1 === preg_match('/[^\x00-\x7F]/', $label)) {
+                    $info->errors |= self::ERROR_PUNYCODE;
+
+                    continue;
+                }
+
+                // Step 4.2. Attempt to convert the rest of the label to Unicode according to Punycode [RFC3492]. If
+                // that conversion fails and if not IgnoreInvalidPunycode, record that there was an error, and continue
```
stof: The reference to `IgnoreInvalidPunycode` should be removed

TRowbotham: Thanks for catching that

src/Intl/Idn/Idn.php
```diff
@@ -294,7 +290,7 @@ private static function mapCodePoints($input, array $options, Info $info)
                     break;
 
                 case 'mapped':
-                    $str .= $data['mapping'];
+                    $str .= ($transitional && $codePoint === 0x1E9E ? 'ss' : $data['mapping']);
```
nicolas-grekas:
```suggestion
                    $str .= $transitional && 0x1E9E === $codePoint ? 'ss' : $data['mapping'];
```

src/Intl/Idn/Idn.php
```diff
@@ -346,6 +342,18 @@ private static function process($domain, array $options, Info $info)
             $validationOptions = $options;
 
             if ('xn--' === substr($label, 0, 4)) {
+                // Step 4.1. If the label contains any non-ASCII code point (i.e., a code point greater than U+007F),
+                // record that there was an error, and continue with the next label.
+                if (1 === preg_match('/[^\x00-\x7F]/', $label)) {
```
nicolas-grekas:
```suggestion
                if (preg_match('/[^\x00-\x7F]/', $label)) {
```

src/Intl/Idn/Idn.php
```diff
@@ -516,6 +524,8 @@ private static function validateLabel($label, Info $info, array $options, $canBe
             if ('-' === substr($label, -1, 1)) {
                 $info->errors |= self::ERROR_TRAILING_HYPHEN;
             }
+        } elseif (0 === strncmp($label, 'xn--', 4)) {
```
nicolas-grekas: `'xn--' === substr($label, 0, 4)` to match the same check above (or replace the code above maybe?)

nicolas-grekas (commented):
Thanks for the PR.
Can you please add some tests?

src/Intl/Idn/Idn.php
```diff
@@ -280,10 +280,6 @@ private static function mapCodePoints($input, array $options, Info $info)
 
             switch ($data['status']) {
                 case 'disallowed':
-                    $info->errors |= self::ERROR_DISALLOWED;
```
nicolas-grekas, line 283: Can you explain why this change?

nicolas-grekas (commented):
I fixed my CS comments by force-pushing on your fork.
Tests would still be welcome. And I have one question (see inline comments).
Are you still interested in finishing this PR?

src/Intl/Idn/Idn.php
TRowbotham: The reasoning for this change according to the spec is "Checking for disallowed characters both before and after normalization yielded inconsistent results for unnormalized vs. normalized text".

TRowbotham:
> I fixed my CS comments by force-pushing on your fork. Tests would still be welcome. And I have one question (see inline comments). Are you still interested in finishing this PR?

Yes, sorry, haven't had a lot of free time lately.  I'll try to put some tests together this week.

Thanks for taking care of the style issues.

nicolas-grekas:
Thank you @TRowbotham.

nicolas-grekas:
@TRowbotham we have the CI fail because IDN tests don't work anymore. Could you please have a look?

nicolas-grekas:
`./phpunit --filter Idn` BTW

TRowbotham:
I'll try to take a look at this over the next couple of weeks. Since it doesn't look like anything has changed in the polyfill, my first guess would be that the the intl extension that PHP was built with in the test env is now using Unicode 16, which has a number of changes to it on the IDN side of things. Thanks for reaching out!

xabbuh:
@TRowbotham I think it's quite likely that the intl extension version has changed as we also see other intl related failures (or have seen in the past) around date formats.

TRowbotham:
I had some time to dig into this today, and the version of ICU being used by the intl extension seems to have changed in the test environment. This coupled with the test data being meant for Unicode version 13 means that the data and implementation are too far out of sync. Updating the unicode data and test data will require code changes to the IDN polyfill, which I am happy to make, but a quick fix would be pinning the ICU version for the intl extension. The last working ICU version was 73.2.

https://github.com/symfony/polyfill/blob/80b0d3fac4768e5987a2c39e34815efa60fd9662/.github/workflows/tests.yml#L37
```diff
- extensions: "apcu, intl, mbstring, odbc, uuid"
+ extensions: "apcu, intl-73.2, mbstring, odbc, uuid"
```

nicolas-grekas:
Thanks for having a look. I'm fine with pinning, and upgrading separately.

PR #457, opened 2024-01-09
Fix ini_parse_quantity()

### LINKED ISSUE #448: Backport fix for ini_parse_quantity from PHP
Tests are currently failing, likely related to https://github.com/php/php-src/pull/11910  and https://github.com/php/php-src/issues/11876

@fisharebest would you like to have a look and submit a fix?

nicolas-grekas:
Fix #448

/cc @fisharebest could you please have a look?
@TimWolla also maybe?

### DIFF
--- a/src/Php82/Php82.php
+++ b/src/Php82/Php82.php
@@ -174,6 +174,42 @@ class Php82
 
                 return 0;
             }
+
+            $digits_consumed = $digits;
+            /* Ignore leading whitespace. */
+            while ($digits_consumed < $str_end && false !== strpos($ctype_space, $value[$digits_consumed])) {
+                ++$digits_consumed;
+            }
+            if ($digits_consumed !== $str_end && ($value[$digits_consumed] === '+' || $value[$digits_consumed] === '-')) {
+                ++$digits_consumed;
+            }
+
+            if ($value[$digits_consumed] === '0') {
+                /* Value is just 0 */
+                if ($digits_consumed + 1 === $str_end) {
+                    goto evaluation;
+                }
+                switch ($value[$digits_consumed + 1]) {
+                    case 'x':
+                    case 'X':
+                    case 'o':
+                    case 'O':
+                    case 'b':
+                    case 'B':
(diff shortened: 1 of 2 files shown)


src/Php82/Php82.php
```diff
@@ -174,6 +174,42 @@ public static function ini_parse_quantity(string $value): int
...
+                ++$digits_consumed;
+            }
+            if ($digits_consumed !== $str_end && ($value[$digits_consumed] === '+' || $value[$digits_consumed] === '-')) {
+                ++$digits_consumed;
+            }
+
+            if ($value[$digits_consumed] === '0') {
+                /* Value is just 0 */
+                if ($digits_consumed + 1 === $str_end) {
+                    return 0;
```
TimWolla: I believe `goto evaluation;` or actually adding the same helper function `zend_ini_consume_quantity_prefix` would match the “style” of the original implementation better.

But trying to understand the code, I realize that the original implementation is needlessly complicated / still allows for funky stuff like `0x+0` and `0x     0`. /cc @girgias

Girgias: I don't think it allows a second + or whitespace, this is why we needed to use two loops to handle the prefix and suffix shenanigans. But I might have implemented this wrong, I do know there are some weird cases that are still possible tho :-/

fisharebest:
@nicolas-grekas - the logic seems sound, and it passes my "brute-force" test, so it looks good to me.

src/Php82/Php82.php
nicolas-grekas: > goto evaluation

updated!

src/Php82/Php82.php
```diff
@@ -197,16 +233,6 @@ public static function ini_parse_quantity(string $value): int
 
         $digits_end = $digits;
 
-        // The native function treats 0x0x123 the same as 0x123.  This is a bug which we must replicate.
```
nicolas-grekas, line 200: I just removed this block /cc @fisharebest 
Nothing else to cleanup AFAYK?

Issue #459, opened 2024-01-10
Intl Collator types are (slightly) incoherent


marien-probesys:
After installing [the Form component](https://github.com/symfony/form) which requires the `symfony/polyfill-intl-icu` package, I noticed that PHPStan started to complain about one of my method:

```php
public function localeCompare(string $field1, string $field2): int
{
    $collator = new \Collator('fr_FR');
    $fieldsComparison = $collator->compare($field1, $field2);

    if ($fieldsComparison === false) {
        $fieldsComparison = strcmp($field1, $field2);
    }

    return $fieldsComparison;
}
```

The PHPStan output:

```
 ------ ----------------------------------------------------------------------- 
  Line   src/Service/Sorter/LocaleSorter.php                                    
 ------ ----------------------------------------------------------------------- 
  51     Method App\Service\Sorter\LocaleSorter::localeCompare() should return  
         int but returns int|true.                                              
 ------ -----------------------------------------------------------------------
```

It worked well before because [`\Collator::compare` returns `int|false`](https://www.php.net/manual/en/class.collator.php).

However, the Collator class from this repo says that [it returns `int|bool`](https://github.com/symfony/polyfill/blob/1.x/src/Intl/Icu/Collator.php#L123).

I'm a bit surprised that Composer installed the `symfony/polyfill-intl-icu` package in first place. I have double-checked, and the PHP `intl` extension is correctly installed. But I don't know how Composer and polyfill work, so it may be normal :)

My workaround for now is to hardcode the type:

```php
/** @var int|false */
$fieldsComparison = $collator->compare($field1, $field2);
```

I noticed more incoherent methods/functions: `__construct`, `create`, `getAttribute`, `getErrorCode`, `getErrorMessage`, `getLocale`, `getSortKey` and `getStrength`.

stof:
This looks like a mistake in the phpdoc in this repo. Can you send a PR fixing it ?

marien-probesys:
Yep, I can totally do that :+1:

PR #460, opened 2024-01-10
Fix the Collator return types in phpdoc

marien-probesys:
The types were incoherent with the PHP documentation. There are still some issues with types (e.g. constructor should not accept `null` value). But changing them would break the API so I thought it wasn't a good idea to fix them yet.

Reference: https://www.php.net/manual/class.collator.php

Contribute to https://github.com/symfony/polyfill/issues/459

### DIFF
--- a/src/Intl/Icu/Collator.php
+++ b/src/Intl/Icu/Collator.php
@@ -120,7 +120,7 @@ abstract class Collator
     /**
      * Not supported. Compare two Unicode strings.
      *
-     * @return bool|int
+     * @return int|false
      *
      * @see https://php.net/collator.compare
      *
@@ -134,7 +134,7 @@ abstract class Collator
     /**
      * Not supported. Get a value of an integer collator attribute.
      *
-     * @return bool|int The attribute value on success or false on error
+     * @return int|false The attribute value on success or false on error
      *
      * @see https://php.net/collator.getattribute
      *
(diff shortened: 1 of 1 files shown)


src/Intl/Icu/Collator.php
```diff
@@ -193,7 +193,7 @@ public function getSortKey(string $string)
     /**
      * Not supported. Get current collator's strength.
      *
-     * @return bool|int The current collator's strength or false on failure
+     * @return int The current collator's strength or false on failure
```
marien-probesys, line 196: I didn't change the comment because the PHP documentation itself is incoherent (the return type is `int`, but the doc mentions a `false` value), see https://www.php.net/manual/en/collator.getstrength.php

stof: could it be that *older* PHP versions can return false for failures while new versions throw exceptions ?

stof:
> e.g. constructor should not accept `null` value

We must accept `null` due to the way internal arguments for scalars are all accepting `null` even when not marked as nullable (see the deprecation introduced in PHP 8.1), while userland types were always stricter in that regard.

stof (approved):

src/Intl/Icu/Collator.php
marien-probesys: I think it is! I think the relevant change is https://github.com/php/php-src/commit/1b93cfee0ca23d95893fa6ecd53d0a0c02aeb7bf#diff-22c8fd8eefcbc1be5e379b048ce5726a61b34dc4352d7d2568e76c56095f2c25 (PHP 8)

nicolas-grekas:
Thank you @marien-probesys.

Issue #458, opened 2024-01-10
[PHP 8.4] `mb_trim`, `mb_ltrim`, and `mb_rtrim` polyfills


Ayesh:
RFC: [Multibyte for trim function mb_trim, mb_ltrim and mb_rtrim](https://wiki.php.net/rfc/mb_trim)
php.net: _TBD_
PHP.Watch: [PHP 8.4: Mbstring: New `mb_trim`, `mb_ltrim`, and `mb_rtrim` functions](https://php.watch/versions/8.4/mb_trim-mb_ltrim-mb_rtrim)

I think we can provide a polyfill for these functions. I created a small repo [here](https://github.com/PHPWatch/mb_trim-polyfill/blob/main/src/mb_trim.php) with sample code that passes the test suite in `php-src` for this function. I'll happily submit a PR with it if you think that approach is good enough.

Thank you.

stof:
PRs are happily accepted.

This should be added in 2 places:
- in the mbstring polyfill
- in the php 8.4 polyfill (which needs to be created as it does not exist yet), but only when mbstring is available (see how `mb_str_split` is handled in the PHP 7.4 polyfill)

PR #462, opened 2024-01-10
[PHP 8.4] Add initial stubs

Ayesh:
- Adds PHP 8.4 polyfill for splits
 - Enables PHP 8.4 in GitHub Actions tests

This is also to prepare for #458 to add new `mb_trim`, `mb_ltrim`, and `mb_rtrim` functions, which will be a separate PR.

### DIFF
--- a/src/Php84/LICENSE
+++ b/src/Php84/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2024-present Fabien Potencier
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is furnished
+to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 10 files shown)


src/Php84/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2022-present Fabien Potencier
```
OskarStark:
```suggestion
Copyright (c) 2024-present Fabien Potencier
```
⌚

src/Php84/composer.json
```diff
@@ -0,0 +1,35 @@
...
+            "name": "Nicolas Grekas",
+            "email": "p@tchwork.com"
+        },
+        {
+            "name": "Symfony Community",
+            "homepage": "https://symfony.com/contributors"
+        }
+    ],
+    "require": {
+        "php": ">=7.1"
```
OskarStark, line 19: Is this correct? just asking

src/Php84/LICENSE
Ayesh: Joys of copy-pasta 🥲. Thanks for the good catch.

src/Php84/composer.json
Ayesh: PHP 8.3 polyfill has the same constraint. I don't immediately see why we have to raise the minimal PHP version, but we can increase this if we have to when we add the individual polyfill classes/functions.

stof (approved):

nicolas-grekas:
Thank you @Ayesh.

PR #464, opened 2024-02-21
Add translit for EUR symbol

louismariegaborit:
Add translit for € to have same result.

Today, we have this : 

```
\Symfony\Polyfill\Iconv\Iconv::iconv('UTF-8', 'ASCII//TRANSLIT', '€');
Result : false

iconv('UTF-8', 'ASCII//TRANSLIT', '€');
Result : EUR
```

With PR change, we have : 
```
\Symfony\Polyfill\Iconv\Iconv::iconv('UTF-8', 'ASCII//TRANSLIT', '€');
Result : EUR

iconv('UTF-8', 'ASCII//TRANSLIT', '€');
Result : EUR
```


nicolas-grekas:
We've got the current list from https://github.com/unicode-org/cldr/blob/main/common/transforms/Latin-ASCII.xml
Our approach to defining these lists is to find a good source that we don't have to maintain.
It'd be interesting to see which implementation of iconv has the € => EUR mapping and see if we could leverage it.
There might be more characters to map.
(and we don't want to maintain our own custom map).
Could you please have a look?

louismariegaborit:
> We've got the current list from https://github.com/unicode-org/cldr/blob/main/common/transforms/Latin-ASCII.xml Our approach to defining these lists is to find a good source that we don't have to maintain. It'd be interesting to see which implementation of iconv has the € => EUR mapping and see if we could leverage it. There might be more characters to map. (and we don't want to maintain our own custom map). Could you please have a look?

It's the conversion that debian do.  
```
docker run --rm -it php:8.2-fpm-bullseye php -r "echo iconv('UTF-8', 'ASCII//TRANSLIT', '€')
EUR
```  
The cldr repository have a specific section on currency symbol.  
I understand that you can't maintain this specific list and my change is not good.  
How we can support this character in the polyfill ?

nicolas-grekas:
We should try to understand how debian does the conversion. Where it gets the mapping from. Might be from cldr, or something else. Once we have the source, we can borrow from there.

nicolas-grekas:
libiconv uses https://github.com/roboticslibrary/libiconv/blob/master/lib/translit.def to do its transliterations, we might want to borrow from it. This file contains more that ASCII translit so I'm not sure exactly how we should leverage it, but that's a start.

nicolas-grekas:
I had a quick look at using translit.def and it looks like something else is done since eg with this map `É` is mapped to `´E` while testing locally with the iconv function gives `E`.
Closing this as we won't merge a specific code point. We'd merge a generic map that gets closer to an accurate polyfilling of course.

PR #466, opened 2024-03-09
[mbstring][PHP 8.4] Add `mb_ucfirst` and `mb_lcfirst` to polyfills

Ayesh:
Adds polyfills for `mb_ucfirst` and `mb_lcfirst` functions based on the polyfill shown in PHP.Watch. It basically takes the first mb character, calls `mb_convert_case` with `MB_CASE_TITLE` or `MB_CASE_LOWER`, and returns with the concat of the remaining string.

The tests are taken from the php-src PR.

 - [RFC](https://wiki.php.net/rfc/mb_ucfirst)
 - [php-src PR: php-src#13161](https://github.com/php/php-src/pull/13161)
 - [PHP.Watch - PHP 8.4: New `mb_ucfirst` and `mb_lcfirst` functions](https://php.watch/versions/8.4/mb_ucfirst-mb_ucfirst)

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -48,6 +48,8 @@ namespace Symfony\Polyfill\Mbstring;
  * - mb_strstr               - Finds first occurrence of a string within another
  * - mb_strwidth             - Return width of string
  * - mb_substr_count         - Count the number of substring occurrences
+ * - mb_ucfirst              - Make a string's first character uppercase
+ * - mb_lcfirst              - Make a string's first character lowercase
  *
  * Not implemented:
  * - mb_convert_kana         - Convert "kana" one from another ("zen-kaku", "han-kaku" and more)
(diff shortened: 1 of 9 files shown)


src/Mbstring/Mbstring.php
```diff
@@ -871,6 +871,49 @@ public static function mb_str_pad(string $string, int $length, string $pad_strin
         }
     }
 
+    public static function mb_ucfirst(string $string, ?string $encoding = null): string {
```
zonuexe: It is good to add line breaks according to [Symfony Coding Standards](https://symfony.com/doc/current/contributing/code/standards.html).

```suggestion
    public static function mb_ucfirst(string $string, ?string $encoding = null): string
    {
```

zonuexe: It seems good to add `mb_ucfirst` and `mb_ucfirst` to Implemented in the class header.

https://github.com/symfony/polyfill/pull/471/files#diff-3e84767d56f3132c8784f8423c914322116cef95a35b583c80070f666e9360d7R51

src/Mbstring/Mbstring.php
```diff
@@ -871,6 +871,49 @@ public static function mb_str_pad(string $string, int $length, string $pad_strin
...
+        if (null === $encoding) {
+            $encoding = self::mb_internal_encoding();
+        }
+
+        try {
+            $validEncoding = @self::mb_check_encoding('', $encoding);
+        } catch (\ValueError $e) {
+            throw new \ValueError(sprintf('mb_ucfirst(): Argument #2 ($encoding) must be a valid encoding, "%s" given', $encoding));
+        }
```
zonuexe, line 883: `self::mb_internal_encoding()` performs validation checks when setting internal values, so adding a `try-catch` to the `else` avoids a small amount of overhead. What do you think?

```suggestion
        } else {
            try {
                $validEncoding = @self::mb_check_encoding('', $encoding);
            } catch (\ValueError $e) {
                throw new \ValueError(sprintf('mb_ucfirst(): Argument #2 ($encoding) must be a valid encoding, "%s" given', $encoding));
            }
        }
```

src/Mbstring/Mbstring.php
```diff
@@ -871,6 +871,49 @@ public static function mb_str_pad(string $string, int $length, string $pad_strin
...
+
+        // BC for PHP 7.3 and lower
+        if (!$validEncoding) {
+            throw new \ValueError(sprintf('mb_ucfirst(): Argument #2 ($encoding) must be a valid encoding, "%s" given', $encoding));
+        }
```
zonuexe, line 891: `self::mb_internal_encoding()` performs validation checks when setting internal values, so adding a `try-catch` to the `else` avoids a small amount of overhead. What do you think?

```suggestion
        } else {
            try {
                $validEncoding = @self::mb_check_encoding('', $encoding);
            } catch (\ValueError $e) {
                throw new \ValueError(sprintf('mb_ucfirst(): Argument #2 ($encoding) must be a valid encoding, "%s" given', $encoding));
            }
            
            // BC for PHP 7.3 and lower
            if (!$validEncoding) {
                throw new \ValueError(sprintf('mb_ucfirst(): Argument #2 ($encoding) must be a valid encoding, "%s" given', $encoding));
            }
        }
```

Ayesh: You are right I understand the `else` block would avoid the non-null calls.i tried to follow the rest of the polyfill methods, so I'd prefer to keep this one like it is now.

Ayesh:
Thanks a lot for the review @zonuexe those were great. I addressed all the points except the `else` block remark, mainly to make sure we follow the other polyfill methods, although there is indeed a merit if that changes. Perhaps it's something we can do for all methods in a separate PR?

src/Mbstring/Mbstring.php
zonuexe: Perhaps this code can be condensed by combining it into a method like `self::assertEncoding()`.  Let's try another PR.

nicolas-grekas:
Thank you @Ayesh.

PR #467, opened 2024-03-14
Update nullable types for PHP 8.4

mbeccati:
The PR is meant to fix the deprecation notices raised since php/php-src@330cc5cdb2096c5d41041eaae1f1cc0ed7e36898

I had tried to update the php-cs-fixer rule, but running the fixer is re-formatting all the method bodies and I couldn't figure out how to cleanly get the desired result, so I used a trusted editor and some preg-fu to get the desired result.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,1 @@
+a7548a261ad4bcdb675adceb6604cf1681b1f6e2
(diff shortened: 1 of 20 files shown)


nicolas-grekas:
Thank you @mbeccati.

mbeccati:
@nicolas-grekas However, I'm afraid that the commit to ignore now is aaaa1c99e3c9e40b26c9177a6d018d5a2aad8360 due to the squash merge ;-)

derrabus:
Send a PR, please.

PR #469, opened 2024-03-30
Mb trim functions

### LINKED ISSUE #458: [PHP 8.4] `mb_trim`, `mb_ltrim`, and `mb_rtrim` polyfills

Fan2Shrek:
Fixes #458 

Implementaion of `mb_trim`, `mb_ltrim` and `mb_rtrim`
Tests from the php-src repository

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -79,8 +79,6 @@ final class Mbstring
         ['μ', 's', 'ι',        'σ', 'β',        'θ',        'φ',        'π',        'κ',        'ρ',        'ε',        "\xE1\xB9\xA1", 'ι'],
     ];
 
-    private const CHARACTERS = " \f\n\r\t\v\x00\u{00A0}\u{1680}\u{2000}\u{2001}\u{2002}\u{2003}\u{2004}\u{2005}\u{2006}\u{2007}\u{2008}\u{2009}\u{200A}\u{2028}\u{2029}\u{202F}\u{205F}\u{3000}\u{0085}\u{180E}";
-
     private static $encodingList = ['ASCII', 'UTF-8'];
     private static $language = 'neutral';
     private static $internalEncoding = 'UTF-8';
@@ -88,7 +86,7 @@ final class Mbstring
     public static function mb_convert_encoding($s, $toEncoding, $fromEncoding = null)
     {
         if (\is_array($s)) {
-            if (PHP_VERSION_ID < 70200) {
+            if (\PHP_VERSION_ID < 70200) {
                 trigger_error('mb_convert_encoding() expects parameter 1 to be string, array given', \E_USER_WARNING);
 
                 return null;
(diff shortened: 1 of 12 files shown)


Ayesh:
Upstream issue https://github.com/php/php-src/issues/13815 might affect the targets we are about to write polyfill for.

zonuexe:
https://github.com/php/php-src/pull/13820 has been merged.

src/Mbstring/Mbstring.php
```diff
@@ -991,4 +995,66 @@ private static function getEncoding($encoding)
...
+        $regex = sprintf($regex, $regexCharacter, $regexCharacter);
+
+        if ('ASCII' === mb_detect_encoding($characters) && 'ASCII' === mb_detect_encoding($string) && !empty(array_intersect(str_split(self::CHARACTERS), str_split($string)))) {
+            $options = 'g';
+        } else {
+            $options = '';
+        }
+        
+        try {
+            $test = mb_ereg_replace($regex, "", $string, $options);
```
nicolas-grekas: when using the mbstring polyfill, this function won't be available I suppose.
Can't we rely only on PCRE?

src/Mbstring/Mbstring.php
```diff
@@ -991,4 +995,66 @@ private static function getEncoding($encoding)
...
+
+            if (null === $test) {
+                throw new \Exception();
+            }
+
+            return $test;
+        } catch (\Exception $e) {
+            return preg_replace(sprintf('/%s/', $regex), "", $string);
```
nicolas-grekas: let's use concatenation for such simple things (or better: add the `/` from the start if we know we're always going to use PCRE)

src/Php84/Php84.php
```diff
@@ -13,11 +13,13 @@
 
 /**
  * @author Ayesh Karunaratne <ayesh@aye.sh>
+ * @author Pierre Ambroise <pierre27.ambroise@gmail.com>
  *
  * @internal
  */
 final class Php84
 {
+    private const CHARACTERS = " \f\n\r\t\v\x00\u{00A0}\u{1680}\u{2000}\u{2001}\u{2002}\u{2003}\u{2004}\u{2005}\u{2006}\u{2007}\u{2008}\u{2009}\u{200A}\u{2028}\u{2029}\u{202F}\u{205F}\u{3000}\u{0085}\u{180E}";
```
nicolas-grekas: missing blank line after

src/Mbstring/Mbstring.php
```diff
@@ -79,6 +82,7 @@ final class Mbstring
     private static $encodingList = ['ASCII', 'UTF-8'];
     private static $language = 'neutral';
     private static $internalEncoding = 'UTF-8';
+    private const CHARACTERS = " \f\n\r\t\v\x00\u{00A0}\u{1680}\u{2000}\u{2001}\u{2002}\u{2003}\u{2004}\u{2005}\u{2006}\u{2007}\u{2008}\u{2009}\u{200A}\u{2028}\u{2029}\u{202F}\u{205F}\u{3000}\u{0085}\u{180E}";
```
nicolas-grekas: should be moved next to the previous private const

tests/Mbstring/MbstringTest.php
```diff
@@ -780,4 +780,127 @@ public static function lcFirstDataProvider(): array {
...
+    }
+
+    /**
+     * @covers \Symfony\Polyfill\Php84\Php84::mb_ltrim
+     * 
+     * @dataProvider mbLTrimProvider
+     */
+    public function testMbLTrim(string $expected, string $string, ?string $characters = null, ?string $encoding = null): void
+    {
+        $this->assertEquals($expected, mb_ltrim($string, $characters, $encoding));
```
nicolas-grekas: please use assertSame everywhere possible

src/Php84/bootstrap.php
```diff
@@ -38,3 +38,15 @@ function array_any(array $array, callable $callback): bool { return p\Php84::arr
 if (!function_exists('array_all')) {
     function array_all(array $array, callable $callback): bool { return p\Php84::array_all($array, $callback); }
 }
+
+if (!function_exists('mb_trim')) {
```
stof: Those `mb_*` function should be defined in the php84 polyfill only when mbstring is loaded, to avoid risks of breaking feature detection by having those mbstring functions without having others.
Projects wanting to use those functions without depending on the extension are expected to require the mbstring polyfill.
this is also what we did for the php74 polyfill.

stof (changes requested):

src/Mbstring/Mbstring.php
Fan2Shrek: I'm not sure I understand. Is it something like this ?

```suggestion
            $test = preg_replace($regex, "", $string, $options);
```

If so this does not work :( 
?

nicolas-grekas: I updated the implementation with what I meant

nicolas-grekas:
Thank you @Fan2Shrek.

PR #471, opened 2024-03-30
[PHP 8.4] Add mb_ucfirst() and mb_lcfirst()

### LINKED ISSUE #470: [PHP 8.4] `mb_ucfirst()` and `mb_lcfirst()` polyfills
* [PHP RFC: Multibyte for ucfirst, lcfirst functions, mb_ucfirst mb_lcfirst](https://wiki.php.net/rfc/mb_ucfirst)
* [New `mb_ucfirst` and `mb_lcfirst` functions - PHP 8.4 • PHP.Watch](https://php.watch/versions/8.4/mb_ucfirst-mb_ucfirst)

These functions have been merged at https://github.com/php/php-src/pull/13161.

zonuexe:
resolve https://github.com/symfony/polyfill/issues/470

This implementation is also tested and benchmarked in my repository: https://github.com/zonuexe/polyfill-mb_ulcfirst

### DIFF
--- a/tests/Mbstring/MbstringTest.php
+++ b/tests/Mbstring/MbstringTest.php
@@ -727,4 +727,28 @@ class MbstringTest extends TestCase
         yield ['mb_str_pad(): Argument #4 ($pad_type) must be STR_PAD_LEFT, STR_PAD_RIGHT, or STR_PAD_BOTH', '▶▶', 6, ' ', 123456];
         yield ['mb_str_pad(): Argument #5 ($encoding) must be a valid encoding, "unexisting" given', '▶▶', 6, ' ', \STR_PAD_BOTH, 'unexisting'];
     }
+
+    /**
+     * @covers \Symfony\Polyfill\Mbstring\Mbstring::mb_lcfirst
+     */
+    public function test_mb_lcfirst(): void
+    {
+        $this->assertSame('', p::mb_lcfirst('', 'UTF-8'));
+        $this->assertSame('ａＢＳ', p::mb_lcfirst('ＡＢＳ', 'UTF-8'));
+        $this->assertSame('xin chào', p::mb_lcfirst('Xin chào', 'UTF-8'));
+        $this->assertSame('đẹp quá!', p::mb_lcfirst('Đẹp quá!', 'UTF-8'));
+    }
+
+    /**
+     * @covers \Symfony\Polyfill\Mbstring\Mbstring::mb_ucfirst
+     */
+    public function test_mb_ucfirst(): void
+    {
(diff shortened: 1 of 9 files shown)


zonuexe:
[These errors](https://github.com/symfony/polyfill/actions/runs/8494636417/job/23270054252?pr=471#step:6:1350) below looks strange:

```
40) Warning
Incompatible signature for PHP >= 8:
- mb_ucfirst(string $string, ?string $encoding = null): string
+ mb_ucfirst(string $string, ?string $encoding = null): string

41) Warning
Incompatible signature for PHP >= 8:
- mb_lcfirst(string $string, ?string $encoding = null): string
+ mb_lcfirst(string $string, ?string $encoding = null): string
```

Internal functions up to PHP 8.0 implicitly accepted `null` as a `string`. It seems that the type is declared as `?string` instead of `string` for compatibility in polyfills up to php80.

https://github.com/symfony/polyfill/blob/df4793d8de3390de9bb0aaf02cff7fa67ddc4a2a/src/Util/TestListenerTrait.php#L126-L136

In the code below, `?` is removed to check the difference between the internal function and the polyfill interface.

PHP 8.4 deprecates implicit nullable, so this test should be replaced with more stringent checks.

* [PHP: rfc:deprecate-implicitly-nullable-types](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types)

tests/Mbstring/MbstringTest.php
```diff
@@ -105,7 +105,7 @@ public function testDecodeNumericEntity()
         $this->assertSame('déjà &#0; â ã', mb_decode_numericentity('déjà &#0; &#225; &#226;', $convmap, 'UTF-8'));
 
         $bogusDecEntities = 'déjà &#0; &#225;&#225; &#&#225&#225 &#225 &#225t';
-        $this->assertSame('déjà &#0; ââ &#&#225â â ât', mb_decode_numericentity($bogusDecEntities, $convmap, 'UTF-8'));
+        $this->assertSame('déjà &#0; ââ &#&#225â â ât', p::mb_decode_numericentity($bogusDecEntities, $convmap, 'UTF-8'));
```
zonuexe, line 108: This test was affected by changes in the native `mb_decode_numericentity()` function.

https://3v4l.org/fpGFH

It seems that other tests also need to call polyfill methods instead of built-in function.

Ayesh:
Possibly duplicate of #466

zonuexe:
@Ayesh Sorry, I missed that PR because the issue didn't exist. It looks like my patch is covered by your implementation.

Ayesh:
Let's put the efforts together :) 
If you see anything that we could improve in the other PR, or if you think merging this PR is the way to go, please do comment.

PR #479, opened 2024-05-30
[Mbstring] The first arg of mb_convert_encoding can be an array

alanpoulain:
As the [documentation](https://www.php.net/manual/en/function.mb-convert-encoding.php) says, since PHP 7.2 the first argument of `mb_convert_encoding` can be an array.


alanpoulain:
CI failures seem unrelated.

nicolas-grekas:
Thank you @alanpoulain.

PR #481, opened 2024-06-01
Add PHP 8.4 `array_*` polyfill

joshuaruesweg:
See: https://wiki.php.net/rfc/array_find

### DIFF
--- a/src/Php84/Php84.php
+++ b/src/Php84/Php84.php
@@ -63,4 +63,48 @@ final class Php84
 
         return $firstChar . mb_substr($string, 1, null, $encoding);
     }
+
+    public static function array_find(array $array, callable $callback)
+    {
+        foreach ($array as $key => $value) {
+            if ($callback($value, $key)) {
+                return $value;
+            }
+        }
+
+        return null;
+    }
+
+    public static function array_find_key(array $array, callable $callback)
+    {
+        foreach ($array as $key => $value) {
+            if ($callback($value, $key)) {
+                return $key;
+            }
+        }
+
+        return null;
+    }
+
+    public static function array_any(array $array, callable $callback): bool
+    {
+        foreach ($array as $key => $value) {
+            if ($callback($value, $key)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public static function array_all(array $array, callable $callback): bool
+    {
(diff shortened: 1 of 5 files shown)


src/Php84/Php84.php
```diff
@@ -63,4 +63,48 @@ public static function mb_lcfirst(string $string, ?string $encoding = null): str
 
         return $firstChar . mb_substr($string, 1, null, $encoding);
     }
+
+    public static function array_find(array $array, callable $callback)
+    {
+        foreach ($array as $key => $value) {
+            if ($callback($value, $key)) {
```
keradus, line 70:
```suggestion
            if (true === $callback($value, $key)) {
```

i really wonder why not to rely on strict, boolean interface of callback result

joshuaruesweg: Thus typecheck would not match the nativ implementation which also has no strict typecheck: 

```
var_dump(array_find(['', 'aa'], fn($value) => strlen($value))); // string(2) "aa"
```

Ayesh:
LGTM, you are the RFC author after all :)

I also wrote a polyfill as a separate package, and used a [conditional declaration](https://github.com/PHP-Polyfills/array-find/blob/69874552b2c2c988585929d722e292a3a859e673/src/array_find.php#L52) to add the `mixed` return type only in PHP >=8.0

IonBazan (approved):

tests/Php84/Php84Test.php
```diff
@@ -30,6 +30,34 @@ public function testMbLcFirst(string $string, string $expected): void {
         $this->assertSame($expected, mb_lcfirst($string));
     }
 
+    /**
+     * @dataProvider arrayFindDataProvider
+     */
+    public function testArrayFind(array $array, callable $callback, $expected): void {
```
OskarStark, line 36: Don't we ensure our Symfony CS here in this repo?

nicolas-grekas: not automatically, I run php-cs-fixer from time to time for the moment

OskarStark: OK

vudaltsov (approved):

nicolas-grekas:
Thank you @joshuaruesweg.

PR #483, opened 2024-06-05
[PHP 8.4][Intl] Add `grapheme_str_split`

Ayesh:
Adds a polyfill for the `grapheme_str_split` function added in PHP 8.4.

Requires PHP 7.3, because the polyfill is based on `\X` Regex, and it only works properly on PCRE2, which [only comes with PHP 7.3+](https://php.watch/versions/7.3/pcre2).

Further, there are some cases that the polyfill cannot split complex characters (such as two consecutive country flag Emojis). This is now fixed in [PCRE2Project/pcre2#410](https://github.com/PCRE2Project/pcre2/issues/410). However, this change will likely only make it to PHP 8.4.

References:
 - [RFC: Grapheme cluster for `str_split` function: `grapheme_str_split`](https://wiki.php.net/rfc/grapheme_str_split)
 - [PHP.Watch: PHP 8.4: New `grapheme_str_split` function](https://php.watch/versions/8.4/grapheme_str_split) (polyfill source)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Polyfills are provided for:
 - the `Deprecated` attribute introduced in PHP 8.4;
 - the `mb_trim`, `mb_ltrim` and `mb_rtrim` functions introduced in PHP 8.4;
 - the `CURL_HTTP_VERSION_3` and `CURL_HTTP_VERSION_3ONLY` constants introduced in PHP 8.4;
+- the `grapheme_str_split` function introduced in PHP 8.4;
 - the `get_error_handler` and `get_exception_handler` functions introduced in PHP 8.5;
 - the `NoDiscard` attribute introduced in PHP 8.5;
 - the `array_first` and `array_last` functions introduced in PHP 8.5;
--- a/src/Intl/Grapheme/Grapheme.php
+++ b/src/Intl/Grapheme/Grapheme.php
@@ -26,6 +26,7 @@ namespace Symfony\Polyfill\Intl\Grapheme;
  * - grapheme_strrpos  - Find position (in grapheme units) of last occurrence of a string
  * - grapheme_strstr   - Returns part of haystack string from the first occurrence of needle to the end of haystack
  * - grapheme_substr   - Return part of a string
+ * - grapheme_str_split - Splits a string into an array of individual or chunks of graphemes
  *
  * @author Nicolas Grekas <p@tchwork.com>
  *
@@ -189,6 +190,37 @@ final class Grapheme
     public static function grapheme_strstr($s, $needle, $beforeNeedle = false)
     {
         return mb_strstr($s, $needle, $beforeNeedle, 'UTF-8');
+    }
+
+    public static function grapheme_str_split($s, $len = 1)
+    {
+        if (0 > $len || 1073741823 < $len) {
+            if (80000 > \PHP_VERSION_ID) {
+                return false;
+            }
+
+            throw new \ValueError('grapheme_str_split(): Argument #2 ($length) must be greater than 0 and less than or equal to 1073741823.');
+        }
+
+        if ('' === $s) {
+            return [];
+        }
+
+        if (!preg_match_all('/('.SYMFONY_GRAPHEME_CLUSTER_RX.')/u', $s, $matches)) {
+            return false;
+        }
+
+        if (1 === $len) {
+            return $matches[0];
+        }
+
+        $chunks = array_chunk($matches[0], $len);
+
+        foreach ($chunks as &$chunk) {
+            $chunk = implode('', $chunk);
+        }
+
+        return $chunks;
     }
 
     private static function grapheme_position($s, $needle, $offset, $mode)
--- a/src/Intl/Grapheme/README.md
+++ b/src/Intl/Grapheme/README.md
@@ -21,6 +21,7 @@ This component provides a partial, native PHP implementation of the
 - [`grapheme_strstr`](https://php.net/grapheme_strstr): Returns part of haystack string from
   the first occurrence of needle to the end of haystack
 - [`grapheme_substr`](https://php.net/grapheme_substr): Return part of a string
+- [`grapheme_str_split](https://php.net/grapheme_str_split): Splits a string into an array of individual or chunks of graphemes
 
 More information can be found in the
 [main Polyfill README](https://github.com/symfony/polyfill/blob/main/README.md).
--- a/src/Intl/Grapheme/bootstrap.php
+++ b/src/Intl/Grapheme/bootstrap.php
@@ -56,3 +56,6 @@ if (!function_exists('grapheme_strstr')) {
 if (!function_exists('grapheme_substr')) {
     function grapheme_substr($string, $offset, $length = null) { return p\Grapheme::grapheme_substr($string, $offset, $length); }
 }
+if (!function_exists('grapheme_str_split')) {
+    function grapheme_str_split($string, $length = 1) { return p\Grapheme::grapheme_str_split($string, $length); }
+}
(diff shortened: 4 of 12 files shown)


Ayesh:
(working on the Intl changes, I'll mark the PR ready then)

derrabus:
Thank you. I think, we should add this polyfill to the intl-grapheme polyfill as well.

Ayesh:
Thank you @derrabus - I added polyfill and tests for `grapheme_str_split` to the Intl polyfill too.

README.md
```diff
@@ -68,6 +68,7 @@ Polyfills are provided for:
 - the `Date*Exception/Error` classes introduced in PHP 8.3;
 - the `SQLite3Exception` class introduced in PHP 8.3;
 - the `mb_ucfirst` and `mb_lcfirst` functions introduced in PHP 8.4;
+- the `grapheme_str_split` function introduced in PHP 8.4 (requires PHP >= 7.3);
```
nicolas-grekas: I'd remove the need for PHP 7.3 as that makes the polyfill dangerous IMHO
Instead, we can borrow from SYMFONY_GRAPHEME_CLUSTER_RX as defined in the intl-grapheme polyfill
we can copy/paste the string I think

src/Intl/Grapheme/Grapheme.php
```diff
@@ -191,6 +192,38 @@ public static function grapheme_strstr($s, $needle, $beforeNeedle = false)
...
+            throw new \ValueError('grapheme_str_split(): Argument #2 ($length) must be greater than 0 and less than or equal to 1073741823.');
+        }
+
+        if ($s === '') {
+            return [];
+        }
+
+        preg_match_all('/\X/u', $s, $matches);
+
+        if (empty($matches[0])) {
```
nicolas-grekas: please avoid using empty

src/Intl/Grapheme/Grapheme.php
```diff
@@ -191,6 +192,38 @@ public static function grapheme_strstr($s, $needle, $beforeNeedle = false)
...
+            return false;
+        }
+
+        if ($len === 1) {
```
nicolas-grekas: we use yoda style everywhere (I can fix that using php-cs-fixer, there are way more things to fix, cs-wise)

src/Intl/Grapheme/Grapheme.php
```diff
@@ -191,6 +192,38 @@ public static function grapheme_strstr($s, $needle, $beforeNeedle = false)
...
+            return $matches[0];
+        }
+
+        $chunks = array_chunk($matches[0], $len);
+
+        array_walk($chunks, static function(&$value) {
```
nicolas-grekas: I'm pretty sure using a foreach is going to be faster here: no need for one function call per iteration

src/Intl/Grapheme/bootstrap.php
```diff
@@ -56,3 +56,7 @@ function grapheme_strstr($haystack, $needle, $beforeNeedle = false) { return p\G
 if (!function_exists('grapheme_substr')) {
     function grapheme_substr($string, $offset, $length = null) { return p\Grapheme::grapheme_substr($string, $offset, $length); }
 }
+
+if (\PHP_VERSION_ID >= 70300) {
```
nicolas-grekas: should be removed following my comment above

src/Intl/Grapheme/bootstrap73.php
```diff
@@ -0,0 +1,17 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Php84 as p;
+
+if (!function_exists('grapheme_str_split') && function_exists('grapheme_substr')) {
+    function grapheme_str_split(string $string, int $length = 1) { return p\Php84::grapheme_str_split($string, $length); }
```
nicolas-grekas: should be moved to bootstrap.php following my comment above

nicolas-grekas:
Friendly ping @Ayesh :)

Ayesh:
Thank you @nicolas-grekas - really helpful comments, I addressed them and force-pushed. `\X` regex polyfill for PCRE1 is _very_ cool, it worked beautifully, 10,000 IQ regex 🤯 :)

tests/Intl/Grapheme/GraphemeTest.php
```diff
@@ -209,4 +209,30 @@ public function testGraphemeStrstr()
         $this->assertSame('국어', grapheme_strstr('한국어', '국'));
         $this->assertSame('ÉJÀ', grapheme_stristr('DÉJÀ', 'é'));
     }
+
+    /**
+     * @dataProvider graphemeStrSplitDataProvider
+     * @requires PHP 7.3
```
nicolas-grekas: stil needed?

tests/Intl/Grapheme/GraphemeTest.php
```diff
@@ -209,4 +209,30 @@ public function testGraphemeStrstr()
...
+     */
+    public function testGraphemeStrSplit(string $string, int $length, array $expectedValues) {
```
nicolas-grekas:
```suggestion
    public function testGraphemeStrSplit(string $string, int $length, array $expectedValues)
    {
```

tests/Intl/Grapheme/GraphemeTest.php
```diff
@@ -209,4 +209,30 @@ public function testGraphemeStrstr()
...
+        $this->assertSame($expectedValues, grapheme_str_split($string, $length));
+    }
+
+    public static function graphemeStrSplitDataProvider(): array {
```
nicolas-grekas:
```suggestion
    public static function graphemeStrSplitDataProvider(): array
    {
```

tests/Php84/Php84Test.php
```diff
@@ -319,4 +319,23 @@ public static function mbRTrimProvider(): iterable
 
         yield ["foo\n", "foo\n", 'o'];
     }
+
+    /**
+     * @dataProvider graphemeStrSplitDataProvider
+     */
+    public function testGraphemeStrSplit(string $string, int $length, array $expectedValues) {
```
nicolas-grekas:
```suggestion
    public function testGraphemeStrSplit(string $string, int $length, array $expectedValues)
    {
```

tests/Php84/Php84Test.php
```diff
@@ -319,4 +319,23 @@ public static function mbRTrimProvider(): iterable
...
+        $this->assertSame($expectedValues, grapheme_str_split($string, $length));
+    }
+
+    public static function graphemeStrSplitDataProvider(): array {
```
nicolas-grekas:
```suggestion
    public static function graphemeStrSplitDataProvider(): array
    {
```

nicolas-grekas (commented):
Just one question + minor CS things and GTM

Ayesh:
One last push, thank you for being patient with this 💜

nicolas-grekas:
So we have a test failure on PHP 7.2 :)
Maybe we should remove the corresponding test case? The fallback regexp doesn't account for ZWJ emojis IIRC

src/Intl/Grapheme/README.md
```diff
@@ -21,6 +21,7 @@ This component provides a partial, native PHP implementation of the
 - [`grapheme_strstr`](https://php.net/grapheme_strstr): Returns part of haystack string from
   the first occurrence of needle to the end of haystack
 - [`grapheme_substr`](https://php.net/grapheme_substr): Return part of a string
+- [`grapheme_str_split](https://php.net/grapheme_str_split): Splits a string into an array of individual or chunks of graphemes.
```
nicolas-grekas:
```suggestion
- [`grapheme_str_split](https://php.net/grapheme_str_split): Splits a string into an array of individual or chunks of graphemes
```

nicolas-grekas (approved):

Ayesh:
Perfect, fixed. 
So far, we exclude this ZW joiner case on PHP 7.2, and a known buggy PCRE2 \X capture on PCRE2 < 10.44 regardless of the PHP version.

nicolas-grekas:
Thank you @Ayesh.

PR #486, opened 2024-07-30
fix function signature tests

xabbuh:


stof:
IIRC, allowing `null` was done to allow compatibility with the coalescing mode of PHP's type, which allows `null` (but deprecated) until PHP 9 for native functions

nicolas-grekas:
@stof is right

xabbuh:
makes sense, I added two additional entries to the map to satisfy the tests instead

nicolas-grekas:
Thank you @xabbuh.

PR #489, opened 2024-08-27
[PHP 8.4] Add `CURL_HTTP_VERSION_3(ONLY)` constants

### LINKED ISSUE #488: [Feature Request] Constant CURL_HTTP_VERSION_3 in PHP 8.4
Can backporting new constant HTTP in 8.4?

Ayesh:
If the underlying Curl build supports it, setting the HTTP version should be theoretically possible. The availability of the constants do not guarantee the feature support, so declaring them in the polyfill should be same and effective.

 - PR that added constants: [php-src#15350](https://github.com/php/php-src/pull/15350)
 - libcurl doc on `CURLOPT_HTTP_VERSION`](https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html)
 - [PHP.Watch: HTTP/3 Request with PHP Curl Extension](https://php.watch/articles/php-curl-http3)
 - Codex for [`CURL_HTTP_VERSION_3`](https://php.watch/codex/CURL_HTTP_VERSION_3) and [`CURL_HTTP_VERSION_3ONLY`](https://php.watch/codex/CURL_HTTP_VERSION_3ONLY)

Fixes GH-488.

### DIFF
--- a/src/Php84/bootstrap.php
+++ b/src/Php84/bootstrap.php
@@ -15,6 +15,14 @@ if (\PHP_VERSION_ID >= 80400) {
     return;
 }
 
+if (defined('CURL_VERSION_HTTP3') || PHP_VERSION_ID < 80200 && function_exists('curl_version') && curl_version()['version'] >= 0x074200) { // libcurl >= 7.66.0
+    define('CURL_HTTP_VERSION_3', 30);
+
+    if (defined('CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256')) { // libcurl >= 7.80.0 (7.88 would be better but is slow to check)
+        define('CURL_HTTP_VERSION_3ONLY', 31);
+    }
+}
+
 if (!function_exists('array_find')) {
     function array_find(array $array, callable $callback) { return p\Php84::array_find($array, $callback); }
 }
--- a/tests/Php84/Php84Test.php
+++ b/tests/Php84/Php84Test.php
@@ -64,6 +64,18 @@ class Php84Test extends TestCase
         $this->assertSame($expected, array_all($array, $callback));
     }
 
+    /**
+     * @requires extension curl
+     */
+    public function testCurlHttp3Constants()
+    {
+        $ch = curl_init();
+        $hasH3 = defined('CURL_VERSION_HTTP3') && (curl_version()['features'] & CURL_VERSION_HTTP3);
+
+        $this->assertSame($hasH3, curl_setopt($ch, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_3));
+        $this->assertSame($hasH3 && defined('CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256'), curl_setopt($ch, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_3ONLY));
+    }
+
     public static function ucFirstDataProvider(): array
     {
         return [
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Polyfills are provided for:
 - the `array_find`, `array_find_key`, `array_any` and `array_all` functions introduced in PHP 8.4;
 - the `Deprecated` attribute introduced in PHP 8.4;
 - the `mb_trim`, `mb_ltrim` and `mb_rtrim` functions introduced in PHP 8.4;
+- the `CURL_HTTP_VERSION_3` and `CURL_HTTP_VERSION_3ONLY` constants introduced in PHP 8.4;
 
 It is strongly recommended to upgrade your PHP version and/or install the missing
 extensions whenever possible. This polyfill should be used only when there is no
--- a/src/Php84/README.md
+++ b/src/Php84/README.md
@@ -6,6 +6,7 @@ This component provides features added to PHP 8.4 core:
 - [`mb_ucfirst` and `mb_lcfirst`](https://wiki.php.net/rfc/mb_ucfirst)
 - [`array_find`, `array_find_key`, `array_any` and `array_all`](https://wiki.php.net/rfc/array_find)
 - [`Deprecated`](https://wiki.php.net/rfc/deprecated_attribute)
+- `CURL_HTTP_VERSION_3` and `CURL_HTTP_VERSION_3ONLY` constants
 
 More information can be found in the
 [main Polyfill README](https://github.com/symfony/polyfill/blob/main/README.md).


src/Php84/bootstrap.php
```diff
@@ -15,6 +15,11 @@
     return;
 }
 
+if (!defined('CURL_HTTP_VERSION_3')) {
+    define('CURL_HTTP_VERSION_3', 30);
+    define('CURL_HTTP_VERSION_3ONLY', 31);
+}
```
derrabus: In php/php-src#15350 the constants are only declared if the linked libcurl exposes them. Your patch however would register the constants unconditionally, even if ext-curl isn't loaded at all. cURL constants are often used for feature detection, which is why we should make sure they are available only in cases where PHP would also declare them.

stof:
Defining constants that cannot have effect would make the polyfilled case worse as it breaks feature detection without making the feature works. So -1 from me.

nicolas-grekas:
> Defining constants that cannot have effect would make the polyfilled case worse

Actually, passing the raw values does work. Curl accepts them (provided a recent enough version is installed). Only PHP doesn't expose these values as constants.

Ayesh:
Thank you for the reviews @derrabus @stof.

@derrabus you are right that declaring Curl constants gives a false pretense that the feature is available. However, all of our constant declarations are only dependent on libcurl and PHP version numbers; They do not check the availability of those features. Recent additions include DoH and HSTS support. The constants are declared as long as the Curl extension is compiled with the minimum require version or later, but it does not guarantee those features are available.

For HTTP/3, the version check php-src uses is whether libcurl is >= v7.66. While most setups will have the libcurl version, it will not have HTTP/3 support because as of now, default builds for libcurl do not come with HTTP/3 support (checked on Fedora, Ubuntu and Debian). Even without HTTP/3 support, PHP 8.4 will have these constants declared.

Like @nicolas-grekas said, any PHP version can use HTTP/3 if built with support for it.

```php
curl_setopt($ch, CURLOPT_HTTP_VERSION, 30); // returns false if HTTP/3 is not supported.
```

What the polyfill does is merely declare this as a proper constant. I think having this in the polyfill will can be helpful to avoid magic numbers.

The technically correct way to check HTTP/3 availability would be:

```
$hasHttp3 = !empty(curl_version()['feature_list']['HTTP3']); // PHP 8.4+ only
$hasHttp3 = curl_version()['features'] & 33554432); // or use CURL_VERSION_HTTP3 available on PHP 8.2.
```

That said, we cannot polyfill all Curl constants. We can't polyfill Curl options, but we should be able to polyfill Curl option _values_ where they make sense. For example, we [added](https://github.com/php/php-src/pull/13255) [`CURLOPT_PREREQFUNCTION`](https://php.watch/codex/CURLOPT_PREREQFUNCTION) for PHP 8.4 but we cannot technically polyfill it.

derrabus:
> However, all of our constant declarations are only dependent on libcurl and PHP version numbers; They do not check the availability of those features.

Understood. But this PR ignores the libcurl version and the availability of ext-curl completely.

* Do we expect the constants to be declared if PHP is compiled without ext-curl?
* Do we expect the constants to be declared if PHP is compiled against a libcurl older than 7.66?

Let's look at how this is implemented in PHP:

https://github.com/php/php-src/blob/3ed884fab73c5807fe2b3001ee79545c27ad5e70/ext/curl/curl_arginfo.h#L854-L856

If I read that piece of code correctly, the answer to both questions should be no. But this is not what this PR implements.

src/Php84/bootstrap.php
```diff
@@ -15,6 +15,11 @@
     return;
 }
 
+if (!defined('CURL_HTTP_VERSION_3')) {
```
nicolas-grekas:
```suggestion
if (extension_loaded('curl')) {
```

tests/Php84/Php84Test.php
```diff
@@ -64,6 +64,22 @@ public function testArrayAll(array $array, callable $callback, bool $expected)
         $this->assertSame($expected, array_all($array, $callback));
     }
 
+    public function testCurlHttp3Constants() {
```
nicolas-grekas:
```suggestion
    public function testCurlHttp3Constants()
    {
```

tests/Php84/Php84Test.php
```diff
@@ -64,6 +64,22 @@ public function testArrayAll(array $array, callable $callback, bool $expected)
...
+        if (!function_exists('curl_version')) {
+            $this->markTestSkipped('Curl extension is not available.');
```
nicolas-grekas: better use an `@requires extension curl` annotation

tests/Php84/Php84Test.php
```diff
@@ -64,6 +64,22 @@ public function testArrayAll(array $array, callable $callback, bool $expected)
...
+        }
+
+        // If the curl_version()['features'] bitmask contains the bits for CURL_VERSION_HTTP3,
+        // it means Curl supports HTTP3.
+        if (!(curl_version()['features'] & 33554432)) { // 33554432 = CURL_VERSION_HTTP3
```
nicolas-grekas: why is this check needed?

nicolas-grekas: per https://github.com/php/php-src/blob/3ed884fab73c5807fe2b3001ee79545c27ad5e70/ext/curl/curl_arginfo.h#L854-L856
we'd should rather check the version of curl in addition to checking if the ext is installed?

Ayesh:
Thank you for the pointers @nicolas-grekas @derrabus. I think you are right, it doesn't make sense to expose the constants on Curl < 7.66. I updated the PR to check for it and a `function_exists` check. 

`CURL_HTTP_VERSION_3ONLY` constant is Curl 7.88+, so I updated the declaration to check for it too. 

In my previous attempt, the bitmask check was to make sure Curl supports HTTP/3 with a `true` assertion. As @nicolas-grekas pointed out, I changed it to just check the Curl version regardless of the HTTP/3 availability of the underlying system. The check now makes sure the constant is declared and Curl accepted or rejected it with a `true` or `false` return value.

nicolas-grekas (approved):

derrabus (approved):

src/Php84/bootstrap.php
```diff
@@ -15,6 +15,14 @@
     return;
 }
 
+if (function_exists('curl_version') && curl_version()['version'] >= 0x074200) { // libcurl >= 7.66.0
```
nicolas-grekas: We can turn this into a fast check in PHP 8.2, see https://github.com/php/php-src/pull/8720

```suggestion
if (defined('CURL_VERSION_HTTP3') || PHP_VERSION_ID < 80200 && function_exists('curl_version') && curl_version()['version'] >= 0x074200) { // libcurl >= 7.66.0
```

src/Php84/bootstrap.php
```diff
@@ -15,6 +15,14 @@
...
+    define('CURL_HTTP_VERSION_3', 30);
+
+    if (curl_version()['version'] >= 0x075800) { // libcurl >= 7.88.0
```
nicolas-grekas: I'm a bit annoyed by this slow path, how about not doing the check and always declare the const next to the previous one?
Accuracy of the polyfilling on this aspect doesn't look more important than raw perf here to me.

derrabus: Is the `curl_version()` call really that expensive? 😞

nicolas-grekas: Things add up: there are many polyfills to load: if we're not careful with each of them, this might make things slow without us realizing. For now, most (all?) checks are resolved at compile-time so they're free.

Ayesh: It's not the slowest call, but `curl_version()` has a lot of information mostly by checking bitmasks for protocol and feature support. We only need the version here, so yeah I agree it makes sense to use something like `defined('CURLOPT_UPLOAD_BUFFERSIZE')` that also implicitly ensures libcurl >= 7.62

nicolas-grekas:
```suggestion
    if (defined('CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256')) { // libcurl >= 7.80.0 (7.88 would be better but is slow to check)
```

derrabus: LGTM

nicolas-grekas:
Thank you @Ayesh.

src/Php84/bootstrap.php
```diff
@@ -15,6 +15,14 @@
     return;
 }
 
+if (defined('CURL_VERSION_HTTP3') || PHP_VERSION_ID < 80200 && function_exists('curl_version') && curl_version()['version'] >= 0x074200) { // libcurl >= 7.66.0
```
stof, line 18: Why PHP 8.2.0 as boundary ? the PR you linked was merged in 8.4-dev so we might need the polyfill on 8.2 or 8.3

nicolas-grekas: https://github.com/php/php-src/pull/8720 this one is merged in 8.2

stof: but it does not add those constants. So we could be on PHP 8.2.12 or 8.3.5 with `CURL_VERSION_HTTP3` not being defined and we would never enter this `if` even if we have the latest libcurl

derrabus: @stof: If the `CURL_VERSION_HTTP3` constant does not exist although we're on PHP 8.2, we know for sure that we either don't have ext-curl or the linked libcurl is too old, so we don't need to call `curl_version()` anymore.

nicolas-grekas: https://github.com/php/php-src/pull/8720/files#diff-29336a112e08c5b5caa7950afb6601010867f40a4643a33263646504277f7b33R1195

this adds the constant if php is compiled with a recent enough version of curl

PR #491, opened 2024-08-29
Security Fix: Argument of a function used in an include/require, could lead to File Inclusion

vyskoczilova:
Argument of a function used in an include/require, could lead to File Inclusion | audit.php.lang.security.file.inclusion-arg
-- | --

Error from semgrep
audit.php.lang.security.file.inclusion-arg

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -953,8 +953,8 @@ final class Mbstring
 
     private static function getData($file)
     {
-        if (file_exists($file = __DIR__.'/Resources/unidata/'.$file.'.php')) {
-            return require $file;
+        if (file_exists(__DIR__.'/Resources/unidata/'.$file.'.php')) {
+            return require __DIR__.'/Resources/unidata/'.$file.'.php';
         }
 
         return false;


derrabus (commented):
Sorry, but I don't understand this change. What problem does it fix?

nicolas-grekas:
I'm like @derrabus. I has several guesses and concerns about this PR, let me share them with you in case this helps in any way:

I feel like this is something spotted by an automated security scanner. Here, the tool is missing that the method is private and is never called with arbitrary data. A quick look will indeed show you that we pass it only static data. The report is thus invalid.

About the patch, supposing there would be a security issue, it doesn't fix anything since it reuses the argument without doing any filtering nor validation of it. If the tool doesn't spot this as a security anymore, then you'd better trash that tool.

Then, such tools are just stupid automata. Please perform the analyses I just did before submitting issues in the future, that'll save time for everybody.

And last but not least, you should know that sharing possible security issues publicly is a very much discouraged practice. Pretty much all serious open-source projects have a security policy in place that you can read to know how to report security issues.

With all that said, I hope you understand why I'm closing.

Ayesh:
Looks pretty close to an automated scan result, and the fix doesn't seem to offer any value.

vyskoczilova:
@nicolas-grekas, you're right. I appreciate your approach and explanation. I will remember and keep it in my mind for the future; I have learned quite a lot from it. I'm really sorry for the inconvenience.

Unfortunately, this is not my tool; I was part of a quality check for WooCoommerce official plugins that takes the analysis from Semver as linked: https://qit.woo.com/docs/managed-tests/security/. I will report it as a false positive.

Issue #425, opened 2023-01-26
[Tests] Polyfills not loaded


IonBazan:
When running the test suite on my local environment (PHP 8.1), the polyfill is not loaded correctly, causing the new functions to be not registered. Interestingly, the test is marked as skipped, instead of error/warning 🤔 

I bet it has something to do with the custom test runner/hooks but not sure what's wrong there.

```
vendor/bin/simple-phpunit --filter php83 -vvv
PHPUnit 9.5.26 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.14
Configuration: /Users/ionbazan/work/polyfill/phpunit.xml.dist
Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!

Testing
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS                        42 / 42 (100%)

Time: 00:00.317, Memory: 74.00 MB

There were 42 skipped tests:

1) Symfony\Polyfill\Tests\Php83\Php83Test::testJsonValidate with data set #0 (false, '', 'Syntax error')
Internal function not found: json_validate

/Users/ionbazan/work/polyfill/tests/Php83/Php83Test.php:24

2) Symfony\Polyfill\Tests\Php83\Php83Test::testJsonValidate with data set #1 (false, '.', 'Syntax error')
Internal function not found: json_validate

/Users/ionbazan/work/polyfill/tests/Php83/Php83Test.php:24

[...]
```

nicolas-grekas:
That's how it's supposed to work to me: the internal function (the native one, non-polyfilled) is not found since you're on PHP 8.1
Try running this on PHP 8.3, it should be green.

IonBazan:
That doesn't make much sense to me. From what I remember, we were including `bootstrap.php` files using `TestListenerTrait` which made sure the function is registered globally - this is to use same test code on PHP 8.1 and 8.3 and compare results between polyfilled and native implementation.

IonBazan:
OK, I think I got it now - I always thought we run the future-compatible tests on old versions too, and that the functions are globally registered but now I see the mechanism is much more sophisticated. I just wonder if that won't produce any regressions, as we test `json_validate` only on PHP 8.3 and up, we could technically put some new syntax which is not compatible with older versions and still observe a green CI 🤔

nicolas-grekas:
[...]

PR #492, opened 2024-09-09
Bump to PHP 7.2, stick to phpunit 8.5

nicolas-grekas:
I think it's start to shade PHP 7.1 out. Of course the polyfill will be kept functional, but we won't care about it anymore from a maintenance PoV.

Sticking to phpunit 8.5 so that we can focus on one version of it, and because it has [lifetime support](https://phpunit.de/supported-versions.html) and PHP 7.2 as a minimum version.


derrabus:
What has changed since #428?

From Symfony's PoV, taking into account that the active maintenance of the 5.4 branch will end about two months from now, we might as well wait a little longer and bump to PHP 8.1 directly.

src/Php72/composer.json
```diff
@@ -1,32 +0,0 @@
...
-            "name": "Nicolas Grekas",
-            "email": "p@tchwork.com"
-        },
-        {
-            "name": "Symfony Community",
-            "homepage": "https://symfony.com/contributors"
-        }
-    ],
-    "require": {
-        "php": ">=7.1"
```
derrabus, line 19: In #428, I did not remove the package just yet, but instead bumped it to PHP 7.2 as well and made it a no-op package. We've done this with other polyfills in the past as well, e.g. https://packagist.org/packages/symfony/polyfill-php71#v1.20.0

nicolas-grekas:
I made this PR after realizing that #486 was required because on PHP 7.1 we have to use var-dumper 4.4, which didn't get the fixes for nullable arguments.
Then I also realized that phpunit 8.5 has lifetime support and also php >= 7.2.
The CI is behind at the moment with e.g. #425 + failures and I don't think it would be productive to work on this while also having to consider several versions of phpunit.

PHP 8.1 is too early IMHO, at least I don't see similar arguments.

src/Php72/composer.json
nicolas-grekas: why did we do this? what's the benefit?
people with php < 7.2 can still install a previous version

derrabus:
If PHP 7.1 support really blocks our CI, go ahead then. But please, let's do one release of `polyfill-php72` as a no-op package as I proposed in https://github.com/symfony/polyfill/pull/492#discussion_r1749948439

> PHP 8.1 is too early IMHO, at least I don't see similar arguments.

Let's revisit that topic in December or January. 🙂

nicolas-grekas:
> let's do one release of polyfill-php72 as a no-op package

why did we do this? what's the benefit?
people with php < 7.2 can still install a previous version

derrabus:
> > let's do one release of polyfill-php72 as a no-op package
> 
> why did we do this? what's the benefit?

If you're on PHP >= 7.2, but one of your dependency's dependencies requires the polyfill, you get a package with zero overhead. Once we tagged a new release this way, we can still remove the package from the monorepo.

nicolas-grekas:
Got it, thanks. PR updated.

derrabus (approved):

PR #494, opened 2024-09-09
Update unicode data

nicolas-grekas:


phpunit
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env php
+<?php
+
+if (!file_exists(__DIR__.'/vendor/symfony/phpunit-bridge/bin/simple-phpunit')) {
+    echo "Unable to find the `simple-phpunit` script in `vendor/symfony/phpunit-bridge/bin/`.\nPlease run `composer update` before running this command.\n";
```
stof: `fwrite(STDERR, "...");` would be better by writing it to the right stream (helps if someone pipes the phpunit output somewhere else)

stof (commented):
would it make sense to separate the test setup change and the unicode updates in 2 different PRs to make the diff easier to understand ?

nicolas-grekas:
@TRowbotham would you be up to give me some help on the IDN topic again?
The test suite we added in #267 makes maintaining the corresponding polyfill very hard because it's not possible to have the implementation in sync with the latest Unicode data.
Instead, I think we should have a limited set of data to test that works with all php/icu versions so that we can be reasonably confident about the polyfill without proving we're cutting edge data-wise.

phpunit
nicolas-grekas: I copy/pasted from symfony/symfony, this is good enough :)

TRowbotham:
> @TRowbotham would you be up to give me some help on the IDN topic again? The test suite we added in #267 makes maintaining the corresponding polyfill very hard because it's not possible to have the implementation in sync with the latest Unicode data. Instead, I think we should have a limited set of data to test that works with all php/icu versions so that we can be reasonably confident about the polyfill without proving we're cutting edge data-wise.

Using a small subset of test data makes sense given the challenges of the implementation and data may not be in sync. I can take a look over the weekend to see what subset might work for this

nicolas-grekas:
It looks like not doing anything might be the best course of action - IDN tests pass at the moment :)

stof:
Maybe we should have 2 sets of tests:

- a basic set of tests that can run on any php/icu versions
- the full tests running when the ICU version matches.

the CI would then force the version of ICU being used (similar to what the symfony/intl testsuite does)

PR #495, opened 2024-09-09
Give testing some love

nicolas-grekas:
`./phpunit` FTW + some goodies

### DIFF
--- a/src/Util/TestListenerTrait.php
+++ b/src/Util/TestListenerTrait.php
@@ -25,10 +25,6 @@ class TestListenerTrait
 
     public function startTestSuite($mainSuite)
     {
-        if (null !== self::$enabledPolyfills) {
-            return;
-        }
-        self::$enabledPolyfills = false;
         $warnings = [];
 
         foreach ($mainSuite->tests() as $suite) {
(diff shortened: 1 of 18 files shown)


src/Util/TestListenerTrait.php
```diff
@@ -123,6 +123,10 @@ function {$f['name']}{$f['signature']}
                     $polyfillSignature = ReflectionCaster::castFunctionAbstract(new \ReflectionFunction($testNamespace.'\\'.$f['name']), [], new Stub(), true);
                     $polyfillSignature = ReflectionCaster::getSignature($polyfillSignature);
 
+                    if ('mb_get_info' === $r->name && false === strpos($originalSignature, '|null') && false !== strpos($polyfillSignature, '|null')) {
```
stof, line 122: can you add a comment about which PHP versions are concerned by this signature patch, to allow cleaning it in the future ?

nicolas-grekas: comment added

nicolas-grekas:
ValueError on PHP 8.4 is a regression, reported at https://github.com/php/php-src/issues/15824

nicolas-grekas:
Tests fixed, they show some issues we now can wonder about.
Help wanted @IonBazan @xabbuh @alexandre-daubois or anyone else really :)

PR #496, opened 2024-09-09
Bump to PHP 8.1

derrabus:
Active support for Symfony 5.4 will end in November. After that we won't need to maintain codebases anymore that support PHP 7. This repository is full of code that targets PHP 7 exclusively which is why I believe it's worth considering that we drop PHP 7 support here as well as soon as the maintenance of Symfony 5.4 is over.

This PR demonstrates the amount of code that we could get rid of. I'm flagging the PR as draft for the time being because I believe we don't want to merge it before November.


stof:
The Symfony 5.4 EOL is only in 1 year. Making the 5.4 LTS depend on an unmaintained version of polyfills looks wrong to me.
I think the maintenance cost of the polyfills for PHP 7 is manageable for now, so we might want to delay this PR until next year when 5.4 is fully EOL. This could be less work than creating a LTS version for polyfills.

derrabus:
> The Symfony 5.4 EOL is only in 1 year.

The active support for PHP 5.4 ends this year in November. Beyond that is pure security fixing and all non-critical bugs become permanent features.

> Making the 5.4 LTS depend on an unmaintained version of polyfills looks wrong to me.

I'm happy to provide equal level of support to the polyfills which would mean we backport security fixes to the last tag that was compatible with PHP 7. I don't expect much to come here.

> I think the maintenance cost of the polyfills for PHP 7 is manageable for now

I mean, look at all the code I'm deleting here. Managable? Yes, but managaging this stuff comes with a price tag.

stof:
> I mean, look at all the code I'm deleting here. Managable? Yes, but managaging this stuff comes with a price tag.

How often have you *modified* that code in the last year ?

derrabus:
> How often have you _modified_ that code in the last year ?

It's more like, how often do we add new polyfills and have to jump through more hoops to get the support PHP 7 right. See #499/#501: We're fixing new polyfills for PHP 7 here that Symfony 5 itself does not even use. That's fine now, but once Symfony itself does not support PHP 7 anymore, this really feels like a pointless endeavor.

nicolas-grekas:
Let me close as I don't think we'd want to bump existing polyfills.
What could make sense is bumping new ones only.
This could even start with the PHP 8.5 one.

derrabus:
> Let me close as I don't think we'd want to bump existing polyfills.

… like never? I think that's a mistake. Really, what's the point in maintaining a polyfill for ext-mbstring that supports PHP 7 forever? That code is already complicated and it will get more complicated with every PHP version that adds new functions to mbstring.

PR #497, opened 2024-09-11
Various fixes and removed deprecations

alexandre-daubois:
First batch of fixes with the "new" CI.

### DIFF
--- a/src/Iconv/Iconv.php
+++ b/src/Iconv/Iconv.php
@@ -430,32 +430,6 @@ final class Iconv
 
     public static function iconv_strlen($s, $encoding = null)
     {
-        static $hasXml = null;
-        if (null === $hasXml) {
-            $hasXml = \extension_loaded('xml');
-        }
-
-        if ($hasXml) {
-            return self::strlen1($s, $encoding);
-        }
-
-        return self::strlen2($s, $encoding);
-    }
-
-    public static function strlen1($s, $encoding = null)
-    {
-        if (null === $encoding) {
-            $encoding = self::$internalEncoding;
-        }
-        if (0 !== stripos($encoding, 'utf-8') && false === $s = self::iconv($encoding, 'utf-8', $s)) {
-            return false;
-        }
-
-        return \strlen(utf8_decode($s));
-    }
-
-    public static function strlen2($s, $encoding = null)
-    {
         if (null === $encoding) {
             $encoding = self::$internalEncoding;
         }
--- a/tests/Mbstring/MbstringTest.php
+++ b/tests/Mbstring/MbstringTest.php
@@ -68,13 +68,21 @@ class MbstringTest extends TestCase
      */
     public function testConvertEncoding()
     {
-        $this->assertSame(utf8_decode('déjà'), mb_convert_encoding('déjà', 'Windows-1252'));
+        $this->assertSame(iconv('UTF-8', 'ISO-8859-1', 'déjà'), mb_convert_encoding('déjà', 'Windows-1252'));
+        $this->assertSame('déjà', mb_convert_encoding(mb_convert_encoding('déjà', 'ISO-8859-1', 'UTF-8'), 'Utf-8', 'ASCII,ISO-2022-JP,UTF-8,ISO-8859-1'));
+        $this->assertSame('déjà', mb_convert_encoding(mb_convert_encoding('déjà', 'ISO-8859-1', 'UTF-8'), 'Utf-8', ['ASCII', 'ISO-2022-JP', 'UTF-8', 'ISO-8859-1']));
+    }
+
+    /**
+     * @group legacy
+     */
+    public function testConvertLegacyEncoding()
+    {
+        // handling base64 and html entities with mb_convert_encoding is deprecated in PHP 8.2
         $this->assertSame(base64_encode('déjà'), mb_convert_encoding('déjà', 'Base64'));
         $this->assertSame('&#23455;<&>d&eacute;j&agrave;', mb_convert_encoding('実<&>déjà', 'Html-entities'));
         $this->assertSame('déjà', mb_convert_encoding(base64_encode('déjà'), 'Utf-8', 'Base64'));
         $this->assertSame('déjà', mb_convert_encoding('d&eacute;j&#224;', 'Utf-8', 'Html-entities'));
-        $this->assertSame('déjà', mb_convert_encoding(utf8_decode('déjà'), 'Utf-8', 'ASCII,ISO-2022-JP,UTF-8,ISO-8859-1'));
-        $this->assertSame('déjà', mb_convert_encoding(utf8_decode('déjà'), 'Utf-8', ['ASCII', 'ISO-2022-JP', 'UTF-8', 'ISO-8859-1']));
     }
 
     /**
@@ -567,7 +575,7 @@ class MbstringTest extends TestCase
     {
         $this->assertSame(3, mb_strwidth("\000実", 'UTF-8'));
         $this->assertSame(4, mb_strwidth('déjà', 'UTF-8'));
-        $this->assertSame(4, mb_strwidth(utf8_decode('déjà'), 'CP1252'));
+        $this->assertSame(4, mb_strwidth(mb_convert_encoding('déjà', 'ISO-8859-1', 'UTF-8'), 'CP1252'));
     }
 
     /**
--- a/tests/Php81/Php81Test.php
+++ b/tests/Php81/Php81Test.php
@@ -34,6 +34,7 @@ class Php81Test extends TestCase
 
     /**
      * @requires extension mysqli
+     * @group legacy
      */
     public function testMysqliRefreshReplicaDefined()
     {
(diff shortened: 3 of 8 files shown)


tests/Mbstring/MbstringTest.php
```diff
@@ -68,13 +68,18 @@ public function testInternalEncodingWithInvalidEncoding()
...
-        $this->assertSame('&#23455;<&>d&eacute;j&agrave;', mb_convert_encoding('実<&>déjà', 'Html-entities'));
-        $this->assertSame('déjà', mb_convert_encoding(base64_encode('déjà'), 'Utf-8', 'Base64'));
-        $this->assertSame('déjà', mb_convert_encoding('d&eacute;j&#224;', 'Utf-8', 'Html-entities'));
-        $this->assertSame('déjà', mb_convert_encoding(utf8_decode('déjà'), 'Utf-8', 'ASCII,ISO-2022-JP,UTF-8,ISO-8859-1'));
-        $this->assertSame('déjà', mb_convert_encoding(utf8_decode('déjà'), 'Utf-8', ['ASCII', 'ISO-2022-JP', 'UTF-8', 'ISO-8859-1']));
+        $this->assertSame(mb_convert_encoding('déjà', 'ISO-8859-1', 'UTF-8'), mb_convert_encoding('déjà', 'Windows-1252'));
+        $this->assertSame('déjà', mb_convert_encoding(mb_convert_encoding('déjà', 'ISO-8859-1', 'UTF-8'), 'Utf-8', 'ASCII,ISO-2022-JP,UTF-8,ISO-8859-1'));
+        $this->assertSame('déjà', mb_convert_encoding(mb_convert_encoding('déjà', 'ISO-8859-1', 'UTF-8'), 'Utf-8', ['ASCII', 'ISO-2022-JP', 'UTF-8', 'ISO-8859-1']));
+
+        if (\PHP_VERSION_ID < 80200) {
```
nicolas-grekas: maybe putting this in a dedicated `@group legacy` would work?

derrabus:
> There's one `utf8_decode()` occurrence left in Iconv. The recommended replacement `mb_convert_encoding($string, 'ISO-8859-1', 'UTF-8')` doesn't work in this case, so it deserves a bit more research. However, this fixes some problems here and there already.

We used to have a polyfill for `utf8_decode()` in the now-removed PHP 7.2 polyfill package. Would it help if we copied the code of that polyfill over to the iconv package?

tests/Mbstring/MbstringTest.php
alexandre-daubois: It does. Thanks!

alexandre-daubois:
Indeed @derrabus, that's smart. I copied the function and added it to Iconv as a private method. Seems to work really well 👍

src/Intl/Idn/bootstrap80.php
```diff
@@ -118,8 +118,8 @@
 }
 
 if (!function_exists('idn_to_ascii')) {
-    function idn_to_ascii(?string $domain, ?int $flags = IDNA_DEFAULT, ?int $variant = INTL_IDNA_VARIANT_UTS46, &$idna_info = null): string|false { return p\Idn::idn_to_ascii((string) $domain, (int) $flags, (int) $variant, $idna_info); }
+    function idn_to_ascii(string $domain, int $flags = IDNA_DEFAULT, int $variant = INTL_IDNA_VARIANT_UTS46, &$idna_info = null): string|false { return p\Idn::idn_to_ascii($domain, $flags, $variant, $idna_info); }
 }
 if (!function_exists('idn_to_utf8')) {
-    function idn_to_utf8(?string $domain, ?int $flags = IDNA_DEFAULT, ?int $variant = INTL_IDNA_VARIANT_UTS46, &$idna_info = null): string|false { return p\Idn::idn_to_utf8((string) $domain, (int) $flags, (int) $variant, $idna_info); }
+    function idn_to_utf8(string $domain, int $flags = IDNA_DEFAULT, int $variant = INTL_IDNA_VARIANT_UTS46, &$idna_info = null): string|false { return p\Idn::idn_to_utf8($domain, $flags, $variant, $idna_info); }
```
stof: Those changes should be reverted, as it makes it incompatible with passing `null` in non-strict-types files (which is deprecated in PHP 8.1+ but will only be removed in PHP 9.0)

src/Iconv/Iconv.php
```diff
@@ -741,4 +741,34 @@ private static function getData($file)
 
         return false;
     }
+
+    private static function utf8Decode($s)
```
stof: I suggest naming `utf8ToLatin1` instead, to reflect what this method actually does (the bad name of `utf8_decode` was the main reason why it has been deprecated)

stof: Alternatively, we could drop `strlen1` entirely, falling back directly to the `strlen2` implementation that was used when the `xml` extension was not loaded and so we did not have the native `utf8_decode` function. The `utf8_decode` fallback was done because using the native function was faster than the userland logic of `strlen2`, but this probably does not hold anymore with a userland function doing more work.

tests/Php81/Php81Test.php
```diff
@@ -34,6 +34,7 @@ public function testArrayIsList()
 
     /**
      * @requires extension mysqli
+     * @requires PHP < 8.4
```
derrabus: This test should be executed on PHP 8.4 as well. We should rather allow the test to trigger deprecations.

alexandre-daubois: Right

src/Intl/Idn/bootstrap80.php
alexandre-daubois: I reverted the casts

src/Iconv/Iconv.php
alexandre-daubois: I went for the alternative solution, which really clean things up.

tests/Php82/SensitiveParameterValuePolyfillTest.php
```diff
@@ -17,7 +17,7 @@
 /**
  * @author Tim Düsterhus <duesterhus@woltlab.com>
  */
-class SensitiveParameterValueTest extends TestCase
+class SensitiveParameterValuePolyfillTest extends TestCase
```
alexandre-daubois, line 20: TestListenerTrait tries to find a polyfill declaration in `bootstrap.php` base on the test classname. However, `SensitiveParameterValue` is a "special" polyfill that doesn't need any declaration in `bootstrap.php`.

So, PHPUnit warned that no polyfill was found, but that's normal. By renaming, the file to not match the `SensitiveParametrerValue` classname, we avoid `TestListenerTrait` to raise a wrong warning.

src/Intl/Idn/bootstrap80.php
stof: You need to revert the whole change. The most important part of my comment is about the parameter type.
PHP does not support passing `null` to a non-nullable *userland* function parameter. The reason of the PHP 8.1 deprecation is precisely because the behavior of native functions cannot be emulated exactly in userland for now (in PHP 9, they will behave the same than userland functions in that regard)

stof: @nicolas-grekas what do you think about adding legacy tests covering passing `null` to all our polyfilled functions (that will have a non-nullable parameter type in PHP 9), so that the testsuite catches such broken changes to the polyfills, as that's not the first time we get such changes in a PR.

alexandre-daubois: I see what you mean, but what about [this warning](https://github.com/symfony/polyfill/actions/runs/10795014872/job/29940652441#step:5:733)? This is the reason I updated the signature

stof: @alexandre-daubois this should have been fixed by https://github.com/symfony/polyfill/pull/495 AFAIK

alexandre-daubois: It doesn't seem to be touching anything related to the `Incompatible signature for PHP >= 8` warning. Am I missing something?

stof: it touches the detection logic at https://github.com/symfony/polyfill/pull/495/files#diff-05ed2f919a0a2d68912601230246f2c39077cd90c3ac9df8ac74651dd7c197aeR140

alexandre-daubois: Oh right I get it now. That wasn't the root cause of the warning message. That's reverted 👍

tests/Mbstring/MbstringTest.php
```diff
@@ -68,13 +68,21 @@ public function testInternalEncodingWithInvalidEncoding()
      */
     public function testConvertEncoding()
     {
-        $this->assertSame(utf8_decode('déjà'), mb_convert_encoding('déjà', 'Windows-1252'));
+        $this->assertSame(mb_convert_encoding('déjà', 'ISO-8859-1', 'UTF-8'), mb_convert_encoding('déjà', 'Windows-1252'));
```
xabbuh: Should we really replace `utf8_decode()` with `mb_convert_encoding()` here as `mb_convert_encoding()` is the polyfill that we test here, right?

alexandre-daubois: That's right. I moved the test case in a separate method to still use `utf8_decode()`, marked as `@group legacy`.

stof: Or use `iconv`

alexandre-daubois: Suits me as well, PR updated

xabbuh (approved):

nicolas-grekas:
Thank you @alexandre-daubois.

PR #498, opened 2024-09-16
Introduce PHP 8.5 polyfill

alexandre-daubois:
⚠️ This is a first, naive draft of a new always-available `ext-uri` introduced in PHP 8.5 (or 9.0, whichever comes first). It isn't complete yet.

Follow-up of https://github.com/symfony/symfony/pull/57192#issuecomment-2340133849

I used TDD on this one, using https://github.com/php/php-src/pull/14461 test cases. There's a lot to do still. I'm publishing this PR if one wants to track progress on this, but also to get potential feedback during development. Keep in mind that "everything" is subject to change if you take a look at it 😄 

RFC: https://wiki.php.net/rfc/url_parsing_api
Implementation: https://github.com/php/php-src/pull/14461


src/Php85/Exception/ParsingException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Symfony\Polyfill\Php85\Exception;
+
+class ParsingException extends \RuntimeException
+{
+}
```
OskarStark: newline missing

src/Php85/Uri/Rfc3986Uri.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->scheme = $matches['scheme'] ?? null;
+        $this->user = isset($matches['user']) ? rawurldecode($matches['user']) : null;
+        $this->password = isset($matches['pass']) ? rawurldecode($matches['pass']) : null;
+        $this->host = $matches['host'] ?? null;
+        $this->port = $matches['port'] ?? null;
+        $this->path = isset($matches['path']) ? ltrim($matches['path'], '/') : null;
+        $this->query = $matches['query'] ?? null;
+        $this->fragment = $matches['fragment'] ?? null;
+    }
+}
```
OskarStark: newline missing

tests/Php85/Rfc3986UriTest.php
```diff
@@ -0,0 +1,122 @@
...
+
+    public function testWithBaseUriAndRelativeUrl()
+    {
+        $uri = Uri::fromRfc3986("/path/to/file2", "https://test.com");
+
+        $this->assertSame('https', $uri->getScheme());
+        $this->assertSame('test.com', $uri->getHost());
+        $this->assertSame('path/to/file2', $uri->getPath());
+    }
+}
```
OskarStark: newline missing

tests/Php85/WhatWgUriTest.php
```diff
@@ -0,0 +1,133 @@
...
+
+    public function testWithBaseUriAndRelativeUrl()
+    {
+        $uri = Uri::fromWhatWg("/path/to/file2", "https://test.com");
+
+        $this->assertSame('https', $uri->getScheme());
+        $this->assertSame('test.com', $uri->getHost());
+        $this->assertSame('path/to/file2', $uri->getPath());
+    }
+}
```
OskarStark: newline missing

src/Php85/Uri/Uri.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public function getFragment(): ?string
+    {
+        return $this->fragment;
+    }
+
+    public function __toString(): string
+    {
+
```
OskarStark: ?

src/Php85/Uri/Rfc3986Uri.php
```diff
@@ -0,0 +1,81 @@
...
+namespace Symfony\Polyfill\Php85\Uri;
+
+use Symfony\Polyfill\Php85\Exception\ParsingException;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ *
+ * @internal
+ */
+class Rfc3986Uri extends Uri
```
stof: This needs to extend `Uri\Uri` instead of extending the stub abstract class directly. Otherwise, the stub `Uri\Rfc3986Uri` class will not extend `Uri\Uri`

src/Php85/Php85.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+namespace Symfony\Polyfill\Php85;
+
+/**
+ * @author Alexandre Daubois <alex.daubois@gmail.com>
+ *
+ * @internal
+ */
+final class Php85
```
derrabus: I would not add this class until we need it.

src/Php85/Resources/stubs/Uri/Rfc3986Uri.php
```diff
@@ -0,0 +1,18 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Uri;
+
+if (\PHP_VERSION_ID < 80500) {
+    class Rfc3986Uri extends \Symfony\Polyfill\Php85\Uri\Rfc3986Uri
```
derrabus, line 15: The classes in core don't have a parent. Shall we alias the class instead?

src/Php85/Uri/Rfc3986Uri.php
```diff
@@ -0,0 +1,81 @@
...
+{
+    public function __construct(string $uri, ?string $baseUrl = null)
+    {
+        if ('' === trim($uri)) {
+            throw new \ValueError('Argument #1 ($uri) cannot be empty');
```
derrabus, line 17: The PHP 8.5 polyfill needs to either depend on the PHP 8.0 polyfill or on PHP 8.0 itself. Otherwise, the `ValueError` class might not be available on PHP 7.

See also the discussions on #499 and #496.

derrabus (changes requested):

src/Php85/Uri/Uri.php
alexandre-daubois: Yet to come 😄

src/Php85/Resources/stubs/Uri/Rfc3986Uri.php
alexandre-daubois: I can see `class Rfc3986Uri extends Uri` in the RFC. Do I miss your point?

src/Php85/Uri/Rfc3986Uri.php
alexandre-daubois: Indeed, thank you! I added the polyfill for 8.0, seems the best option to me.

alexandre-daubois:
Thank you everyone for reviewing this first draft! I applied the requested changes. I'll continue the work soon.

src/Php85/Resources/stubs/Uri/Rfc3986Uri.php
derrabus: Yes, but `Symfony\Polyfill\Php85\Uri\Rfc3986Uri` does not appear in the original inheritance chain. What is the point of the Symfony namespaced classes anyway? Do we need them?

src/Php85/Uri/Rfc3986Uri.php
derrabus: Yeah, it's probably the best for the time being. I'm leaning a bit towards making PHP 8.1 a requirement for the PHP 8.5 polyfill package, but we don't have to decide this now.

src/Php85/Resources/stubs/Uri/Rfc3986Uri.php
stof: Having all the logic being in internal namespaced classes allows to test the logic of those classes even on versions where the polyfill is not applied because the native class exist. That's why all our polyfill classes have such internal parent class.

TimWolla:
see https://github.com/bakame-php/aide-uri for a complete userland implementation of the RFC.

/cc @nyamsprod

nicolas-grekas:
Let me close here, thanks for proposing.
This topic is solved either by using bakame-php/aide-uri directly or by porting it here. I'm looking forward to @nyamsprod opinion of the topic.

PR #500, opened 2024-09-16
drop utf8_decode()

xabbuh:


stof:
@xabbuh isn't this a duplicate of #497 (at least in the goal of the PR) ?

xabbuh:
indeed, I missed that one

PR #501, opened 2024-09-17
mbstring polyfills must not raise value errors in PHP 7

### LINKED ISSUE #499: `ValueError` class not available in PHP < 8
The MBString polyfill supports PHP >= 7.2 but it makes use of the `ValueError` exception class, which was only introduced in PHP 8.

In a lot of instances, there is a version check before it is used, and in these cases there is no problem.

However, there are three unguarded locations that refer to the class, which will result in fatal errors on PHP 7:

https://github.com/symfony/polyfill/blob/731f6e101c33d3d1c2e4a8a87e4e4fe9b2f50f72/src/Mbstring/Mbstring.php#L837-L841

https://github.com/symfony/polyfill/blob/731f6e101c33d3d1c2e4a8a87e4e4fe9b2f50f72/src/Mbstring/Mbstring.php#L849-L851

https://github.com/symfony/polyfill/blob/731f6e101c33d3d1c2e4a8a87e4e4fe9b2f50f72/src/Mbstring/Mbstring.php#L1040-L1043

derrabus:
Fixes #499.

This PR changes all new mbstring polyfills on PHP 7 so that they trigger an oldschool PHP warning instead of raising a `ValueError`. The reason for that is that the `ValueError` class might not be available on PHP 7 and the behavior is more consitent with the other polyfills of the mbstring extension.

The somewhat weird side-effect is that we now polyfill a behavior that was never implemented in PHP.

### DIFF
--- a/src/Mbstring/bootstrap.php
+++ b/src/Mbstring/bootstrap.php
@@ -133,27 +133,27 @@ if (!function_exists('mb_str_split')) {
 }
 
 if (!function_exists('mb_str_pad')) {
-    function mb_str_pad(string $string, int $length, string $pad_string = ' ', int $pad_type = STR_PAD_RIGHT, ?string $encoding = null): string { return p\Mbstring::mb_str_pad($string, $length, $pad_string, $pad_type, $encoding); }
+    function mb_str_pad(string $string, int $length, string $pad_string = ' ', int $pad_type = STR_PAD_RIGHT, ?string $encoding = null) { return p\Mbstring::mb_str_pad($string, $length, $pad_string, $pad_type, $encoding); }
 }
 
 if (!function_exists('mb_ucfirst')) {
-    function mb_ucfirst(string $string, ?string $encoding = null): string { return p\Mbstring::mb_ucfirst($string, $encoding); }
+    function mb_ucfirst(string $string, ?string $encoding = null) { return p\Mbstring::mb_ucfirst($string, $encoding); }
 }
 
 if (!function_exists('mb_lcfirst')) {
-    function mb_lcfirst(string $string, ?string $encoding = null): string { return p\Mbstring::mb_lcfirst($string, $encoding); }
+    function mb_lcfirst(string $string, ?string $encoding = null) { return p\Mbstring::mb_lcfirst($string, $encoding); }
 }
 
 if (!function_exists('mb_trim')) {
-    function mb_trim(string $string, ?string $characters = null, ?string $encoding = null): string { return p\Mbstring::mb_trim($string, $characters, $encoding); }
+    function mb_trim(string $string, ?string $characters = null, ?string $encoding = null) { return p\Mbstring::mb_trim($string, $characters, $encoding); }
 }
 
 if (!function_exists('mb_ltrim')) {
-    function mb_ltrim(string $string, ?string $characters = null, ?string $encoding = null): string { return p\Mbstring::mb_ltrim($string, $characters, $encoding); }
+    function mb_ltrim(string $string, ?string $characters = null, ?string $encoding = null) { return p\Mbstring::mb_ltrim($string, $characters, $encoding); }
 }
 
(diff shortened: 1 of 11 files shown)


tests/Php84/Php84Test.php
```diff
@@ -229,7 +229,12 @@ public function testMbRTrim(string $expected, string $string, ?string $character
 
     public function testMbTrimException()
     {
-        $this->expectException(\ValueError::class);
+        if (PHP_VERSION_ID < 80000) {
+            $this->expectException(\ValueError::class);
+        } else {
+            $this->expectWarning();
+        }
```
jdreesen: Shouldn't it be the other way around?
```suggestion
        if (PHP_VERSION_ID < 80000) {
            $this->expectWarning();
        } else {
            $this->expectException(\ValueError::class);
        }
```

derrabus: Yes. I just caught this as well, but you were a bit faster. 🙂

src/Mbstring/bootstrap.php
```diff
@@ -133,27 +133,27 @@ function mb_str_split($string, $length = 1, $encoding = null) { return p\Mbstrin
 }
 
 if (!function_exists('mb_str_pad')) {
-    function mb_str_pad(string $string, int $length, string $pad_string = ' ', int $pad_type = STR_PAD_RIGHT, ?string $encoding = null): string { return p\Mbstring::mb_str_pad($string, $length, $pad_string, $pad_type, $encoding); }
```
nicolas-grekas, line 136: what about a bootstrap80.php file to add the return types when running on PHP8+?

derrabus: That file already exists. ✌🏻

nicolas-grekas: ah indeed :) 
there are some failures to fix ;)

derrabus: Can you help we with those? Apparently, the test listener is skipping tests at a strange point in time which conflicts with the `expectWarning()` feature of PHPUnit. How do we solve that usually?

nicolas-grekas:
Let's also address #506 here?

nicolas-grekas:
Thank you @derrabus.

eexit:
Hello,

I just updated my codebase with the latest changes, and my PHPStan breaks:

```php
\PHPStan\dumpType(mb_trim('foo'));
// Dumped type: mixed
```

```bash
$ composer info symfony/polyfill-php84
name     : symfony/polyfill-php84
descrip. : Symfony polyfill backporting some PHP 8.4+ features to lower PHP versions
keywords : compatibility, polyfill, portable, shim
versions : * v1.34.0
released : 2026-04-10, this week
type     : library
license  : MIT License (MIT) (OSI approved) https://spdx.org/licenses/MIT.html#licenseText
homepage : https://symfony.com
source   : [git] https://github.com/symfony/polyfill-php84.git 88486db2c389b290bf87ff1de7ebc1e13e42bb06
dist     : [zip] https://api.github.com/repos/symfony/polyfill-php84/zipball/88486db2c389b290bf87ff1de7ebc1e13e42bb06 88486db2c389b290bf87ff1de7ebc1e13e42bb06
path     : /opt/php/vendor/symfony/polyfill-php84
names    : symfony/polyfill-php84

$ php -v
PHP 8.3.30 (cli) (built: Jan 30 2026 01:25:40) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.3.30, Copyright (c) Zend Technologies
    with Zend OPcache v8.3.30, Copyright (c), by Zend Technologies
    with Xdebug v3.5.1, Copyright (c) 2002-2026, by Derick Rethans
```

Is there a way to fix this?

Thanks!

nicolas-grekas:
Please open an issue. Commenting on a closed PR/issue is a good way to not get attention.

PR #503, opened 2024-09-24
[PHP 8.4] Add `bcdivmod`

Ayesh:
https://wiki.php.net/rfc/add_bcdivmod_to_bcmath

### DIFF
--- a/src/Php84/Php84.php
+++ b/src/Php84/Php84.php
@@ -169,4 +169,23 @@ final class Php84
 
         return mb_convert_encoding($string, $encoding, 'UTF-8');
     }
+
+    public static function bcdivmod(string $num1, string $num2, ?int $scale = null): array {
+        if ($num2 === '0') {
+            throw new \DivisionByZeroError('Division by zero');
+        }
+
+        if (!is_numeric($num1)) {
+            throw new \ValueError('Argument #1 ($num1) is not well-formed');
+        }
+
+        if (!is_numeric($num2)) {
+            throw new \ValueError('Argument #2 ($num2) is not well-formed');
+        }
+
+        return [
+            \bcdiv($num1, $num2, 0),
+            \bcmod($num1, $num2, $scale),
+        ];
+    }
 }
--- a/src/Php84/bootstrap.php
+++ b/src/Php84/bootstrap.php
@@ -60,3 +60,7 @@ if (extension_loaded('mbstring')) {
         function mb_rtrim(string $string, ?string $characters = null, ?string $encoding = null): string { return p\Php84::mb_rtrim($string, $characters, $encoding); }
     }
 }
+
+if (function_exists('bcdiv')) {
+    function bcdivmod(string $num1, string $num2, ?int $scale = null): array {return p\Php84::bcdivmod($num1, $num2, $scale); }
+}
--- a/tests/Php84/Php84Test.php
+++ b/tests/Php84/Php84Test.php
@@ -256,6 +256,16 @@ class Php84Test extends TestCase
         mb_internal_encoding($old);
     }
 
+    /**
+     * @covers \Symfony\Polyfill\Php84\Php84::bcdivmod
+     *
+     * @dataProvider bcDivModProvider
+     */
+    public function testBcDivMod(string $num1, string $num2, ?int $scale = null, array $expected): void
+    {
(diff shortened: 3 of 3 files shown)


src/Php84/bootstrap.php
```diff
@@ -60,3 +60,7 @@ function mb_ltrim(string $string, ?string $characters = null, ?string $encoding
         function mb_rtrim(string $string, ?string $characters = null, ?string $encoding = null): string { return p\Php84::mb_rtrim($string, $characters, $encoding); }
     }
 }
+
+if (function_exists('bcdiv')) {
+    function bcdivmod(string $num1, string $num2, ?int $scale = null): array {return p\Php84::bcdivmod($num1, $num2, $scale); }
```
stof, line 65: such polyfill function should be defined only when the bcmath extension is loaded (similar to how we deal with mbstring new functions just above)

tests/Php84/Php84Test.php
```diff
@@ -256,6 +256,16 @@ public function testMbTrimCharactersEncoding()
         mb_internal_encoding($old);
     }
 
+    /**
+     * @covers \Symfony\Polyfill\Php84\Php84::bcdivmod
+     *
+     * @dataProvider bcDivModProvider
+     */
+    public function testBcDivMod(string $num1, string $num2, ?int $scale = null, array $expected): void
```
stof, line 264: this test needs `@requires extension bcmath` as the implementation would break without that extension

src/Php84/bootstrap.php
Ayesh: I was thinking the `function_exists('bcdiv')` call will check for this, and will also play nicely with any potential polyfills that may declare a `bcdiv` function. If you think, I can update this with a full `extension_loaded` call.

tests/Php84/Php84Test.php
Ayesh: Brilliant, thank you.

src/Php84/bootstrap.php
stof: oh, I missed that this was checking `bcdiv` and not `bcdivmod`. But then, it misses the check for `bcdivmod` (all polyfilled functions need to be guarded to avoid double function definition).

and for the detection of bcmath, please use `extension_loaded` for consistency with our other polyfills. Thus, a `extension_loaded` call is fully resolved at compile-time by OPCache.

src/Php84/Php84.php
```diff
@@ -169,4 +169,23 @@ private static function mb_internal_trim(string $regex, string $string, ?string
...
+            throw new \DivisionByZeroError('Division by zero');
+        }
+
+        if (!is_numeric($num1)) {
+            throw new \ValueError('Argument #1 ($num1) is not well-formed');
+        }
+
+        if (!is_numeric($num2)) {
+            throw new \ValueError('Argument #2 ($num2) is not well-formed');
+        }
```
derrabus, line 184: Those two errors are not available on PHP 7. I suggest to emit a `E_USER_WARNING` and return `null` instead if we're on PHP < 8.

tests/Php84/Php84Test.php
```diff
@@ -319,4 +329,9 @@ public static function mbRTrimProvider(): iterable
 
         yield ["foo\n", "foo\n", 'o'];
     }
+
+    public static function bcDivModProvider(): iterable
+    {
+        yield ['10', '10', null, ['1', '0']];
```
derrabus, line 335: A couple more tests would be nice. You're only testing the simplest path.

Also, beware that by setting the scale to `null`, your test depends on the global `bcmath.scale` setting. Currently, the test will fail if it's run in an environment where that setting is not `0`.

Ayesh:
Thanks a lot for the reviews @stof @derrabus, really appreciate them, genuinely helpful ones. 
I will flesh out the polyfill and the tests soon.

stof:
@Ayesh can you update your PR ?

stof:
@Ayesh do you have any plan to finish this PR ?

alexandre-daubois:
@Ayesh Do you wish to finish this PR? Otherwise, please let me know if I can take it over

Ayesh:
Hi @alexandre-daubois - please feel free to take over, I got a bit too much in my plate for the coming weeks, unfortunately.

alexandre-daubois:
Closing in favor of #519, thanks

PR #504, opened 2024-09-25
Ctype: Give the correct deprecations for PHP 8.1+

BackEndTea:
This deprecation isn't just for integers that are outside the integer range of -128, 255.

https://3v4l.org/PgZ6c

### DIFF
--- a/src/Ctype/Ctype.php
+++ b/src/Ctype/Ctype.php
@@ -211,16 +211,16 @@ final class Ctype
      */
     private static function convert_int_to_char_for_ctype($int, $function)
     {
+        if (\PHP_VERSION_ID >= 80100 && !\is_string($int)) {
+            @trigger_error($function.'(): Argument of type '.get_debug_type($int).' will be interpreted as string in the future', \E_USER_DEPRECATED);
+        }
+        
         if (!\is_int($int)) {
             return $int;
         }
 
         if ($int < -128 || $int > 255) {
             return (string) $int;
-        }
-
-        if (\PHP_VERSION_ID >= 80100) {
-            @trigger_error($function.'(): Argument of type int will be interpreted as string in the future', \E_USER_DEPRECATED);
         }
 
         if ($int < 0) {


src/Ctype/Ctype.php
```diff
@@ -219,14 +229,21 @@ private static function convert_int_to_char_for_ctype($int, $function)
...
         }
 
         return \chr($int);
     }
+
+    /**
+     * @param mixed $input
+     * @param string $function
+     */
+    public static function checkType($input, $function)
```
derrabus: Why is this method public?

```suggestion
     */
    private static function checkType($input, string $function): void
```

src/Ctype/composer.json
```diff
@@ -16,7 +16,8 @@
         }
     ],
     "require": {
-        "php": ">=7.2"
+        "php": ">=7.2",
+        "symfony/polyfill-php80": "^1.31"
```
stof: why are you adding this dependency ?

nicolas-grekas: friendly up @BackEndTea

BackEndTea: This is due to the usage of [`get_debug_type`](https://www.php.net/manual/en/function.get-debug-type.php), since that was only added in PHP 8.0

I could switch it out with a [`gettype`](https://www.php.net/manual/en/function.gettype.php) call if we are willing to accept a slight mismatch in error messages.

nicolas-grekas: let's use gettype yes

nicolas-grekas: wait: get_debug_type is behind a `PHP_VERSION_ID >= 80100` check so the polyfill is never needed

BackEndTea: Youre correct, i didnt realize that!

src/Ctype/Ctype.php
nicolas-grekas: can't we inline this method in convert_int_to_char_for_ctype?

nicolas-grekas:
Thank you @BackEndTea.

PR #508, opened 2024-11-22
Require iconv for mbstring

jaapio:
As the mbstring polyfill using the iconv functions a lot, if not in all calls being done, code without iconv using this polyfill will break. 

As this repo also provides a polyfill for iconv I don't think this will be a breaking change.

### DIFF
--- a/src/Mbstring/composer.json
+++ b/src/Mbstring/composer.json
@@ -16,7 +16,8 @@
         }
     ],
     "require": {
-        "php": ">=7.2"
+        "php": ">=7.2",
+        "ext-iconv": "*"
     },
     "provide": {
         "ext-mbstring": "*"


src/Mbstring/composer.json
```diff
@@ -16,7 +16,8 @@
         }
     ],
     "require": {
-        "php": ">=7.2"
+        "php": ">=7.2",
+        "ext-iconv"
```
nicolas-grekas: is that legal? should this be that?
```suggestion
        "ext-iconv": "*"
```

jaapio: Yes you are absolutely correct, I should have take more care before committing this.

stof:
the test failure is because one test is comparing the currencies in this polyfill with the ICU data available in symfony/intl, and symfony/intl has recently updated its data to ICU 76.1 which has added new currencies.

jaapio:
@stof I assume there is no need for me to look into this regarding this PR?

stof:
@jaapio you indeed don't need to care about the failures in the PHP 7.x jobs here. The jobs for 8.x have additional failures, but I don't think they are related to your change either.

stof (approved):

xabbuh (approved):

chalasr (approved):

stof:
Thank you @jaapio.

PR #515, opened 2025-02-19
Fix constants being redefined in PHP 8.4 polyfill

cs278:
Tracked down this error to this library:

```
In bootstrap.php line 19:

  [ErrorException]
  Constant CURL_HTTP_VERSION_3 already defined
```

It appears to me the standard is to conditionally define symbols so these two were missing a check. I've also added a basic test which should prevent this happening again.


derrabus (approved):

IonBazan (approved):

nicolas-grekas:
Thank you @cs278.

ro0NL:
@nicolas-grekas can you make a release? -.- #516

PR #520, opened 2025-03-26
Pin ICU version to prevent unexpected breakage

TRowbotham:
Pinning the ICU version will prevent unexpected breakage from the Unicode version changing before we update the data files and polyfills


TRowbotham:
Oops, I see someone beat me to this. Sorry for the noise.

PR #521, opened 2025-03-27
Add `get_*_handler` polyfills

Fan2Shrek:
Here is the polyfill of newly implemented `get_error_handler` and `get_exception_handler`.
I have created the default files for PHP 8.5 polyfills.

The tests are taken from php/php-src#17693, but some cannot be completed due to new PHP 8 syntax.

There are some issues with the error handler in the tests, I can't figure out why. If someone can help, that would be great! 😅 This seems to be related to the TestListenerTrait, I think.

[RFC](https://wiki.php.net/rfc/get-error-exception-handler)

### DIFF
--- a/tests/Php85/Php85Test.php
+++ b/tests/Php85/Php85Test.php
@@ -1,0 +1,92 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Tests\Php85;
+
+use PHPUnit\Framework\TestCase;
+
+class Php85Test extends TestCase
+{
+    /**
+    * @dataProvider provideHandler
+    */
+    public function testGetErrorHandler($expected, $handler): void
+    {
+        set_error_handler($handler);
+        try {
+            $result = get_error_handler();
+        } finally {
+            restore_error_handler();
+        }
+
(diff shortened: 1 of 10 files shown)


tests/Php85/Php85Test.php
```diff
@@ -0,0 +1,86 @@
...
+{
+    /**
+    * @dataProvider provideHandler
+    */
+    public function testGetErrorHandler($expected, $handler): void
+    {
+        set_error_handler($handler);
+        $result = get_error_handler();
+        restore_error_handler();
+        restore_error_handler();
```
Fan2Shrek: I'm not sure why I need to restore the error handler twice. This prevents the `THE ERROR HANDLER HAS CHANGED!` warning from PHPUnit, but somehow, it doesn't work for PHP 7.2 testing.

nicolas-grekas: would passing a callable ('var_dump') to set_error_handler instead of `null` fix this?

Fan2Shrek: This does not work either, it looks weird to me.

Fan2Shrek: I figured it out! This seems to be related to https://github.com/symfony/polyfill/blob/d139525b14e0605dc92e3354d38e588a7817d1b5/src/Util/TestListenerTrait.php#L112. 
Wrapping the function call in a try block works fine! :)

PhilDaiguille (approved):
LGTM

nicolas-grekas (approved):

nicolas-grekas:
Thank you @Fan2Shrek.

PR #522, opened 2025-04-05
Add #[NoDiscard] to the PHP 8.5 polyfill

schodemeiss:


src/Php85/Resources/stubs/NoDiscard.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+if (\PHP_VERSION_ID < 80500) {
+    #[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_FUNCTION)]
+    final class NoDiscard
+    {
+        public readonly ?string $message;
```
OskarStark: Readonly is not available for PHP 7.2

schodemeiss: Attribute's aren't available in PHP 7.2 at all. Nor are typed properties.

However, both are used in the Deprecated attribute polyfill (src/Php84/Resources/stubs/Deprecated.php) which I used as a reference for this PR.

Given the above, I'm not certain on how I should proceed.

OskarStark: Indeed, didn't thought about that aspect 👍🏻

GromNaN (approved):
Thanks, the stub is identical to the one provided in php-src: https://github.com/TimWolla/php-src/blob/d80682e753af13a89ad71ee78100ce44045f175d/Zend/zend_attributes.stub.php#L88-L97

schodemeiss:
> Thanks, the stub is identical to the one provided in php-src: https://github.com/TimWolla/php-src/blob/d80682e753af13a89ad71ee78100ce44045f175d/Zend/zend_attributes.stub.php#L88-L97

Indeed. I presumed that was the correct thing to do.

Do we know when this Polyfill will be published as non-dev. I noticed it's on Packagist, but not as 1.38 (or whatever the other polyfills are).

https://packagist.org/packages/symfony/polyfill-php85

src/Php85/Resources/stubs/NoDiscard.php
stof: Attributes won't ever be read on PHP 7.2 as there is no API for that in Reflection. In practice, a polyfill for a new attribute class needs to support PHP 8.0+

schodemeiss: @stof So remove the "readonly", as thats the only thing not supported by 8.0?

And we're happy to leave the Deprecated attribute with the readonly field from the Php84 folder? Or shall I scoop that up whilst I'm here?

fabpot (approved):

fabpot:
Thank you @schodemeiss.

schodemeiss:
> Thank you @schodemeiss.

You're welcome. When do the polyfill packages get officially published? I noticed that whilst php85 appears on Packagist, it's still on the dev branch.

fabpot:
> > Thank you @schodemeiss.
> 
> You're welcome. When do the polyfill packages get officially published? I noticed that whilst php85 appears on Packagist, it's still on the dev branch.

I've released v1.32.0 which includes the PHP 8.5 polyfill.

PR #524, opened 2025-05-04
[8.4] Add polyfill for `ReflectionConstant`

DanielEScherzer:


DanielEScherzer:
All failures and errors are unrelated in the PHPUnit tests (mostly mbstring and Grapheme related)

Ayesh:
Wow, I didn't think it will be possible to polyfill `ReflectionConstant`, but great job pulling it off!

src/Php84/Resources/stubs/ReflectionConstant.php
```diff
@@ -0,0 +1,164 @@
...
+        
+        private $value;
+
+        /** @var bool */
+        private $deprecated;
+
+        /** @var bool */
+        private $persistent;
+        
+        public function __construct( string $name )
```
xabbuh:
```suggestion
        public function __construct(string $name)
```

We don't add spaces after opening and before closing parentheses. Can you please update this everywhere?

src/Php84/Resources/stubs/ReflectionConstant.php
```diff
@@ -0,0 +1,164 @@
...
+                // Reasonable fallback
+                $result .= gettype( $this->value );
+            }
+            $result .= ' ';
+            $result .= $this->getName();
+            $result .= ' ] { ';
+            if ( is_array( $this->value ) ) {
+                $result .= 'Array';
+            } else {
+                $result .= (string)$this->value;
```
stof: what about objects that are not stringable ?

tests/Php84/ReflectionConstantTest.php
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    public function testClassConstant()
+    {
+        $this->expectException( ReflectionException::class );
+        $this->expectExceptionMessage(
+            'Constant "ReflectionClass::IS_DEPRECATED" does not exist'
+        );
+        // Confirm that this constant actually exists
+        $this->assertTrue( defined( 'ReflectionFunction::IS_DEPRECATED' ) );
```
stof: this assertion should come *before* the configuration of expecting an exception (as we don't want want to accept such exception during that statement)

tests/Php84/ReflectionConstantTest.php
```diff
@@ -0,0 +1,229 @@
...
+            'MT_RAND_PHP',
+            $constant->getShortName(),
+            'Short name'
+        );
+        $this->assertSame(
+            1,
+            $constant->getValue(),
+            'Value'
+        );
+        if ( version_compare( PHP_VERSION, '8.3.0', '>=' ) ) {
```
stof: Use a `\PHP_VERSION_ID` check instead, to be consistent with our usages in Symfony.

stof (changes requested):

src/Php84/Resources/stubs/ReflectionConstant.php
DanielEScherzer: Done here and everywhere else

src/Php84/Resources/stubs/ReflectionConstant.php
DanielEScherzer: Then an error is thrown - added test cases for a stringable and non-stringable object to confirm

tests/Php84/ReflectionConstantTest.php
DanielEScherzer: Done

tests/Php84/ReflectionConstantTest.php
DanielEScherzer: Done

DanielEScherzer:
@fabpot is there anything I can do to move this forward?

nicolas-grekas:
Thank you @DanielEScherzer.

PR #525, opened 2025-05-07
[PHP 8.5] Add `array_first` and `array_last`

Ayesh:
Adds polyfills for the PHP 8.5's `array_first` and `array_last` functions.

The polyfills would have been a bit cleaner if we used the `array_key_first` and `array_key_last` functions, but they were added in PHP 7.3, and are not available on PHP 7.2.

This adds the same tests from php-src patch.

 - [RFC: `array_first()` and `array_last()`](https://wiki.php.net/rfc/array_first_last)
 - [php-src commit](https://github.com/php/php-src/commit/168343d2e811fe0f2152d39a8f88557cc0b35c86)
 - [PHP.Watch polyfill for PHP 8.0+](https://php.watch/versions/8.5/array_first-array_last)

### DIFF
--- a/src/Php85/Php85.php
+++ b/src/Php85/Php85.php
@@ -33,4 +33,18 @@ final class Php85
 
         return $handler;
     }
+
+    public static function array_first(array $array)
+    {
+        foreach ($array as $value) {
+            return $value;
+        }
+
+        return null;
+    }
+
+    public static function array_last(array $array)
+    {
+        return $array ? current(array_slice($array, -1)) : null;
+    }
 }
--- a/tests/Php85/Php85Test.php
+++ b/tests/Php85/Php85Test.php
@@ -78,6 +78,44 @@ class Php85Test extends TestCase
         $handler = new TestHandlerInvokable();
         yield [$handler, $handler];
     }
+
+    public function testArrayFirstArrayLast()
+    {
+        $this->assertNull(array_first([]));
+        $this->assertNull(array_last([]));
+
+        $array = [1, 2, 3];
+        unset($array[0], $array[1], $array[2]);
+
+        $this->assertNull(array_first([]));
+        $this->assertNull(array_last([]));
+
+
+        $this->assertSame('single element', array_first(['single element']));
+        $this->assertSame('single element', array_last(['single element']));
+
+        $str = 'hello world';
+        $this->assertSame($str, array_first([&$str, 1]));
+        $this->assertSame(1, array_last([&$str, 1]));
+
+        $this->assertSame(1, array_first([1, &$str]));
+        $this->assertSame($str, array_last([1, &$str]));
+
+        $this->assertSame(1, array_first([1 => 1, 0 => 0, 3 => 3, 2 => 2]));
+        $this->assertSame(2, array_last([1 => 1, 0 => 0, 3 => 3, 2 => 2]));
+
+        $this->assertSame('a1', array_first(['a' => 'a1', 'b' => 'b1', 'c' => 'c1']));
+        $this->assertSame('c1', array_last(['a' => 'a1', 'b' => 'b1', 'c' => 'c1']));
+
+        $this->assertSame([], array_first([100 => []]));
+        $this->assertSame([], array_last([100 => []]));
+
+        $this->assertEquals(new \stdClass(), array_first([new \stdClass, false]));
+        $this->assertFalse(array_last([new \stdClass, false]));
+
+        $this->assertTrue(array_first([true, new \stdClass]));
+        $this->assertEquals(new \stdClass(), array_last([true, new \stdClass]));
+    }
 }
 
 class TestHandler
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Polyfills are provided for:
 - the `CURL_HTTP_VERSION_3` and `CURL_HTTP_VERSION_3ONLY` constants introduced in PHP 8.4;
 - the `get_error_handler` and `get_exception_handler` functions introduced in PHP 8.5;
 - the `NoDiscard` attribute introduced in PHP 8.5;
+- the `array_first` and `array_last` functions introduced in PHP 8.5;
 
 It is strongly recommended to upgrade your PHP version and/or install the missing
 extensions whenever possible. This polyfill should be used only when there is no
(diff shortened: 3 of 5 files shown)


src/Php85/Php85.php
```diff
@@ -33,4 +33,21 @@ public static function get_exception_handler(): ?callable
...
+        return null;
+    }
+
+    public static function array_last(array $array)
+    {
+        if ($array === []) {
+            return null;
+        }
+
+        return end($array);
```
TimWolla: This modifies the internal array pointer and thus is a behavioral difference. The `array_slice()` solution for `array_key_list()` should probably work here as well:

https://github.com/symfony/polyfill/blob/c4ee386e95ccdbea59cea802ea776d806319d506/src/Php73/bootstrap.php#L30C56-L30C67

inilim: image

The pointer changes only in the context of the current function, the pointer at the higher level remains the same.

nicolas-grekas: I'm just wondering if using end() triggers copy-on-write?
```suggestion
        return $array ? end($array) : null;
```

TimWolla: > The pointer changes only in the context of the current function, the pointer at the higher level remains the same.

Ah, of course. `array_last()` doesn't take a reference. However the use of `end()` will then cause a copy of the array as Nicolas mentioned, which is equally bad.

TimWolla: > I'm just wondering if using end() triggers copy-on-write?

Yes.

```
<?php

function array_last(array $array)
{
	if ($array === []) {
		return null;
	}

	echo "before end: ", memory_get_peak_usage() / 1024 / 1024, PHP_EOL;
	$value = end($array);
	echo "after end: ", memory_get_peak_usage() / 1024 / 1024, PHP_EOL;

	return $value;
}
function array_last_slice(array $array)
{
	if ($array === []) {
		return null;
	}

	echo "before slice: ", memory_get_peak_usage() / 1024 / 1024, PHP_EOL;
	[$value] = array_slice($array, -1, 1);
	echo "after slice: ", memory_get_peak_usage() / 1024 / 1024, PHP_EOL;

	return $value;
}

echo "initial: ", memory_get_peak_usage() / 1024 / 1024, PHP_EOL;

$a = array_fill(0, 100_000, new stdClass());

echo "after fill: ", memory_get_peak_usage() / 1024 / 1024, PHP_EOL;

var_dump(array_last($a));

echo "after last: ", memory_get_peak_usage() / 1024 / 1024, PHP_EOL;
```

Outputs for me:

```
initial: 0.41307830810547
after fill: 2.3816223144531
before end: 2.3816604614258
after end: 4.3856353759766
object(stdClass)#1 (0) {
}
after last: 4.3856735229492
```

for the `end()` implementation, doubling the array memory and:

```
initial: 0.4130859375
after fill: 2.3816299438477
before slice: 2.3816680908203
after slice: 2.3818359375
object(stdClass)#1 (0) {
}
after last: 2.3819274902344
```

for the `array_slice()` implementation, which is barely noticeable.

Ayesh: Thank you @nicolas-grekas @TimWolla. I rewrote it with an `array_slice` call.

TimWolla (approved):

src/Php85/Php85.php
Ayesh: Wow it really does double the memory footprint! I rewrote with an array_slice call, but without preserving the keys and directly accessing the `0`th key.

mtarld:
I'm afraid that the current `array_last` implementation doesn't work with associative arrays. 
Indeed:
```php
var_dump(array_slice(['a' => 2, 'b' => 1], -1, 1)[0]);
```
outputs a warning telling `Undefined array key 0`, whereas it should dump `1`.

nicolas-grekas:
I guess array_key_last should be used then

Ayesh:
@nicolas-grekas is it OK to bump the minimum PHP version to 7.3? If we have array_key_first/last, it would make this much more cleaner.

nicolas-grekas:
Ah right, then current(array_slice()) should do it

src/Php85/Php85.php
```diff
@@ -33,4 +33,21 @@ public static function get_exception_handler(): ?callable
...
+        return null;
+    }
+
+    public static function array_last(array $array)
+    {
+        if ($array === []) {
+            return null;
+        }
+
+        return array_slice($array, -1, 1)[0];
```
nicolas-grekas:
```suggestion
        return $array ? current(array_slice($array, -1)) : null;
```

src/Php85/Php85.php
```diff
@@ -33,4 +33,21 @@ public static function get_exception_handler(): ?callable
 
         return $handler;
     }
+
+    public static function array_first(array $array)
+    {
+        foreach ($array as $value) {
+            return $value;
+        }
+        return null;
```
nicolas-grekas:
```suggestion
        }

        return null;
```

tests/Php85/Php85Test.php
```diff
@@ -78,6 +78,41 @@ public static function provideHandler()
         $handler = new TestHandlerInvokable();
         yield [$handler, $handler];
     }
+
+    public function testArrayFirstArrayLast(): void
```
nicolas-grekas:
```suggestion
    public function testArrayFirstArrayLast()
```

Ayesh:
Updated with `current(array_slice)`, and the other change suggestions from Nicolas. Thanks.

TimWolla (approved):

nicolas-grekas (approved):

stof:
> is it OK to bump the minimum PHP version to 7.3? If we have array_key_first/last, it would make this much more cleaner.

We could depend on the polyfill-73 package

tests/Php85/Php85Test.php
```diff
@@ -78,6 +78,44 @@ public static function provideHandler()
...
+
+        $array = [1, 2, 3];
+        unset($array[0], $array[1], $array[2]);
+
+        $this->assertNull(array_first([]));
+        $this->assertNull(array_last([]));
+
+
+        $this->assertSame("single element", array_first(["single element"]));
+        $this->assertSame("single element", array_last(["single element"]));
```
OskarStark: What about using single quotes?

Ayesh: Right, thanks. I pushed with double quotes swapped with single quotes.

OskarStark: Thanks! 😍

nicolas-grekas:
Thank you @Ayesh.

ruudk:
@nicolas-grekas I noticed this is not tagged yet. Could that be done 🙏 ?

PR #527, opened 2025-05-26
[8.5] Add `locale_is_right_to_left`

alexander-schranz:
See https://php.watch/versions/8.5/locale_is_right_to_left-Locale-isRightToleft

### DIFF
--- a/src/Php85/Php85.php
+++ b/src/Php85/Php85.php
@@ -13,6 +13,7 @@ namespace Symfony\Polyfill\Php85;
 
 /**
  * @author Pierre Ambroise <pierre27.ambroise@gmail.com>
+ * @author Alexander Schranz <alexander@sulu.io>
  *
  * @internal
  */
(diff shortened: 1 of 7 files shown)


src/Php85/Php85.php
```diff
@@ -33,4 +34,8 @@ public static function get_exception_handler(): ?callable
 
         return $handler;
     }
+
+    public static function locale_is_right_to_left(string $locale): bool {
+        return (bool) preg_match('/^(?:ar|he|fa|ur|ps|sd|ug|ckb|yi|dv|ku_arab|ku-arab)(?:[_-].*)?$/i', $locale);
```
stof: I don't think this is the right implementation. What is right to left is a script, not a language. And locales might specify a script explicitly which is not the most likely script.

alexander-schranz: I see, any suggestions? Sadly the original code does not help here: https://github.com/unicode-org/icu/commit/53dcbe6141d98f1255c6937f53edb4c1ab1ec483

> * A script is right-to-left according to the CLDR script metadata
> * which corresponds to whether the script's letters have Bidi_Class=R or AL.

nicolas-grekas: This is what Gemini provides, let's take inspiration from this snippet?

```php
    function locale_is_right_to_left(string $locale): bool
    {
        // Define the list of RTL scripts within the function scope.
        // These are 4-letter ISO 15924 script codes.
        static $rtlScripts = [
            'Adlm', 'Arab', 'Armi', 'Hebr', 'Mani', 'Mend', 'Nkoo',
            'Phnx', 'Rohg', 'Samr', 'Syrc', 'Thaa',
        ];

        // This is a minimal, hardcoded version of the CLDR "likelySubtags" data.
        // It maps a language code to its most likely SCRIPT code.
        // This list is NOT exhaustive and is the primary weakness of this approach.
        static $languageToLikelyRtlScript = [
            'ar' => 'Arab', // Arabic
            'fa' => 'Arab', // Persian (Farsi)
            'ur' => 'Arab', // Urdu
            'ps' => 'Arab', // Pashto
            'sd' => 'Arab', // Sindhi
            'ug' => 'Arab', // Uyghur
            'ckb' => 'Arab', // Sorani Kurdish
            'he' => 'Hebr', // Hebrew
            'yi' => 'Hebr', // Yiddish
            'dv' => 'Thaa', // Dhivehi
            'nqo' => 'Nkoo', // N'Ko
        ];

        if (empty($locale)) {
            return false;
        }

        // Normalize separators and split the locale into parts.
        $localeParts = preg_split('/[_-]/', $locale);
        $language = strtolower($localeParts[0] ?? '');
        $script = null;

        // Look for an explicit script subtag (always 4 letters).
        foreach ($localeParts as $part) {
            if (strlen($part) === 4 && ctype_alpha($part)) {
                // Capitalize the first letter for standard format (e.g., "Arab", "Latn").
                $script = ucfirst(strtolower($part));
                break;
            }
        }

        // If no explicit script was found, try to infer it from our map.
        if ($script === null) {
            $script = $languageToLikelyScript[$language] ?? null;
        }

        // If we couldn't determine a script, we can't determine direction.
        if ($script === null) {
            // Fallback for languages where the code itself is a strong indicator
            if (in_array($language, ['ar', 'he', 'fa', 'ur', 'ps', 'sd', 'ug', 'ckb', 'yi', 'dv'])) {
                 return true;
            }
            return false;
        }

        // Check if the determined script is in our list of RTL scripts.
        return in_array($script, $rtlScripts, true);
    }
}
```

stof:
our `Locale` polyfill in `symfony/polyfill-intl-icu` should also support `Locale::isRightToLeft` to support the case of projects requiring PHP 8.5+ but making ext-intl optional thanks to the polyfill (copying the same implementation there).

nicolas-grekas:
Thank you @alexander-schranz.

PR #532, opened 2025-06-28
[Intl] Add PHP 8.5 `IntlListFormatter` to ICU polyfill

### LINKED ISSUE #530: Add `IntlListFormatter`
See 
https://ungureanu.blog/2025/06/24/new-in-php-intl-8-5-intllistformatter-display-arrays-as-locale-aware-lists/

I'd suggest handling only a few locales (or maybe only "en" as a start), with the rules from
https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-misc-full/main/en/listPatterns.json

Ayesh:
Adds a new polyfill to `symfony/polyfill-intl-icu` that provides the functionality of the new `IntlListFormatter` to PHP 7.2 and later.

 - [ICU listPatterns](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-misc-full/main/en/listPatterns.json)
 - [php-src commit](https://github.com/php/php-src/commit/3f7545245)
 - [PHP.Watch: IntlListFormatter](https://php.watch/versions/8.5/IntlListFormatter)

Closes GH-530.

### DIFF
--- a/src/Intl/Icu/IntlListFormatter.php
+++ b/src/Intl/Icu/IntlListFormatter.php
@@ -1,0 +1,169 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Intl\Icu;
+
+/**
+ * @author Ayesh Karunaratne <ayesh@aye.sh>
+ *
+ * @internal
+ */
+class IntlListFormatter
+{
+    public const TYPE_AND = 0;
+    public const TYPE_OR = 1;
+    public const TYPE_UNITS = 2;
+
+    public const WIDTH_WIDE = 0;
+    public const WIDTH_SHORT = 1;
+    public const WIDTH_NARROW = 2;
+
+    private $type;
+    private $width;
+
+    private const TYPE_MAP = [
+        self::TYPE_AND => 'standard',
+        self::TYPE_OR => 'or',
+        self::TYPE_UNITS => 'unit',
+    ];
+
+    private const WIDTH_MAP = [
+        self::WIDTH_WIDE => '',
+        self::WIDTH_SHORT => '-short',
+        self::WIDTH_NARROW => '-narrow',
+    ];
+
+    private const EN_LIST_PATTERNS = [
+        'listPattern-type-standard' => [
+            'start' => '{0}, {1}',
+            'middle' => '{0}, {1}',
+            'end' => '{0}, and {1}',
+            2 => '{0} and {1}',
+        ],
+        'listPattern-type-or' => [
+            'start' => '{0}, {1}',
+            'middle' => '{0}, {1}',
+            'end' => '{0}, or {1}',
+            2 => '{0} or {1}',
+        ],
+        'listPattern-type-or-narrow' => [
+            'start' => '{0}, {1}',
+            'middle' => '{0}, {1}',
+            'end' => '{0}, or {1}',
+            2 => '{0} or {1}',
+        ],
+        'listPattern-type-or-short' => [
+            'start' => '{0}, {1}',
+            'middle' => '{0}, {1}',
+            'end' => '{0}, or {1}',
+            2 => '{0} or {1}',
+        ],
+        'listPattern-type-standard-narrow' => [
+            'start' => '{0}, {1}',
+            'middle' => '{0}, {1}',
+            'end' => '{0}, {1}',
+            2 => '{0}, {1}',
+        ],
+        'listPattern-type-standard-short' => [
+            'start' => '{0}, {1}',
+            'middle' => '{0}, {1}',
+            'end' => '{0}, & {1}',
+            2 => '{0} & {1}',
+        ],
+        'listPattern-type-unit' => [
+            'start' => '{0}, {1}',
+            'middle' => '{0}, {1}',
+            'end' => '{0}, {1}',
+            2 => '{0}, {1}',
+        ],
+        'listPattern-type-unit-narrow' => [
+            'start' => '{0} {1}',
+            'middle' => '{0} {1}',
+            'end' => '{0} {1}',
+            2 => '{0} {1}',
+        ],
+        'listPattern-type-unit-short' => [
+            'start' => '{0}, {1}',
+            'middle' => '{0}, {1}',
+            'end' => '{0}, {1}',
+            2 => '{0}, {1}',
+        ],
+    ];
+
+    public function __construct(string $locale, int $type = self::TYPE_AND, int $width = self::WIDTH_WIDE)
+    {
+        if ('en' !== $locale && 0 !== strpos($locale, 'en')) {
+            if (80000 > \PHP_VERSION_ID) {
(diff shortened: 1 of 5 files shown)


Ayesh:
// cc @BogdanUngureanu, I could use your inputs/review here :)

src/Intl/Icu/IntlListFormatter.php
```diff
@@ -0,0 +1,206 @@
...
+namespace Symfony\Polyfill\Intl\ListFormatter;
+
+/**
+ * A polyfill implementation of the IntlListFormatter class provided by the intl extension.
+ *
+ * @author Ayesh Karunaratne <ayesh@aye.sh>
+ *
+ * @internal
+ */
+class IntlListFormatter
```
BogdanUngureanu, line 19: It should be final, non-serializable and not able to clone it.

src/Intl/ListFormatter/IntlListFormatter.php
```diff
@@ -0,0 +1,206 @@
...
+            ]);
+        }
+
+        return strtr($pattern["end"], [
+            "{0}" => $result,
+            "{1}" => $strings[$itemCount - 1],
+        ]);
+    }
+
+    protected function getListPattern(): array {
```
BogdanUngureanu: Why not private?

src/Intl/Icu/IntlListFormatter.php
```diff
@@ -0,0 +1,206 @@
...
+     * @var int
+     */
+    private $type;
+    /**
+     * @var int
+     */
+    private $width;
+
+    protected static $listPatterns = [
+        'en' => [
```
BogdanUngureanu: Maybe make it clear that this is en_US since it's using the Oxford comma?

src/Intl/Icu/IntlListFormatter.php
```diff
@@ -0,0 +1,206 @@
...
+    ];
+
+    public function __construct(
+        string $locale,
+        int $type = self::TYPE_AND,
+        int $width = self::WIDTH_WIDE
+    ) {
+        $exceptionClass = PHP_VERSION_ID >= 80000 ? \ValueError::class : \InvalidArgumentException::class;
+        if ($locale !== 'en' && strpos($locale, 'en') !== 0) {
+            throw new $exceptionClass('Invalid locale, only "en" and "en-*" locales are supported');
```
BogdanUngureanu: I have some mixed feelings about this. Since it's using the en-US set, for en-gb it would return a different result than the real class.

stof:
This should be added in symfony/polyfill-intl-icu (with the other formatters) instead of being a new package IMO

> It only supports `en` locale. However, PHP applications that require this package can extend the namespaced polyfill
> (`\Symfony\Polyfill\Intl\ListFormatter\IntlListFormatter`) and add their own list patterns by extending the static variable (`\Symfony\Polyfill\Intl\ListFormatter\IntlListFormatter::$listPatterns`) that holds these patterns.

Extending the internal class is not covered by BC and is not a supported use case.
Projects wanting the full support for ICU formatting should install ext-intl.

src/Intl/Icu/IntlListFormatter.php
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * @var int
+     */
+    private $type;
+    /**
+     * @var int
+     */
+    private $width;
+
+    protected static $listPatterns = [
```
stof: no need to make this protected (nor mutable) IMO

src/Intl/Icu/IntlListFormatter.php
stof: this internal class cannot be final, as the actual polyfill class extends it.

stof:
@Ayesh do you plan to finish this PR (by taking review comments into account) ?

Ayesh:
Thank you for your review Christophe. I will try and submit fixed addressing the reviewed points by this weekend.

Ayesh:
Thank you for your reviews. 

I changed the PR, so that this now proposes to add `IntlListFormatter` to the ICU package. I also agree that it makes more sense, as opposed to having a separate installable package. 

I also made it so that the class is self-contained, and not extendable anymore (making `protected` elements `private`, formatting pattern changed to a const, etc).

src/Intl/Icu/IntlListFormatter.php
```diff
@@ -0,0 +1,206 @@
...
+     * @var int
+     */
+    private $type;
+    /**
+     * @var int
+     */
+    private $width;
+
+    private const LIST_PATTERNS = [
+        'en' => [
```
stof: As we support only `en` anyway, I suggest simplifying the implementation even more, by having a `EN_LIST_PATTERNS` constant only.

composer.json
```diff
@@ -38,6 +38,7 @@
         "symfony/polyfill-intl-grapheme": "self.version",
         "symfony/polyfill-intl-icu": "self.version",
         "symfony/polyfill-intl-messageformatter": "self.version",
+        "symfony/polyfill-intl-listformatter": "self.version",
```
stof: should be reverted, as there is no such new component anymore.

src/Intl/Icu/IntlListFormatter.php
```diff
@@ -0,0 +1,206 @@
...
+                break;
+        }
+
+        $pattern = $pattern['listPattern-type-' . $lookupKeyType . $lookupKeyWidth];
+
+        if ($itemCount === 2) {
+            return strtr($pattern[2], ['{0}' => (string) $strings[0], '{1}' => (string) $strings[1]]);
+        }
+
+        if ($itemCount === 3) {
```
stof: You could remove this case. It can be covered by the same code that for 4 items or more (the `for` loop will simply do no iterations, which is not an error)

src/Intl/Icu/IntlListFormatter.php
```diff
@@ -0,0 +1,206 @@
...
+        if ($type !== self::TYPE_AND && $type !== self::TYPE_OR && $type !== self::TYPE_UNITS) {
+            throw new $exceptionClass('Argument #2 ($type) must be one of IntlListFormatter::TYPE_AND, IntlListFormatter::TYPE_OR, or IntlListFormatter::TYPE_UNITS.');
+        }
+
+        if ($width !== self::WIDTH_WIDE && $width !== self::WIDTH_SHORT && $width !== self::WIDTH_NARROW) {
+            throw new $exceptionClass('Argument #3 ($width) must be one of IntlListFormatter::WIDTH_WIDE, IntlListFormatter::WIDTH_SHORT, or IntlListFormatter::WIDTH_NARROW.');
+        }
+
+
+        $this->locale = 'en';
```
stof: We don't even need to store it in a private property, as we don't support other values than `en`

tests/Intl/Icu/IntlListFormatterTest.php
```diff
@@ -0,0 +1,156 @@
...
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @author Ayesh Karunaratne <ayesh@aye.sh>
+ *
+ * @group class-polyfill
+ */
+class IntlListFormatterTest extends TestCase
+{
+    public function testUnsupportedLocales()
```
stof: Testing support locales should be done in a separate test than `testUnsupportedLocales` to make the intent of the tests clear.

stof (changes requested):

nicolas-grekas:
Thank you @Ayesh.

PR #534, opened 2025-07-08
[8.3] Fix json_validate edge case

### LINKED ISSUE #533: Php83::json_validate() - JSON_ERROR_INVALID_PROPERTY_NAME bug
`Symfony\Polyfill\Php83::json_validate()` attempts to decode as an object (which faiils) / requires setting `JSON_OBJECT_AS_ARRAY` flag (not necessary with php's native function)

```php
$json = json_encode(array(
    "\x00null" => 'JSON_ERROR_INVALID_PROPERTY_NAME',
    'string' => 'valid json, but must be decoded as an array,
));

json_decode($json, true); // success as array
json_decode($json, false); //  failure as obj

\Symfony\Polyfill\Php83::json_validate($json); // returns false
json_validate($json); // returns true
```

when using php's native `json_validate`, it is not necessary to set the `JSON_OBJECT_AS_ARRAY` flag

IonBazan:
Fixes an edge case where using a property containing a `\x00` character in property name produces incorrect result in `json_validate()`.

Fixes #533

### DIFF
--- a/src/Php83/Php83.php
+++ b/src/Php83/Php83.php
@@ -35,7 +35,7 @@ final class Php83
             throw new \ValueError(sprintf('json_validate(): Argument #2 ($depth) must be less than %d', self::JSON_MAX_DEPTH));
         }
 
-        json_decode($json, null, $depth, $flags);
+        json_decode($json, true, $depth, $flags);
 
         return \JSON_ERROR_NONE === json_last_error();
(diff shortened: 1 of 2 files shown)


stof (approved):

nicolas-grekas:
Thank you @IonBazan.

PR #538, opened 2025-07-29
Update `Attribute::class` constants to PHP 8.5 values

### LINKED ISSUE #537: Changes to `Attribute::class` constants in PHP 8.5
The following code:
```php
var_export((new ReflectionClass(Attribute::class))->getConstants());
```

In PHP 8.0-8.4 returns:
```
array (
  'TARGET_CLASS' => 1,
  'TARGET_FUNCTION' => 2,
  'TARGET_METHOD' => 4,
  'TARGET_PROPERTY' => 8,
  'TARGET_CLASS_CONSTANT' => 16,
  'TARGET_PARAMETER' => 32,
  'TARGET_ALL' => 63,
  'IS_REPEATABLE' => 64,
)
```

PHP 8.5 introduced [attributes on constants](https://wiki.php.net/rfc/attributes-on-constants), so the array in PHP 8.5 is:
```
array (
  'TARGET_CLASS' => 1,
  'TARGET_FUNCTION' => 2,
  'TARGET_METHOD' => 4,
  'TARGET_PROPERTY' => 8,
  'TARGET_CLASS_CONSTANT' => 16,
  'TARGET_PARAMETER' => 32,
  'TARGET_CONSTANT' => 64,
  'TARGET_ALL' => 127,
  'IS_REPEATABLE' => 128,
)
```

How to address the issue?

Should the [existing stub](https://github.com/symfony/polyfill/blob/v1.32.0/src/Php80/Resources/stubs/Attribute.php) be updated with something like:
```
    public const TARGET_CONSTANT = \PHP_VERSION_ID >= 80500 ? 64 : null;
    public const TARGET_ALL = \PHP_VERSION_ID >= 80500 ? 127 : 63;
    public const IS_REPEATABLE = \PHP_VERSION_ID >= 80500 ? 128 : 64;
```
Or maybe a new `Attribute` class in `src/Php85/Resources/stubs`?

_There is no ternary operator in any constant yet, and also there are not 2 the same classes for different PHP versions, so each of the above would be the first such solution_

Or any better idea to address the problem?

kubawerlos:
Fixes https://github.com/symfony/polyfill/issues/537

### DIFF
--- a/src/Php80/Resources/stubs/Attribute.php
+++ b/src/Php80/Resources/stubs/Attribute.php
@@ -18,8 +18,9 @@ final class Attribute
     public const TARGET_PROPERTY = 8;
     public const TARGET_CLASS_CONSTANT = 16;
     public const TARGET_PARAMETER = 32;
-    public const TARGET_ALL = 63;
-    public const IS_REPEATABLE = 64;
+    public const TARGET_CONSTANT = 64;
+    public const TARGET_ALL = 127;
+    public const IS_REPEATABLE = 128;
 
     /** @var int */
     public $flags;


nicolas-grekas (approved):

xabbuh (approved):

DanielEScherzer (commented):
I had considered sending a patch like this, but decided against it
* PHP 7.4 code with polyfill: Attribute::TARGET_ALL is 127
* PHP 8.0-8.4 code: Attribute::TARGET_ALL is 63
* PHP 8.5: Attribute::TARGET_ALL is 127

Which to me would be confusing. I would expect a PHP 8.0 polyfill to polyfill the class *as it existed in PHP 8.0*. It is also worth pointing out that for TARGET_CONSTANT, actually trying to put attributes on constants was a parser error until 8.5, so it isn't really useful

nicolas-grekas:
OK, let me close per @DanielEScherzer's comment.
Another angle is that the new constant cannot be polyfilled on PHP 8.0 to 8.4, so that polyfiling it on 7.4 don't really help - since portable code will need to be cope for those versions anyway.
Thanks to everybody involved.

PR #539, opened 2025-08-10
uuid: add V6

pavetheway91:
The extension [added support](https://github.com/php/pecl-networking-uuid/commit/ea46ef62d4d0d97c254afbc072d7da3559e2e435) for V6 and V7 few months ago.

This PR contains implementation for V6. I might add V7 in another one, but I'd like to see some opinions about it's implementation first.

Option A:
Adaptation of [Symfony\Component\Uid\UuidV7](https://github.com/symfony/uid/blob/a69f69f3159b852651a6bf45a9fdd149520525bb/UuidV7.php#L54).

Option B:
Plain and simple timestamp + some random bits, just ignore the optional parts of [the spec](https://www.rfc-editor.org/rfc/rfc9562.html#section-5.7). Perhaps a better fit for the style of this library.

### DIFF
--- a/src/Uuid/Uuid.php
+++ b/src/Uuid/Uuid.php
@@ -29,6 +29,7 @@ final class Uuid
     public const UUID_TYPE_NAME = 1; // Deprecated alias
     public const UUID_TYPE_RANDOM = 4;
     public const UUID_TYPE_SHA1 = 5;
+    public const UUID_TYPE_TIME_V6 = 6;
     public const UUID_TYPE_NULL = -1;
     public const UUID_TYPE_INVALID = -42;
 
@@ -51,6 +52,8 @@ final class Uuid
             case self::UUID_TYPE_NAME:
             case self::UUID_TYPE_TIME:
                 return self::uuid_generate_time();
+            case self::UUID_TYPE_TIME_V6:
+                return self::uuid_generate_time_v6();
             case self::UUID_TYPE_DCE:
             case self::UUID_TYPE_RANDOM:
             case self::UUID_TYPE_DEFAULT:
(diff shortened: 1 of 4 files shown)


pavetheway91:
Was that a different test suite?

Screenshot_20250810_213945

edit: Seems like it ran the test suite against an old version of extension (rather than the polyfill) on that CI thing

nicolas-grekas:
> edit: Seems like it ran the test suite against an old version of extension (rather than the polyfill) on that CI thing

That's the issue: what should we do when a legacy extension is installed?
Eg adding the constants doesn't look right, since those would be ineffective and misleading.

Is it worth polyfilling v6 & v7? Shouldn't ppl be encouraged to use symfony/uid instead?

stof:
Polyfilling the new constants when the native function is used is indeed very bad, as it breaks the ability to detect whether the function supports generating v6 UUIDs (by checking whether the constant exists) and would lead to runtime errors in the function when using the constant.

pavetheway91:
One way to resolve this would be to a check for an extension before defining a constant. It would allow people to use this for V6 and V7 if they know there's either no extension or that extension is 1.3.

pavetheway91:
But I do understand Nicolas's point here. I won't get offended if you close this.

Completely out of topic, but since I'm here:

I ran into an situation few days ago where Symfony Http Client surprised me and I ended up writing a very partial polyfill for filter_var. If felt like making it an actual polyfill at some point, would like to adopt it?

Or should I send a PR to Symfony to make that Http Client to not expect filter extension to be there?

nicolas-grekas:
> `filter_var($ip, \FILTER_VALIDATE_IP, $flag)`

this is going to be non-trivial to replace...
better add ext-filter as a dep?

Running on super trimmed php isn't that common :)

Let me close here, thanks for your understanding and for proposing.

PR #543, opened 2025-08-26
[8.5] Add polyfill for FILTER_THROW_ON_FAILURE and exceptions

DanielEScherzer:
See https://wiki.php.net/rfc/filter_throw_on_failure

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ Polyfills are provided for:
 - the `NoDiscard` attribute introduced in PHP 8.5;
 - the `array_first` and `array_last` functions introduced in PHP 8.5;
 - the `DelayedTargetValidation` attribute introduced in PHP 8.5;
+- the `FILTER_THROW_ON_FAILURE` constant introduced in PHP 8.5;
+- the `\Filter\FilterException` and `\Filter\FilterFailedException` classes introduced in PHP 8.5;
 
 It is strongly recommended to upgrade your PHP version and/or install the missing
 extensions whenever possible. This polyfill should be used only when there is no
--- a/src/Php85/README.md
+++ b/src/Php85/README.md
@@ -7,6 +7,8 @@ This component provides features added to PHP 8.5 core:
 - [`NoDiscard`](https://wiki.php.net/rfc/marking_return_value_as_important)
 - [`array_first` and `array_last`](https://wiki.php.net/rfc/array_first_last)
 - [`DelayedTargetValidation`](https://wiki.php.net/rfc/delayedtargetvalidation_attribute)
+- `FILTER_THROW_ON_FAILURE`, `\Filter\FilterException`, and
+`\Filter\FilterFailedException` (see [RFC](https://wiki.php.net/rfc/filter_throw_on_failure))
 
 More information can be found in the
 [main Polyfill README](https://github.com/symfony/polyfill/blob/main/README.md).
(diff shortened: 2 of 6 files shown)


DanielEScherzer:
8.5 failure is
> /home/runner/work/_temp/0c018e61-8d65-4ab3-88d8-32976237b682.sh: line 1:  5649 Segmentation fault      (core dumped) composer --prefer-source --no-progress --ansi install

but probably not caused by me, since the test setup also died:
```
==> Setup PHP
✓ PHP Installed PHP 8.5.0-dev (fc0dd9d4cad68396715274fb9e8450d41e41cab8)

==> Setup Extensions
✓ apcu Enabled
✓ intl Installed and enabled with ICU 73.2
✓ mbstring Enabled
✓ odbc Enabled

==> Setup uuid-1.3.0
/home/runner/work/_actions/shivammathur/setup-php/v2/src/scripts/../scripts/extensions/add_extensions.sh: line 79:  2927 Segmentation fault      (core dumped) php -d'error_reporting=0' "${src:?}"/scripts/extensions/extension_map.php /tmp/map"$version".orig > /dev/null 2>&1
phpize
configure
make
make install
/home/runner/work/_actions/shivammathur/setup-php/v2/src/scripts/../scripts/extensions/add_extensions.sh: line 9:  5253 Segmentation fault      (core dumped) php -m > "$extension_list"
/home/runner/work/_actions/shivammathur/setup-php/v2/src/scripts/../scripts/extensions/add_extensions.sh: line 79:  5280 Segmentation fault      (core dumped) php -d'error_reporting=0' "${src:?}"/scripts/extensions/extension_map.php /tmp/map"$version".orig > /dev/null 2>&1
✓ uuid-1.3.0 Installed from https://pecl.php.net/uuid/uuid and enabled
✓ uuid Installed and enabled

==> Setup Tools
/home/runner/work/_actions/shivammathur/setup-php/v2/src/scripts/../scripts/tools/add_tools.sh: line 27:  5534 Segmentation fault      (core dumped) php -r "try {\$p=new Phar('$tool_path.phar', 0);exit(0);} catch(Exception \$e) {exit(1);}"
✓ composer Added composer 2.8.11

nicolas-grekas:
I'm not sure we should polyfill these: they're just sidekick of a main feature we cannot polyfill, so having a partial polyfill might be confusing.

DanielEScherzer:
> I'm not sure we should polyfill these: they're just sidekick of a main feature we cannot polyfill, so having a partial polyfill might be confusing.

Static analysis might complain about the classes not existing?
I'm not entirely clear on the scope of the Symfony polyfills, e.g. we can't actually polyfill the behavior of `#[\NoDiscard]` or `#[\DelayedTargetValidation]` but those attributes are still included in the 8.5 polyfills

nicolas-grekas:
attributes are declarative so there's no real need of engine support to have them convey semantics, that's why they're fine

DanielEScherzer:
> attributes are declarative so there's no real need of engine support to have them convey semantics, that's why they're fine

Okay. I figured it wouldn't hurt to send this PR, it wasn't too much work if you decide you don't want to include it

stof:
I also vote -1 on adding the polyfill for the `FILTER_THROW_ON_FAILURE`. The only purpose of this constant is to configure the new behavior in `filter_var`, which we cannot polyfill.
It is better to leave the constant undefined (which can be used for feature detection) than to define it without its behavior.

DanielEScherzer:
We can remove the constant and just add the classes if there are no objections to the classes being included?
I'll update this tomorrow

nicolas-grekas:
Let's add only the classes, works for me.

nicolas-grekas:
Let me close as is for history. PR welcome for the classes only, if that can be useful.

PR #545, opened 2025-09-07
PHP 8.5 - add `PHP_BUILD_DATE` constant

arshidkv12:
Provide a polyfill for the `PHP_BUILD_DATE` constant (added in PHP 8.5).
- Define `PHP_BUILD_DATE` if not already defined, using `Php85::php_build_date()`
- Introduce `php_build_date()` helper to extract the build timestamp from `phpinfo()`
- Add `PHPUnit` test to ensure the helper returns a valid and parseable build date string

### DIFF
--- a/src/Php85/Php85.php
+++ b/src/Php85/Php85.php
@@ -47,4 +47,15 @@ final class Php85
     {
         return $array ? current(array_slice($array, -1)) : null;
     }
+
+    public static function php_build_date()
+    {
+        ob_start();
+        phpinfo(INFO_GENERAL);
+        $info = ob_get_clean();
+
+        preg_match('@Build Date(?:( => | </td><td class="v">))(?<buildtime>[A-Za-z]{3} (?: \d|\d\d) \d{4} \d{2}:\d{2}:\d{2})@', $info, $matches);
+
+        return $matches['buildtime'];
+    }
 }
--- a/src/Php85/Php85.php
+++ b/src/Php85/Php85.php
@@ -56,6 +56,6 @@ final class Php85
 
         preg_match('@Build Date(?:( => | </td><td class="v">))(?<buildtime>[A-Za-z]{3} (?: \d|\d\d) \d{4} \d{2}:\d{2}:\d{2})@', $info, $matches);
 
-        return $matches['buildtime'];
+        return $matches['buildtime'] ?? '';
     }
 }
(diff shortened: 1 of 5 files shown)


tests/Php85/Php85Test.php
```diff
@@ -114,6 +114,15 @@ public function testArrayFirstArrayLast()
         $this->assertTrue(array_first([true, new \stdClass()]));
         $this->assertEquals(new \stdClass(), array_last([true, new \stdClass()]));
     }
+
+    public function testPhpBuildDate(){
```
OskarStark:
```suggestion
    public function testPhpBuildDate()
    {
```

arshidkv12: Ok

src/Php85/Php85.php
```diff
@@ -47,4 +47,15 @@ public static function array_last(array $array)
...
         return $array ? current(array_slice($array, -1)) : null;
     }
+
+    public static function php_build_date()
+    {
+        ob_start();
+        phpinfo(INFO_GENERAL);
+        $info = ob_get_clean();
+
+        preg_match('@Build Date(?:( => | </td><td class="v">))(?<buildtime>[A-Za-z]{3} (?: \d|\d\d) \d{4} \d{2}:\d{2}:\d{2})@', $info, $matches);
```
Spomky, line 57: What if `preg_match` returns `false`?

tests/Php85/Php85Test.php
```diff
@@ -114,6 +114,16 @@ public function testArrayFirstArrayLast()
         $this->assertTrue(array_first([true, new \stdClass()]));
         $this->assertEquals(new \stdClass(), array_last([true, new \stdClass()]));
     }
+
+    public function testPhpBuildDate()
```
keradus, line 118: based on [docs of phpinfo](https://www.php.net/phpinfo) , would be good to run (or mock the run) under:
- cli
- webserver
- webserver with [expose_php](https://www.php.net/manual/en/ini.core.php#ini.expose-php) configuration setting is set to off

nicolas-grekas:
I'm going to save us all some time: this cannot be polyfilled efficiently.

The polyfill relies on running some non-trivial code for everysingle requests, to populate a const that's going to be needed very few times.

That's not worth the perf (and CO2) impact !

Thanks for your understanding.

arshidkv12:
Thank you

PR #546, opened 2025-09-19
[8.4] implement bcround, bcceil and bcfloor

Dean151:
- introduced `bcround`, `bcfloor` and `bcceil` polyfills
- RoundingMode cannot be an enum, and const cannot contain an object. This is the best pick I got to simulate an enum for PHP < 8.1
- Tests are based on PHP bcmath extension official tests: https://github.com/php/php-src/tree/master/ext/bcmath/tests

### DIFF
--- a/src/Php84/Resources/stubs/RoundingMode.php
+++ b/src/Php84/Resources/stubs/RoundingMode.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+if (\PHP_VERSION_ID < 80100) {
+    final class RoundingMode
+    {
+        const HalfAwayFromZero = 0;
+        const HalfTowardsZero = 1;
+        const HalfEven = 2;
+        const HalfOdd = 3;
+        const TowardsZero = 4;
+        const AwayFromZero = 5;
+        const NegativeInfinity = 6;
+        const PositiveInfinity = 7;
+
+        private function __construct()
+        {
+        }
+
+        public static function cases(): array
+        {
+            return [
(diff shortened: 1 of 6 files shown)


src/Php84/Resources/stubs/RoundingMode.php
```diff
@@ -0,0 +1,40 @@
...
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+if (\PHP_VERSION_ID < 80400 && extension_loaded('bcmath')) {
```
stof: RoundingMode is not part of BCMath (BCMath reuses the enum introduced for `round`)

Dean151: Removed that. Good catch!

src/Php84/Resources/stubs/RoundingMode.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+if (\PHP_VERSION_ID < 80400) {
+    // @author Thomas Durand <github@thomasdurand.fr>
+    final class RoundingMode {
```
nicolas-grekas: when PHP >= 8.1 is used, we should use a real enum

Dean151: @nicolas-grekas When I try to introduce an enum, even behind an if PHP_VERSION_ID, older PHP version are complaining pretty badly on the syntax.
See [this CI run](https://github.com/symfony/polyfill/actions/runs/18573212824/job/52952067059?pr=546)
An idea?

nicolas-grekas:
Thank you @Dean151.

PR #549, opened 2025-10-16
Add polyfill for PDO driver specific subclasses

jnoordsij:
As a follow-up to #547, this PR aims to implement a full-fledged set of polyfills for the driver specific PDO subclasses introduced in PHP 8.4; see also https://wiki.php.net/rfc/pdo_driver_specific_subclasses.

As of PHP 8.5, the driver specific constants and methods available on the base `PDO` class will be deprecated, to be removed in PHP 9.0. See also https://wiki.php.net/rfc/deprecations_php_8_5#deprecate_driver_specific_pdo_constants_and_methods.

This PR adds a polyfill to use the driver-specific child classes for PHP versions prior to PHP 8.4. This in turn allows one to already refer to the 'new' constant, to prevent deprecation errors on PHP 8.5, while retaining backwards compatibility. This can be particularly useful for code requiring support for multiple versions; see e.g. https://github.com/laravel/framework/issues/57141.

The classes should be suitable for constructing on older PHP versions.

Note the `PDO::connect` static method that has been introduced simultaneously has not been polyfilled, as I can't think of a proper way to do such a thing currently.

### DIFF
--- a/src/Php84/Resources/stubs/Pdo/Dblib.php
+++ b/src/Php84/Resources/stubs/Pdo/Dblib.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Pdo;
+
+if (\PHP_VERSION_ID < 80400 && \extension_loaded('pdo_dblib')) {
+    class Dblib extends \PDO
+    {
+        public const ATTR_CONNECTION_TIMEOUT = \PDO::DBLIB_ATTR_CONNECTION_TIMEOUT;
+        public const ATTR_QUERY_TIMEOUT = \PDO::DBLIB_ATTR_QUERY_TIMEOUT;
+        public const ATTR_STRINGIFY_UNIQUEIDENTIFIER = \PDO::DBLIB_ATTR_STRINGIFY_UNIQUEIDENTIFIER;
+        public const ATTR_VERSION = \PDO::DBLIB_ATTR_VERSION;
+        public const ATTR_TDS_VERSION = \PHP_VERSION_ID >= 70300 ? \PDO::DBLIB_ATTR_TDS_VERSION : 1004;
+        public const ATTR_SKIP_EMPTY_ROWSETS = \PHP_VERSION_ID >= 70300 ? \PDO::DBLIB_ATTR_SKIP_EMPTY_ROWSETS : 1005;
+        public const ATTR_DATETIME_CONVERT = \PHP_VERSION_ID >= 70300 ? \PDO::DBLIB_ATTR_DATETIME_CONVERT : 1006;
+
+        public function __construct(string $dsn, ?string $username = null, ?string $password = null, ?array $options = null)
+        {
+            parent::__construct($dsn, $username, $password, $options);
+
+            if ('dblib' !== $driver = $this->getAttribute(\PDO::ATTR_DRIVER_NAME)) {
+                throw new \PDOException(\sprintf('Pdo\Dblib::__construct() cannot be used for connecting to the "%s" driver', $driver));
+            }
+        }
+
+        public static function connect(string $dsn, ?string $username = null, ?string $password = null, ?array $options = null): self
+        {
+            try {
+                return new self($dsn, $username, $password, $options);
+            } catch (\PDOException $e) {
+                throw preg_match('/^Pdo\\\\Dblib::__construct\(\) cannot be used for connecting to the "([a-z]+)" driver/', $e->getMessage(), $matches) ? new \PDOException(\sprintf('Pdo\Dblib::connect() cannot be used for connecting to the "%s" driver', $matches[1])) : $e;
+            }
+        }
+    }
+}
--- a/src/Php84/Resources/stubs/Pdo/Mysql.php
+++ b/src/Php84/Resources/stubs/Pdo/Mysql.php
@@ -1,0 +1,101 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Pdo;
+
+use PDO;
+
+if (\PHP_VERSION_ID < 80400 && \extension_loaded('pdo_mysql')) {
+    // Feature detection for non-mysqlnd; see also https://www.php.net/manual/en/class.pdo-mysql.php#pdo-mysql.constants.attr-max-buffer-size
+    if (\defined('PDO::MYSQL_ATTR_MAX_BUFFER_SIZE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_FILE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_GROUP')) {
+        class Mysql extends \PDO
+        {
+            public const ATTR_COMPRESS = \PDO::MYSQL_ATTR_COMPRESS;
+            public const ATTR_DIRECT_QUERY = \PDO::MYSQL_ATTR_DIRECT_QUERY;
+            public const ATTR_FOUND_ROWS = \PDO::MYSQL_ATTR_FOUND_ROWS;
+            public const ATTR_IGNORE_SPACE = \PDO::MYSQL_ATTR_IGNORE_SPACE;
+            public const ATTR_INIT_COMMAND = \PDO::MYSQL_ATTR_INIT_COMMAND;
+            public const ATTR_LOCAL_INFILE = \PDO::MYSQL_ATTR_LOCAL_INFILE;
+            public const ATTR_LOCAL_INFILE_DIRECTORY = \PHP_VERSION_ID >= 80100 ? \PDO::MYSQL_ATTR_LOCAL_INFILE_DIRECTORY : 1015;
+            public const ATTR_MAX_BUFFER_SIZE = \PDO::MYSQL_ATTR_MAX_BUFFER_SIZE;
+            public const ATTR_MULTI_STATEMENTS = \PDO::MYSQL_ATTR_MULTI_STATEMENTS;
+            public const ATTR_READ_DEFAULT_FILE = \PDO::MYSQL_ATTR_READ_DEFAULT_FILE;
+            public const ATTR_READ_DEFAULT_GROUP = \PDO::MYSQL_ATTR_READ_DEFAULT_GROUP;
+            public const ATTR_SERVER_PUBLIC_KEY = \PDO::MYSQL_ATTR_SERVER_PUBLIC_KEY;
+            public const ATTR_SSL_CA = \PDO::MYSQL_ATTR_SSL_CA;
+            public const ATTR_SSL_CAPATH = \PDO::MYSQL_ATTR_SSL_CAPATH;
+            public const ATTR_SSL_CERT = \PDO::MYSQL_ATTR_SSL_CERT;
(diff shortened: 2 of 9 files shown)


src/Php84/Resources/stubs/Pdo/Dblib.php
```diff
@@ -0,0 +1,37 @@
...
+        public const ATTR_SKIP_EMPTY_ROWSETS = \PHP_VERSION_ID >= 70300 ? PDO::DBLIB_ATTR_SKIP_EMPTY_ROWSETS : 1005;
+        public const ATTR_DATETIME_CONVERT = \PHP_VERSION_ID >= 70300 ? \PDO::DBLIB_ATTR_DATETIME_CONVERT : 1006;
+
+        public function __construct()
+        {
+            parent::__construct();
+
+            if (($driver = $this->getAttribute(PDO::ATTR_DRIVER_NAME)) !== 'dblib')
+            {
+                throw new PDOException("Pdo\Dblib::__construct() cannot be used for connecting to the \"$driver\" driver");
```
nicolas-grekas: is this how the native classes behave?

nicolas-grekas (commented):
Should we add the static connect method to every child class, since we can polyfill it there?
How does this method behave when given a DSN for another driver and when it's called on a child class?

jnoordsij:
> Should we add the static connect method to every child class, since we can polyfill it there? How does this method behave when given a DSN for another driver and when it's called on a child class?

We could, but that feels a bit off to me. The whole idea of the `PDO::connect` static method is that it is able to detect the type of driver and thus subclass to use from the provided dsn, making it generic. Calling e.g. `\Pdo\Mysqlite::connect(...)` instead of just `new \Pdo\Mysqlite(...)` makes little sense to me, and probably should not be something to actively encourage in any userland code?

src/Php84/Resources/stubs/Pdo/Sqlite.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        public const ATTR_EXTENDED_RESULT_CODES = \PHP_VERSION_ID >= 70400 ? \PDO::SQLITE_ATTR_EXTENDED_RESULT_CODES : 1002;
+        public const ATTR_OPEN_FLAGS = \PHP_VERSION_ID >= 70300 ? \PDO::SQLITE_ATTR_OPEN_FLAGS : 1000;
+        public const ATTR_READONLY_STATEMENT = \PHP_VERSION_ID >= 70400 ? \PDO::SQLITE_ATTR_READONLY_STATEMENT : 1001;
+        public const DETERMINISTIC = PDO::SQLITE_DETERMINISTIC;
+        public const OPEN_READONLY = \PHP_VERSION_ID >= 70300 ? \PDO::SQLITE_OPEN_READONLY : 1;
+        public const OPEN_READWRITE = \PHP_VERSION_ID >= 70300 ? \PDO::SQLITE_OPEN_READWRITE : 2;
+        public const OPEN_CREATE = \PHP_VERSION_ID >= 70300 ? \PDO::SQLITE_OPEN_CREATE : 4;
+
+        public function __construct(
```
nicolas-grekas: CS note: all signatures should be put on one line

src/Php84/Resources/stubs/Pdo/Dblib.php
jnoordsij: Yeah, this is described in the RFC (see https://wiki.php.net/rfc/pdo_driver_specific_subclasses#add_a_way_of_creating_them_through_pdo_static_factory_method) and is also verified here as tests also run on PHP 8.5, expecting the same outcome on polyfilled and non-polyfilled calls.

src/Php84/Resources/stubs/Pdo/Mysql.php
```diff
@@ -0,0 +1,54 @@
...
+    class Mysql extends PDO
+    {
+        public const ATTR_COMPRESS = PDO::MYSQL_ATTR_COMPRESS;
+        public const ATTR_DIRECT_QUERY = PDO::MYSQL_ATTR_DIRECT_QUERY;
+        public const ATTR_FOUND_ROWS = PDO::MYSQL_ATTR_FOUND_ROWS;
+        public const ATTR_IGNORE_SPACE = PDO::MYSQL_ATTR_IGNORE_SPACE;
+        public const ATTR_INIT_COMMAND = PDO::MYSQL_ATTR_INIT_COMMAND;
+        public const ATTR_LOCAL_INFILE = PDO::MYSQL_ATTR_LOCAL_INFILE;
+        public const ATTR_LOCAL_INFILE_DIRECTORY = \PHP_VERSION_ID >= 80100 ? \PDO::MYSQL_ATTR_LOCAL_INFILE_DIRECTORY : 1015;
+        public const ATTR_MAX_BUFFER_SIZE = PDO::MYSQL_ATTR_MAX_BUFFER_SIZE;
```
jnoordsij: https://www.php.net/manual/en/class.pdo-mysql.php#pdo-mysql.constants.attr-max-buffer-size

This constant (and also `ATTR_READ_DEFAULT_FILE` and `ATTR_READ_DEFAULT_GROUP`) are not available when `mysqlnd` is used. Unfortunately, this means an exception is thrown when a class instance is constructed in such cases.

Options:
- just use a numeric value rather than refer to potentially defined value
- find some way to fallback if not defined (probably impossible as what can be done within the constant declaration seems to be very limited)
- leave them out

nicolas-grekas: we can should define two classes, in an if/else then, to mimic what PHP does

nicolas-grekas:
About the connect method on child classes, I'd still add it, for completness of the polyfill

src/Php84/Resources/stubs/Pdo/Odbc.php
```diff
@@ -0,0 +1,40 @@
...
+
+            if (($driver = $this->getAttribute(PDO::ATTR_DRIVER_NAME)) !== 'odbc')
+            {
+                throw new PDOException("Pdo\Odbc::__construct() cannot be used for connecting to the \"$driver\" driver");
+            }
+        }
+
+        public static function connect(string $dsn, ?string $username = null, ?string $password = null, ?array $options = null): static
+        {
+            return new static($dsn, $username, $password, $options);
```
jnoordsij: For consistency these also need some driver check to be on par with the PHP 8.4+ message:
`Pdo\Dblib::connect() cannot be used for connecting to the "sqlite" driver`

jnoordsij:
I probably will look into the last bits next week; if anyone is eager to finalize on the `connect` method or think of additional tests to throw in here, feel free to add them.

src/Php84/Resources/stubs/Pdo/Odbc.php
jnoordsij: Also the signature should not use `static` as PHP 7.2 does not know it

nicolas-grekas: we could use `static` when using php >= 7.3 then (or get_called_class() if the syntax isn't allowed back then)

derrabus:
> Calling e.g. `\Pdo\Mysqlite::connect(...)` instead of just `new \Pdo\Mysqlite(...)` makes little sense to me, and probably should not be something to actively encourage in any userland code?

But if you are on PHP 8.4, would you really call the constructor of those subclasses instead of `PDO::connect()`? So either way, we're encouraging userland code that is different with the polyfill than with the real thing, mainly because we cannot polyfill `PDO::connect()`. 😕

src/Php84/Resources/stubs/Pdo/Odbc.php
jnoordsij: I figured that `self` is actually the more accurate choice here anyways, as it seems that extending the classes on PHP 8.4 then using `connect` will also return `self` and never `static`. Added a very simple test to ensure this behavior across versions.

jnoordsij:
> But if you are on PHP 8.4, would you really call the constructor of those subclasses instead of `PDO::connect()`? So either way, we're encouraging userland code that is different with the polyfill than with the real thing, mainly because we cannot polyfill `PDO::connect()`. 😕

To be fair, if one is currently only using something like `new PDO("sqlite:foo");`, then migrating towards `new Pdo\Sqlite("sqlite:foo");` does not seem like a weird strategy to me. In particular for projects were the driver class is already predetermined and you want to have the added benefit of stricter types for static analysis, one could prefer this over `PDO::connect` which will from a static point of view only ever return `PDO`. However I do agree that the main usecase would still be to used `PDO::connect`, which unfortunately is not something we can provide.

jnoordsij:
I think this should be complete now as is. Additional feature-like tests could be nice, but on the other hand are probably very hard to implement in practice without any real database and I think the current set should suffice to prove this should work in practice. In particular the migration towards 'polyfilled' constants should be very safe to do with this.

jnoordsij:
@nicolas-grekas with PHP 8.5 stable coming up next week, is there any chance of aiming to review and release this somewhere within that timeframe?

src/Php84/Resources/stubs/Pdo/Dblib.php
```diff
@@ -0,0 +1,48 @@
...
+        public const ATTR_VERSION = PDO::DBLIB_ATTR_VERSION;
+        public const ATTR_TDS_VERSION = \PHP_VERSION_ID >= 70300 ? PDO::DBLIB_ATTR_TDS_VERSION : 1004;
+        public const ATTR_SKIP_EMPTY_ROWSETS = \PHP_VERSION_ID >= 70300 ? PDO::DBLIB_ATTR_SKIP_EMPTY_ROWSETS : 1005;
+        public const ATTR_DATETIME_CONVERT = \PHP_VERSION_ID >= 70300 ? \PDO::DBLIB_ATTR_DATETIME_CONVERT : 1006;
+
+        public function __construct(string $dsn, ?string $username = null, ?string $password = null, ?array $options = null)
+        {
+            parent::__construct($dsn, $username, $password, $options);
+
+            if (($driver = $this->getAttribute(PDO::ATTR_DRIVER_NAME)) !== 'dblib')
```
nicolas-grekas:
```suggestion
            if ('dblib' !== $driver = $this->getAttribute(PDO::ATTR_DRIVER_NAME))
```

src/Php84/Resources/stubs/Pdo/Dblib.php
```diff
@@ -0,0 +1,48 @@
...
+            {
+                throw new PDOException("Pdo\Dblib::__construct() cannot be used for connecting to the \"$driver\" driver");
```
nicolas-grekas:
```suggestion
                throw new PDOException(\sprintf('Pdo\Dblib::__construct() cannot be used for connecting to the "%s" driver', $driver));
```

src/Php84/Resources/stubs/Pdo/Dblib.php
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+
+        public static function connect(string $dsn, ?string $username = null, ?string $password = null, ?array $options = null): self
+        {
+            try {
+                return new self($dsn, $username, $password, $options);
+            } catch (PDOException $e) {
+                throw preg_match('/Pdo\\\\Dblib::__construct\(\) cannot be used for connecting to the "([a-z]+)" driver/', $e->getMessage(), $matches)
+                    ? new PDOException("Pdo\Dblib::connect() cannot be used for connecting to the \"{$matches[1]}\" driver")
```
nicolas-grekas:
```suggestion
                throw preg_match('/^Pdo\\\\Dblib::__construct\(\) cannot be used for connecting to the "([a-z]+)" driver/', $e->getMessage(), $matches)
                    ? new PDOException(\sprintf('Pdo\Dblib::connect() cannot be used for connecting to the "%s" driver', $matches[1]))
```

nicolas-grekas (commented):
Can you please proceed with these CS changes on all classes?

jnoordsij:
> Can you please proceed with these CS changes on all classes?

Done!

jnoordsij:
Rebased to fix conflicts

nicolas-grekas:
Thank you @jnoordsij.

PR #554, opened 2025-11-19
[8.6] Add `clamp` function

kylekatarnls:
RFC: https://wiki.php.net/rfc/clamp_v2
Original implementation: https://github.com/php/php-src/pull/19434
Documentation: https://github.com/php/doc-en/pull/4814

### DIFF
--- a/src/Php86/Php86.php
+++ b/src/Php86/Php86.php
@@ -1,0 +1,65 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Php86;
+
+/**
+ * @author kylekatarnls <kylekatarnls@gmail.com>
+ *
+ * @internal
+ */
+final class Php86
+{
+    /**
+     * @template Value
+     * @template Minimum
+     * @template Maximum
+     *
+     * @param Value   $value
+     * @param Minimum $min
+     * @param Maximum $max
+     *
+     * @return Value|Minimum|Maximum
+     */
+    public static function clamp($value, $min, $max)
+    {
+        if (\is_float($min) && is_nan($min)) {
+            self::throwValueErrorIfAvailable('clamp(): Argument #2 ($min) must not be NAN');
+        }
+
+        if (\is_float($max) && is_nan($max)) {
+            self::throwValueErrorIfAvailable('clamp(): Argument #3 ($max) must not be NAN');
+        }
+
+        if ($max < $min) {
+            self::throwValueErrorIfAvailable('clamp(): Argument #2 ($min) must be smaller than or equal to argument #3 ($max)');
+        }
+
+        if ($value > $max) {
+            return $max;
+        }
+
+        if ($value < $min) {
+            return $min;
+        }
+
+        return $value;
+    }
+
+    private static function throwValueErrorIfAvailable(string $message): void
+    {
+        if (!class_exists(\ValueError::class)) {
+            throw new \InvalidArgumentException($message);
+        }
+
+        throw new \ValueError($message);
+    }
+}
--- a/src/Php86/bootstrap.php
+++ b/src/Php86/bootstrap.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Php86 as p;
+
+if (\PHP_VERSION_ID >= 80600) {
+    return;
+}
+
+if (\PHP_VERSION_ID >= 80000) {
+    return require __DIR__.'/bootstrap80.php';
+}
+
+if (!function_exists('clamp')) {
+    /**
+     * @template V
+     * @template L
+     * @template H
+     *
+     * @param V $value
+     * @param L $min
+     * @param H $max
+     *
+     * @return V|L|H
+     */
+    function clamp($value, $min, $max) { return p\Php86::clamp($value, $min, $max); }
+}
--- a/tests/Php86/Php86Test.php
+++ b/tests/Php86/Php86Test.php
@@ -1,0 +1,176 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Tests\Php86;
+
+use PHPUnit\Framework\TestCase;
+
+class Php86Test extends TestCase
+{
+    /**
+     * @dataProvider provideValidClampInput
+     */
+    public function testClampSuccess(array $arguments, $result): void
+    {
+        [$value, $min, $max] = $arguments;
+
+        $actual = clamp($value, $min, $max);
+
+        if ($value instanceof \DateTimeImmutable) {
+            $this->assertInstanceOf(\DateTimeImmutable::class, $actual);
+
+            $actual = $actual->format('Y-m-d');
+        }
+
+        $this->assertSame($result, $actual);
+    }
+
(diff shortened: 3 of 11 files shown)


nicolas-grekas:
That would be for PHP 8.6, isn't it?

src/Php85/Php85.php
```diff
@@ -47,4 +47,49 @@ public static function array_last(array $array)
...
+     * @template L
+     * @template H
+     *
+     * @param V $value
+     * @param L $min
+     * @param H $max
+     *
+     * @return V|L|H
+     */
+    public static function clamp(mixed $value, mixed $min, mixed $max): mixed
```
stof: you cannot use the `mixed` type on PHP 7.2

kylekatarnls:
> That would be for PHP 8.6, isn't it?

Correct, moving it.

src/Php85/Php85.php
kylekatarnls: Thanks, fixed

kylekatarnls:
https://github.com/php/php-src/pull/19434 is merged (will land in PHP 8.6)

src/Php86/Php86.php
```diff
@@ -0,0 +1,65 @@
...
+     * @template V
+     * @template L
+     * @template H
+     *
+     * @param V $value
+     * @param L $min
+     * @param H $max
+     *
+     * @return V|L|H
+     */
```
alexandre-daubois: What's the logic behind `L` and `H`? Maybe this could be more explanatory with `TVal`, `TMin`, `TMax`.

src/Php86/bootstrap.php
```diff
@@ -0,0 +1,31 @@
...
+     * @template L
+     * @template H
+     *
+     * @param V $value
+     * @param L $min
+     * @param H $max
+     *
+     * @return V|L|H
+     */
+    function clamp($value, $min, $max) { return p\Php86::clamp($value, $min, $max); }
```
alexandre-daubois, line 34: Let's use the exact signature from upstream?


```suggestion
    function clamp(mixed $value, mixed $min, mixed $max): mixed { return p\Php86::clamp($value, $min, $max); }
```

src/Php86/Php86.php
```diff
@@ -0,0 +1,65 @@
...
+    public static function clamp($value, $min, $max)
+    {
+        if (\is_float($min) && \is_nan($min)) {
+            self::throwValueErrorIfAvailable('clamp(): Argument #2 ($min) cannot be NAN');
+        }
+
+        if (\is_float($max) && is_nan($max)) {
+            self::throwValueErrorIfAvailable('clamp(): Argument #3 ($max) cannot be NAN');
+        }
```
alexandre-daubois: It was merged using `must` in the wording, it should be updated here as well. See https://github.com/php/php-src/blob/f80338cb926a6dbe8007ad62776bdb4f8d6dd047/ext/standard/math.c#L393-L403

src/Php86/Php86.php
kylekatarnls: L = Low, H = High.

It was because min and max both start with an M.

But I'm fine using full words such as Value, Minimum, Maximum.

src/Php86/Php86.php
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @return Value|Minimum|Maximum
+     */
+    public static function clamp($value, $min, $max)
+    {
+        if (\is_float($min) && \is_nan($min)) {
+            self::throwValueErrorIfAvailable('clamp(): Argument #2 ($min) must not be NAN');
+        }
+
+        if (\is_float($max) && is_nan($max)) {
```
Ayesh, line 38:
```suggestion
        if (\is_float($max) && \is_nan($max)) {
```

Nit: FQFN.

Ayesh (commented):
Added a very minor nitpick.

src/Php86/Php86.php
kylekatarnls: `is_nan` is not optimized by Zend: https://github.com/php/php-src/blob/ebef772cfd9116aee0ca801418387cabb6719dbd/Zend/zend_compile.c#L5030

So there is no performance benefit to get adding `\` AFAIK.

kylekatarnls: @Ayesh I think Symfony follows the rule of using FQFN only for optimized functions because I see this:
```php
\is_int($value) || \is_float($value) && !is_nan($value)
```
in [TypeValidator](https://github.com/symfony/symfony/blob/fa807dd48201fb4608aa6e2a68a902c2bd6d446d/src/Symfony/Component/Validator/Constraints/TypeValidator.php#L77)

Ayesh: Sorry, I marked it because the previous check `is_nan($min)` is a FQFN.

src/Php86/Php86.php
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @param Value $value
+     * @param Minimum $min
+     * @param Maximum $max
+     *
+     * @return Value|Minimum|Maximum
+     */
+    public static function clamp($value, $min, $max)
+    {
+        if (\is_float($min) && \is_nan($min)) {
```
kylekatarnls:
```suggestion
        if (\is_float($min) && is_nan($min)) {
```

src/Php86/Php86.php
kylekatarnls: Indeed, fixing the other one. Thanks 🙏

src/Php86/Php86.php
Ayesh: 🚀 thank you.

kylekatarnls:
Hello @fabpot @nicolas-grekas, do you need anything else for this?

lyrixx:
Failures seems related:
```
e 2 errors:

1) Symfony\Polyfill\Tests\Php86\Php86Test::testClamp
TypeError: Argument 1 passed to Symfony\Polyfill\Tests\Php86\clamp() must be an instance of Symfony\Polyfill\Tests\Php86\mixed, integer given, called in /home/runner/work/polyfill/polyfill/tests/Php86/Php86Test.php on line 20

/home/runner/work/polyfill/polyfill/tests/Php86/Php86Test.php:20

2) Symfony\Polyfill\Tests\Php86\Php86Test::testClamp
TypeError: Polyfills enabled, Argument 1 passed to Symfony\Polyfill\Tests\Php86\clamp() must be an instance of Symfony\Polyfill\Tests\Php86\mixed, integer given, called in /home/runner/work/polyfill/polyfill/tests/Php86/Php86Test.php on line 20

/home/runner/work/polyfill/polyfill/tests/Php86/Php86Test.php:20

```

src/Php86/bootstrap.php
```diff
@@ -0,0 +1,31 @@
...
+     * @template L
+     * @template H
+     *
+     * @param V $value
+     * @param L $min
+     * @param H $max
+     *
+     * @return V|L|H
+     */
+    function clamp(mixed $value, mixed $min, mixed $max): mixed { return p\Php86::clamp($value, $min, $max); }
```
stof: We need conditional definitions of this function, to use `mixed` parameter types only on PHP versions where it is supported (as done in the mbstring polyfill for instance, even though this is about union types in this case)

tests/Php86/Php86Test.php
```diff
@@ -0,0 +1,97 @@
...
+
+        $errorMessage = null;
+
+        try {
+            clamp(4, NAN, 6);
+        } catch (\ValueError $error) {
+            $errorMessage = $error->getMessage();
+        }
+
+        $this->assertSame('clamp(): Argument #2 ($min) cannot be NAN', $errorMessage);
```
stof: Tests would be a lot more maintainable by using separate tests for the error cases, and using the PHPUnit `expectException` API in them.

kylekatarnls: The idea (at time of writing) was to have the same assertions in the same order as in https://github.com/php/php-src/pull/19434/changes#diff-8356d1030e629592aa3e60ff1366c8a8ffcd1e77df1729708c0370775d828696 so we can compare and check it's behaving as the native function.

stof: you can keep the same order by putting those separate tests in the same order.

kylekatarnls: Switched to data providers.

src/Php86/bootstrap.php
kylekatarnls: Added.

nicolas-grekas:
Thank you @kylekatarnls.

PR #556, opened 2025-11-19
Fix PHP 7.2 compatibility for PHP 8.4 polyfill

### LINKED ISSUE #551: PHP 8.4 polyfill for Deprecated attribute is not compatible with PHP 8.0
The PHP 8.4 polyfills have a minimum supported PHP version of PHP 7.2.

In the polyfill code for the `Deprecated` attribute uses `readonly` properties, which are only available in PHP 8.1+:
https://github.com/symfony/polyfill-php84/blob/1.x/Resources/stubs/Deprecated.php

I might be missing something, but I have a feeling that this means that using the (polyfilled) `#[Deprecated]` attribute in code which also needs to run on PHP 8.0 will lead to a parse error when the class is loaded via the Composer classmap: https://3v4l.org/f9oRo

I don't think the class using `readonly` and typed properties is a problem on PHP 7.2-7.4 as attributes are not supported, so the class will never be loaded anyway (unless manually requested, in which case I'd call that user-error).

Seldaek:
Closes https://github.com/symfony/polyfill/issues/551

### DIFF
--- a/src/Php84/Resources/Deprecated.php
+++ b/src/Php84/Resources/Deprecated.php
@@ -9,19 +9,12 @@
  * file that was distributed with this source code.
  */
 
-if (\PHP_VERSION_ID < 80100) {
+if (\PHP_VERSION_ID < 80400) {
     #[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_FUNCTION | Attribute::TARGET_CLASS_CONSTANT)]
     final class Deprecated
     {
-        /**
-         * @readonly
-         */
-        public ?string $message;
-
-        /**
-         * @readonly
-         */
-        public ?string $since;
+        public readonly ?string $message;
+        public readonly ?string $since;
 
         public function __construct(?string $message = null, ?string $since = null)
         {
@@ -29,6 +22,4 @@ if (\PHP_VERSION_ID < 80100) {
             $this->since = $since;
         }
     }
-} elseif (\PHP_VERSION_ID < 80400) {
-    require dirname(__DIR__).'/Deprecated.php';
 }


src/Php84/Resources/stubs/Deprecated.php
```diff
@@ -13,8 +13,17 @@
     #[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_FUNCTION | Attribute::TARGET_CLASS_CONSTANT)]
     final class Deprecated
     {
-        public readonly ?string $message;
-        public readonly ?string $since;
+        /**
+         * @readonly
+         * @var ?string
+         */
+        public $message;
```
stof: I would keep using typed properties. Attribute classes should not be autoloaded on PHP 7.x anyway as attributes are not supported.

Regarding the `readonly` keyword, it might make sense to use conditional class definitions to keep them readonly in PHP 8.1+

Seldaek: Yeah the problem I had is that Composer's phar compiler is linting the files and that goes boom.. 

Edit: fixed in https://github.com/composer/composer/pull/12627/files#diff-df7082676f4891241241b23e47108e77b204ff234200bdc4203477d672e1554cR199

Maybe we can merge as is and undo it in https://github.com/symfony/polyfill/pull/496 ?

Seldaek: Ok I did some changes.. not a pro here with this conditional def mess, so please review carefully :D

stof: The composer compiler already have a list of files excluded from linting to account for the fact that Symfony polyfill packages contain some files loaded conditionally: https://github.com/composer/composer/blob/bfc0e3143841dd156d1099a9c6080e1870b470d2/src/Composer/Compiler.php#L192-L199

That's where the case of typed properties need to be handled IMO (the linter would stick complain about one of the files when using conditional definition files anyway, as the file targeting new PHP versions would still use the new syntaxes)

stof:
The conditional class using features not supported on PHP 8.0 must be extracted in a separate file that is loaded conditionally. Otherwise, you will still get a parse error on the file (the `if` doing a PHP version check is resolved at runtime, not before parsing the PHP syntax. It is not a preprocessor macro of C)

Seldaek:
Ok I see.. Then I'm not sure I'll find time for this right now if it gets more complex sorry but got a whole lot of other stuff on my plate.

nicolas-grekas:
Thank you @Seldaek.

PR #558, opened 2025-11-25
Add the grapheme_levenshtein polyfill

sudam802:
This PR implements the grapheme_levenshtein() function for:

- symfony/polyfill-intl-grapheme
- symfony/polyfill-php85 (when intl is loaded)

Includes:
✔ Function implementation  
✔ ICU-compatible segmentation  
✔ Test file: tests/Intl/Grapheme/grapheme_levenshtein.phpt

### DIFF
--- a/src/Intl/Grapheme/Grapheme.php
+++ b/src/Intl/Grapheme/Grapheme.php
@@ -27,6 +27,7 @@ namespace Symfony\Polyfill\Intl\Grapheme;
  * - grapheme_strstr   - Returns part of haystack string from the first occurrence of needle to the end of haystack
  * - grapheme_substr   - Return part of a string
  * - grapheme_str_split - Splits a string into an array of individual or chunks of graphemes
+ * - grapheme_levenshtein - Calculate the grapheme-unit Levenshtein distance between two strings
  *
  * @author Nicolas Grekas <p@tchwork.com>
  *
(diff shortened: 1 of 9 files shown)


src/Intl/Grapheme/Grapheme.php
```diff
@@ -11,7 +11,12 @@
 
 namespace Symfony\Polyfill\Intl\Grapheme;
 
-\define('SYMFONY_GRAPHEME_CLUSTER_RX', ((float) \PCRE_VERSION < 10 ? (float) \PCRE_VERSION >= 8.32 : (float) \PCRE_VERSION >= 10.39) ? '\X' : Grapheme::GRAPHEME_CLUSTER_RX);
+\define('SYMFONY_GRAPHEME_CLUSTER_RX',
+    (false !== @preg_match('/\X/u', "a"))
```
stof: This change looks suspicious to me. We were explicitly checking the PCRE version because older versions of PCRE also have a _broken_ implementation of `\X` (there is even an open discussion mentioning that we might need to increase the bound from which we use the native feature)

src/Intl/Grapheme/Grapheme.php
```diff
@@ -222,6 +227,82 @@ public static function grapheme_str_split($s, $len = 1)
...
+     * @param int    $replacement_cost
+     * @param int    $deletion_cost
+     *
+     * @return int|false
+     *
+     * @see https://wiki.php.net/rfc/grapheme_levenshtein
+     */
+    public static function grapheme_levenshtein($string1, $string2, $insertion_cost = 1, $replacement_cost = 1, $deletion_cost = 1)
+    {
+        // Cast (PHP does this)
```
stof: use native parameter types, so that PHP actually does it.

src/Php85/bootstrap.php
```diff
@@ -30,3 +31,10 @@ function array_first(array $array) { return p\Php85::array_first($array); }
 if (!function_exists('array_last')) {
     function array_last(array $array) { return p\Php85::array_last($array); }
 }
+
+if (\extension_loaded('intl') && !function_exists('grapheme_levenshtein')) {
+    function grapheme_levenshtein($string1, $string2, $insertion_cost = 1, $replacement_cost = 1, $deletion_cost = 1)
+    {
+        return Grapheme::grapheme_levenshtein($string1, $string2, $insertion_cost, $replacement_cost, $deletion_cost);
```
stof: The implementation must be duplicated in the `Php85` package if we want to provide it here, not call the other package.

tests/Intl/Grapheme/grapheme_levenshtein.phpt
```diff
@@ -0,0 +1,28 @@
+<?php
+
+use Symfony\Polyfill\Intl\Grapheme\Grapheme;
+
+return [
```
stof: this file does not make sense. It does not provide a phpt test, and we use PHPUnit tests anyway.

sudam802: I just wanted to make  test file I'll look more deeper in this implementation

stof: the file you added is not a test file at all. It is not testing anything.

sudam802: @stof I want to know is this [function](https://www.php.net/manual/en/function.levenshtein.php) is needed to be implemented in here grapheme and php85

nicolas-grekas:
Thank you @sudam802.

PR #561, opened 2026-04-09
Add polyfill for symfony/php-ext-deepclone

nicolas-grekas:
https://github.com/symfony/php-ext-deepclone

Implementation borrowed from symfony/var-exporter

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,7 @@ jobs:
                 uses: shivammathur/setup-php@v2
                 with:
                     coverage: "none"
-                    extensions: "apcu, intl-73.2, mbstring, odbc, uuid"
+                    extensions: "apcu, intl-73.2, mbstring, odbc, uuid, mongodb"
                     ini-values: "memory_limit=-1, session.gc_probability=0, apc.enable_cli=1"
                     php-version: "${{ matrix.php }}"
                     tools: "composer:v2"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,6 +40,14 @@ jobs:
                     php-version: "${{ matrix.php }}"
                     tools: "composer:v2"
 
+            -   name: Install ext-deepclone via PIE
+                if: ${{ matrix.php >= '8.2' }}
+                run: |
+                    curl -sLO https://github.com/php/pie/releases/latest/download/pie.phar
+                    sudo php pie.phar install symfony/deepclone
+                    php --ri deepclone || echo "extension=deepclone" | sudo tee "$(php -r 'echo PHP_CONFIG_FILE_SCAN_DIR;')/99-deepclone.ini"
+                    php -m | grep -q deepclone && echo "deepclone loaded" || echo "deepclone NOT loaded"
+
             -   name: Install dependencies
                 run: |
                     composer --prefer-source --no-progress --ansi install
--- a/src/DeepClone/DeepClone.php
+++ b/src/DeepClone/DeepClone.php
@@ -1,0 +1,1161 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\DeepClone;
+
+/**
+ * @author Nicolas Grekas <p@tchwork.com>
+ *
+ * @internal
+ */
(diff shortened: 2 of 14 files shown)


src/DeepClone/DeepClone.php
```diff
@@ -0,0 +1,930 @@
...
+ *
+ * @author Nicolas Grekas <p@tchwork.com>
+ *
+ * @internal
+ */
+final class DeepClone
+{
+    /**
+     * Internal classes that round-trip cleanly through PHP's serialize() —
+     * the empty O:N:"X":0:{} skeleton works, no exception thrown — but whose
```
stof: I would also suggest submitting a PR to php-src to mark `ZipArchive` as non-serializable in the next version of PHP instead of this broken serialization losing all state, as you discovered this case.

stof (approved):

src/DeepClone/DeepClone.php
nicolas-grekas: agreed, see https://github.com/php/php-src/issues/21682

src/DeepClone/Resources/stubs/ClassNotFoundException.php
GromNaN, line 1: Do you need the `DeepClone_` prefix? It's the first time we stub namespaced classes, I feel like this prefix is redundant with the package name. 
I suggest: `src/DeepClone/Resources/stubs/ClassNotFoundException.php`

tests/DeepClone/DeepCloneTest.php
```diff
@@ -0,0 +1,1103 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Tests\DeepClone;
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Port of the PHPT test suite shipped with the deepclone PHP extension
```
GromNaN: This is a significant maintenance effort; even if the coding agent can keep the files in sync, it's expensive and subject to divergence. You could add https://github.com/symfony/php-ext-deepclone as a submodule a run the test test suite from the extension.

tests/DeepClone/DeepCloneTest.php
```diff
@@ -0,0 +1,1103 @@
...
+        $x = [123];
+        $d = deepclone_to_array([&$x]);
+
+        $this->assertArrayHasKey('value', $d);
+        $this->assertSame([[123]], $d['value']);
+    }
+
+    public function testToArrayNamedClosureGlobalFunctionWireFormat()
+    {
+        $d = deepclone_to_array(strlen(...));
```
GromNaN: Unsupported syntax on legacy PHP versions:
> PHP Parse error:  syntax error, unexpected ')' in /home/runner/work/polyfill/polyfill/tests/DeepClone/DeepCloneTest.php on line 155

GromNaN (changes requested):

tests/DeepClone/DeepCloneTest.php
nicolas-grekas: I'd prefer not, the test suite in the polyfill repo has special hooks to ensure both C and PHP versions are aligned. phpt wouldn't support that without significant work.

GromNaN: All right. We should definitely run the tests in GHA with the extension installed to verify that the tests pass.

GromNaN (approved):

nicolas-grekas:
(failures unrelated FTR)

stof:
@nicolas-grekas could you add ext-deepclone in the CI setup ?

.github/workflows/tests.yml
```diff
@@ -40,6 +40,14 @@ jobs:
                     php-version: "${{ matrix.php }}"
                     tools: "composer:v2"
```
GromNaN, line 41: You can add `pie` to the tool here
https://github.com/shivammathur/setup-php/pull/948

```suggestion
                    tools: "composer:v2, pie"
```

nicolas-grekas: won't work on the < 8.1 builds...

nicolas-grekas:
> add ext-deepclone in the CI setup

done! (and extension tagged as v0.1.0)

stof:
the CI reports some issues: https://github.com/symfony/polyfill/actions/runs/24245098617/job/70789203397?pr=561#step:6:745

nicolas-grekas:
Artifacts of the way we run the test suite, nothing to care about here.

nicolas-grekas:
Thank you @GromNaN.

PR #562, opened 2026-04-10
[DeepClone] Support internal final classes with __unserialize (e.g. MongoDB BSON types)

GromNaN:
Polyfill for symfony/php-ext-deepclone#4.

The pure-PHP implementation requires an extra step: since we cannot use `newInstanceWithoutConstructor()` and the empty-state `unserialize('O:...:0:{}')` is rejected (e.g. MongoDB's `__unserialize([])` throws on missing keys), we build the full serialized payload from the object's state before calling `unserialize()`.

When the object's state contains no object-reference masks it can be finalized before the properties loop, so that other objects can hold references to it normally.


nicolas-grekas:
thanks!

PR #563, opened 2026-04-10
[DeepClone] Test deepclone of tidyNode (final internal class without serialization API)

GromNaN:
`tidyNode` is a final internal PHP class (bundled `tidy` extension) with a C-level constructor but no `__serialize`/`__unserialize`/`__sleep`/`__wakeup`. It was incorrectly rejected by the polyfill as not instantiable.

This PR adds a test covering this class as the PHP-only counterpart of the PHPT test added in symfony/php-ext-deepclone#5.


nicolas-grekas:
Thanks, merged as 35d7cf1

PR #564, opened 2026-04-10
Add PHP 8.6 to CI configuration

jnoordsij:
Given that PHP 8.6 polyfills are being added, it makes sense to test on this to ensure polyfill behavior matches with implementation behavior.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,6 +25,7 @@ jobs:
                     - php: '8.3'
                     - php: '8.4'
                     - php: '8.5'
+                    - php: '8.6'
             fail-fast: false
 
         steps:
@@ -44,7 +45,7 @@ jobs:
                 if: ${{ matrix.php >= '8.2' }}
                 run: |
                     curl -sLO https://github.com/php/pie/releases/latest/download/pie.phar
-                    sudo php pie.phar install symfony/deepclone
+                    sudo BOX_REQUIREMENT_CHECKER=0 php pie.phar install symfony/deepclone
                     php --ri deepclone || echo "extension=deepclone" | sudo tee "$(php -r 'echo PHP_CONFIG_FILE_SCAN_DIR;')/99-deepclone.ini"
                     php -m | grep -q deepclone && echo "deepclone loaded" || echo "deepclone NOT loaded"
 


jnoordsij:
Ah [PIE does not like this yet](https://github.com/symfony/polyfill/actions/runs/24251184582/job/70810691782?pr=564#step:4:21), that's a bummer...

.github/workflows/tests.yml
```diff
@@ -25,6 +25,7 @@ jobs:
                     - php: '8.3'
                     - php: '8.4'
                     - php: '8.5'
+                    - php: '8.6'
```
nicolas-grekas, line 28: looks like pie is not compatible with php 8.6? /cc @asgrim
(from the CI)

jnoordsij: https://github.com/php/pie/pull/289/changes could be used to bypass this in CI (either in general or just for PHP 8.6)

nicolas-grekas: 🚀

.github/workflows/tests.yml
```diff
@@ -12,6 +12,7 @@ jobs:
 
         env:
             COMPOSER_ROOT_VERSION: 1.x-dev
+            BOX_REQUIREMENT_CHECKER: 0 # https://github.com/php/pie/blob/896aace839a7bfe25de7d3979df413cd8e01b3a5/docs/usage.md?plain=1#L158-L166
```
nicolas-grekas: the comment is not needed, google/git blame/etc will help if ppl wonder about this

jnoordsij: Ah fair. Somehow it doesn't seem to work though...

*edit:* Can always just add it inline I guess...

.github/workflows/tests.yml
```diff
@@ -44,7 +45,7 @@ jobs:
                 if: ${{ matrix.php >= '8.2' }}
                 run: |
                     curl -sLO https://github.com/php/pie/releases/latest/download/pie.phar
-                    sudo php pie.phar install symfony/deepclone
```
jnoordsij, line 47: Not sure why this is still not working. Time for weekend though, will look into this later potentially.

asgrim: Try `sudo -E` perhaps (keep env vars), or put the env after `sudo`

nicolas-grekas:
Thank you @jnoordsij.

.github/workflows/tests.yml
nicolas-grekas: works thanks!

jnoordsij: Ah yeah that makes sense, thanks!

PR #566, opened 2026-04-10
[PHP 8.6] Add `ARRAY_FILTER_USE_VALUE` constant

Ayesh:
Adds the `ARRAY_FILTER_USE_VALUE` constant added as the defualt value for the `array_filter` function's `$mode` parameter in PHP 8.6.

 - [`array_filter`](https://php.watch/codex/array_filter)
 - [`ARRAY_FILTER_USE_VALUE`](https://php.watch/codex/array_filter)
 - [`UPGRADING` entry](https://github.com/php/php-src/blob/3c68fa9adc2b2989483738980d47052b03d82420/UPGRADING#L225)

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ Polyfills are provided for:
 - the `grapheme_levenshtein` function introduced in PHP 8.5;
 - the `locale_is_right_to_left` function introduced in PHP 8.5;
 - the `clamp` function introduced in PHP 8.6;
+- the `ARRAY_FILTER_USE_VALUE` constant introduced in PHP 8.6;
 
 It is strongly recommended to upgrade your PHP version and/or install the missing
 extensions whenever possible. This polyfill should be used only when there is no
--- a/src/Php86/README.md
+++ b/src/Php86/README.md
@@ -4,6 +4,7 @@ Symfony Polyfill / Php86
 This component provides features added to PHP 8.6 core:
 
 - [`clamp`](https://wiki.php.net/rfc/clamp_v2)
+- [`ARRAY_FILTER_USE_VALUE`] constant
 
 More information can be found in the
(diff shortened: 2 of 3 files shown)


nicolas-grekas:
Sure but what's the use case? I think we rejected adding such consts in the past, didn't we?

stof:
If the behavior enabled by that constant in `array_filter` does not work when using the polyfill on older versions, it is better to *not* polyfill the constant so that code can detect the availability of the relevant feature by checking whether the constant is defined (instead of having to detect whether `array_filter` supports that constant).

Ayesh:
Yeah you are right this wouldn't be the most useful polyfill.

With a const declared, it will be possible to safely call `array_filter` with `$mode = ARRAY_FILTER_USE_VALUE` across all PHP versions:


```php
array_filter($arr, $fn, ARRAY_FILTER_USE_VALUE);
```

`ARRAY_FILTER_USE_VALUE` is the default value, so this declaration makes it a less meaningful. 

Please feel free to close this if you think there's no tangible benefit polyflling this constant. I would totally trust your judgement :)

Ayesh:
> If the behavior enabled by that constant in array_filter does not work when using the polyfill on older versions, it is better to not polyfill the constant so that code can detect the availability of the relevant feature by checking whether the constant is defined (instead of having to detect whether array_filter supports that constant).

Right, this is true. If the feature doesn't work, we really shouldn't polyfill it. 

In this case, however, with the constant declared, you can call `array_filter($a, $b, ARRAY_FILTER_USE_VALUE)`, and it _will_ work across all PHP versions. 

The only change in PHP 8.6 is that the default value `$mode = 0` is now set to the new constant (bearing `0`) `ARRAY_FILTER_USE_VALUE`.  But like I said in the other reply, it is the default value, which makes this a bit less useful.

nicolas-grekas:
Thank you @Ayesh.

PR #578, opened 2026-04-19
[RFC] Add grapheme_strrev function php 8.6

PhilDaiguille:
Adds the new grapheme_strrev function added to PHP 8.6. This function is part of the Intl extension's Grapheme functions and reverses a string by grapheme clusters, correctly handling multi-byte and multi code-point characters (such as Emoji sequences).

 - [RFC](https://php.watch/rfcs/grapheme_strrev)
 - [Commit](https://github.com/php/php-src/commit/7a1c2612c01d9bd4d0c2d128fb5b2bf8f0a2e671)
 - [PHP 8.6: New grapheme_strrev function](https://php.watch/versions/8.6/grapheme_strrev)
 - [Codex](https://php.watch/codex/grapheme_strrev)

### DIFF
--- a/src/Intl/Grapheme/Grapheme.php
+++ b/src/Intl/Grapheme/Grapheme.php
@@ -28,6 +28,7 @@ namespace Symfony\Polyfill\Intl\Grapheme;
  * - grapheme_substr   - Return part of a string
  * - grapheme_str_split - Splits a string into an array of individual or chunks of graphemes
  * - grapheme_levenshtein - Calculate the grapheme-unit Levenshtein distance between two strings
+ * - grapheme_strrev - Reverse a string by grapheme clusters
  *
  * @author Nicolas Grekas <p@tchwork.com>
  *
@@ -324,5 +325,16 @@ final class Grapheme
         }
 
         return false !== $needlePos ? self::grapheme_strlen(substr($s, 0, $needlePos)) + $offset : false;
+    }
+
+    public static function grapheme_strrev(string $string)
+    {
+        $units = grapheme_str_split($string);
+
+        if (false === $units) {
+            return false;
+        }
+
+        return implode('', array_reverse($units));
     }
 }
--- a/src/Intl/Grapheme/bootstrap80.php
+++ b/src/Intl/Grapheme/bootstrap80.php
@@ -17,6 +17,9 @@ if (!function_exists('grapheme_str_split')) {
 if (!function_exists('grapheme_levenshtein')) {
     function grapheme_levenshtein(string $string1, string $string2, int $insertion_cost = 1, int $replacement_cost = 1, int $deletion_cost = 1, string $locale = ''): int|false { return p\Grapheme::grapheme_levenshtein($string1, $string2, $insertion_cost, $replacement_cost, $deletion_cost); }
 }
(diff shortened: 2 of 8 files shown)


src/Intl/Grapheme/Grapheme.php
```diff
@@ -325,4 +326,19 @@ private static function grapheme_position($s, $needle, $offset, $mode)
 
         return false !== $needlePos ? self::grapheme_strlen(substr($s, 0, $needlePos)) + $offset : false;
     }
+
+    public static function grapheme_strrev(string $string)
+    {
+        if (!preg_match('//u', $string)) {
```
Ayesh: May I ask what this is for? If a regexp polyfill for `grapheme_str_split` fails, it should also return `false`, so the check below should be enough?

Ayesh (approved):
Hi @PhilDaiguille - thank you for the PR. I'm a regular contributor just like you, but I wrote the article mentioned in the PR. Thank you for the mention. 

I marked one place that I'm not quite sure why there is a `preg_match` check against an empty regex, but other than that, the polyfill and the tests look good to me!

When the `grapheme_str_split` function was added, it was added to _both_ Intl and PHP 8.4 polyfills, so maybe it makes sense to polyfill `grapheme_strrev` for both Intl and PHP 8.6 too? The PHP 8.6 polyfill can use an `extension_loaded('intl')` so it only polyfills if the older PHP version has the "real" Intl extension.

Ayesh (approved):

src/Intl/Grapheme/bootstrap80.php
```diff
@@ -17,6 +17,9 @@ function grapheme_str_split(string $string, int $length = 1): array|false { retu
 if (!function_exists('grapheme_levenshtein')) {
     function grapheme_levenshtein(string $string1, string $string2, int $insertion_cost = 1, int $replacement_cost = 1, int $deletion_cost = 1, string $locale = ''): int|false { return p\Grapheme::grapheme_levenshtein($string1, $string2, $insertion_cost, $replacement_cost, $deletion_cost); }
 }
+if (!function_exists('grapheme_strrev')) {
+    function grapheme_strrev(string $string): string|false { return p\Grapheme::grapheme_strrev($string); }
```
xabbuh, line 21: I guess you need to swap the definitions from the two bootstrap files. This signature will error with PHP 8.0.

stof: @xabbuh I don't understand your comment. `string|false` is a valid return type on PHP 8.0+

PhilDaiguille: Normally this works fine in PHP 8.0 in this case, string|false is a valid return type since PHP 8.0 introduced Union Types which explicitly allows false as a pseudo-type within a union. [https://wiki.php.net/rfc/union_types_v2](https://wiki.php.net/rfc/union_types_v2)

xabbuh: Ah sorry, my mistake. I was confused on when we use which bootstrap file. Nothing to see here.

nicolas-grekas:
Thank you @PhilDaiguille.

PR #582, opened 2026-04-24
[Intl/MessageFormatter] Add missing parseMessage() method

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 1.x
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

Add the static `MessageFormatter::parseMessage()` method to mirror the native intl API. The polyfill does not implement parsing, so the method returns `false`, matching the instance `parse()`.

### DIFF
--- a/src/Intl/MessageFormatter/MessageFormatter.php
+++ b/src/Intl/MessageFormatter/MessageFormatter.php
@@ -150,6 +150,11 @@ class MessageFormatter
         return false;
     }
 
+    public static function parseMessage(string $locale, string $pattern, string $message)
+    {
+        return false;
+    }
+
     private static function parseTokens(array $tokens, array $values, $locale)
     {
         foreach ($tokens as $i => $token) {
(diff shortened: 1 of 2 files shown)


src/Intl/MessageFormatter/MessageFormatter.php
```diff
@@ -148,6 +148,11 @@ public function parse(string $string)
         return false;
     }
 
+    public static function parseMessage(string $locale, string $pattern, string $message)
+    {
+        return static::create($locale, $pattern)->parse($message);
```
stof: as `parse` is not implemented, how do this report missing support in the polyfill ? This static method does not give access to the error getters of the instance.

xabbuh: I am not sure I understand what you mean. Is it that from your POV the exception message raised by `parse()` would be confusing?

nicolas-grekas: My understanding: parse() in the instance will not throw, it will only populate error as state in properties, which are supposed to be read, but no available when using the static API.
`intl_get_error_code()` / `intl_get_error_message()` are the global functions supposed to work for static methods, but here, not sure we can use them since they might be already declared when polyfilling a subpart of intl (guessing, I didn't double check this)
Which means we might prefer throwing as not implemented here and tell to use the non-static API?

nicolas-grekas:
Thank you @xabbuh.

PR #588, opened 2026-05-10
Update `IntlDateFormatter::__construct()` parameters

Seb33300:
Since PHP 8.1, parameters `dateType` and `timeType` are optional.

See: https://www.php.net/manual/en/intldateformatter.create.php#refsect1-intldateformatter.create-changelog

Same as https://github.com/symfony/polyfill-intl-icu/pull/2


nicolas-grekas:
Thank you @Seb33300.

PR #593, opened 2026-05-25
Various fixes

### LINKED ISSUE #415: mb_strlen returns 0 instead of 1 for the char chr(254)
Hi,

With the polyfill, `var_dump(mb_strlen(chr(254)))` return `0`.
With the php8.0-mbstring extension, var_dump(mb_strlen(chr(254))) return `1`;

`versions : * v1.26.0`

Thanks,
Alex

### LINKED ISSUE #506: Concerns about passing null to some polyfill arguments
For years our code-base (like presumably many others) had an existing `mb_trim()` polyfill of our own and as it pre-dates even php 7, it had a slight quirk in that it accepted a `null` as the first argument.

PHP 8.4 will provide its own `mb_trim()` implementation which has its first argument typed as `string`, but which *also* [accepts a `null` argument](https://3v4l.org/A7mv1/rfc#vgit.master), albeit with a deprecation warning.

The symfony polyfill on the other hand does not accept a null argument and thus, it being an userland function will throw a `TypeError` instead.

In the case of our code-base, this caused an issue where a completely non-related `composer update` now turned code which run perfectly well (and would have changed to raising a deprecation notice in 8.4) into a `TypeError`.

I know that this is probably a slight edge-case, but I'm still coming here to raise the issue because it might affect other users. 

IMHO, to be a proper polyfill for PHP 8.4 `mb_trim`, your implementation should accept `null`s too and raise an E_USER_DEPRECATED error if a `null` is passed.

And the same is probably true for many other functions as well.

### LINKED ISSUE #511: The Grapheme polyfill is failing tests when using `\X` as regex for clusters
See CI jobs for PHP 7.3+ which are failing a test when using the polyfill:

```php
$this->assertSame('☎', grapheme_substr('☢☎❄', 1, 1));
```

`\X` does not split those as separate grapheme clusters.

### LINKED ISSUE #540: str_increment wrong for numbers larger PHP_INT_MAX/PHP_INT_MIN
```
var_dump( str_increment('9223372036854775820') );
```

Additionally, the return should have a (string) type cast to be strictly correct

EDIT: same for str_decrement?

### LINKED ISSUE #550: PHP 8.3 polyfill for ldap_exop_sync() is not compatible with PHP 7.2
The PHP 8.3 polyfills have a minimum supported PHP version of PHP 7.2.

The polyfill code for `ldap_exop_sync()`:
https://github.com/symfony/polyfill-php83/blob/dbdcdf1a4dcc2743591f1079d0c35ab1e2dcbbc9/bootstrap.php#L43

... passes the `$controls` parameter, while that parameter did not exist on the function in PHP 7.2.

This will (should) result in a `Warning: ldap_exop() expects at most 3 parameters, 4 given`.

Ref: https://www.php.net/manual/en/function.ldap-exop.php

### LINKED ISSUE #567: polyfill-uuid: add deprecation warnings for deprecated constants
The Uuid polyfill currently marks 2 constants as deprecated, see https://github.com/symfony/polyfill/blob/3db8459472865f8855eca39f894aeddaa547a0c0/src/Uuid/bootstrap.php#L43-L48 and https://github.com/symfony/polyfill/blob/3db8459472865f8855eca39f894aeddaa547a0c0/src/Uuid/bootstrap80.php#L35-L40

On PHP 8.5+, it is possible to actually deprecate global constants using the deprecated attribute, but such code won't parse successfully on PHP <8.5. Can I suggest one of

* on 8.5, have a different bootstrap file that uses `const` for const declarations, and uses the attributes
* using eval on PHP 8.5+ to add the attributes
* using https://packagist.org/packages/danielescherzer/define-deprecated which I just released (though I'll need to update it to support PHP 7.2+ instead of PHP 8.1+)

I'm happy to do the work to implement any of these if the patches would be accepted

nicolas-grekas:
Fix #550
Fix #540
Fix #506
Fix #415
Fix #567
Fix #511


nicolas-grekas:
Split in as many PRs

PR #601, opened 2026-05-25
Match polyfill signatures to native PHP 8.x to silence TestListener warnings

nicolas-grekas:
- `apcu_fetch`: add `mixed` return type matching native PHP 8.0+.
- Grapheme functions: route PHP 8.5+ through a new `bootstrap85.php` exposing the `$locale` parameter added in newer ICU.
- Php83 `bootstrap81.php`: re-declare `json_validate`, `str_increment`, `str_decrement` and `mb_str_pad` so the TestListener can find polyfills to verify on PHP 8.1+.

### DIFF
--- a/src/Intl/Grapheme/bootstrap85.php
+++ b/src/Intl/Grapheme/bootstrap85.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+// The $locale argument added to grapheme_*() in PHP 8.5 / ICU 74 is accepted
+// for native-signature compatibility but ignored by the polyfill, which only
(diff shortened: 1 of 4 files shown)


src/Intl/Grapheme/bootstrap85.php
```diff
@@ -0,0 +1,37 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Polyfill\Intl\Grapheme as p;
+
+if (!function_exists('grapheme_stripos')) {
+    function grapheme_stripos(string $haystack, string $needle, int $offset = 0, string $locale = ''): int|false { return p\Grapheme::grapheme_stripos($haystack, $needle, $offset); }
```
stof: shouldn't we implement the locale-aware logic instead of ignoring the locale ?

nicolas-grekas: that might be too involving - not for this PR for sure

PR #610, opened 2026-05-25
[Mbstring] Fall back to plain iconv when //IGNORE is unsupported (on musl)

### LINKED ISSUE #387: iconv(): Wrong encoding, conversion from "ASCII" to "UTF-8//IGNORE" is not allowed
Hello,

I noticed that there is an incompatibility with the mbstring polyfill and PHP 8.1 / Alpine Linux,
which breaks a lot of my projects as soon as the php81-mbstring is not installed, but php81-iconv is installed:

Example:

```
Warning: iconv(): Wrong encoding, conversion from "ASCII" to "UTF-8//IGNORE" is not allowed in phar:///var/www/localhost/htdocs/phpstan.phar/vendor/symfony/polyfill-mbstring/Mbstring.php on line 736
```

It looks like `//IGNORE` is not accepted since `echo iconv('UTF-8', 'UTF-8', 'test');` works, while `echo iconv('UTF-8', 'UTF-8//IGNORE', 'test');` doesn't

nicolas-grekas:
## Summary

- Route all internal `iconv(..., '...//IGNORE', ...)` calls through a small wrapper that probes once whether the underlying iconv accepts the `//IGNORE` suffix and drops it when unsupported.
- Fixes spurious `iconv(): Wrong encoding, conversion from "X" to "Y//IGNORE" is not allowed` warnings on Alpine/musl for valid input.
- Invalid input still surfaces a warning (now from the underlying conversion itself), which matches the existing guidance to install ext-mbstring on Alpine.

Fixes #387

## Test plan

- [x] `./phpunit tests/Mbstring` (330 tests, all green)
- [x] New test `testConvertEncodingWithoutIconvIgnoreSupport` forces the probe flag off via reflection, asserts the conversion succeeds and emits zero warnings.

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -82,6 +82,7 @@ final class Mbstring
     private static $encodingList = ['ASCII', 'UTF-8'];
     private static $language = 'neutral';
     private static $internalEncoding = 'UTF-8';
+    private static $iconvSupportsIgnore;
 
     public static function mb_convert_encoding($s, $toEncoding, $fromEncoding = null)
     {
@@ -116,7 +117,7 @@ final class Mbstring
                 $fromEncoding = 'Windows-1252';
             }
             if ('UTF-8' !== $fromEncoding) {
-                $s = iconv($fromEncoding, 'UTF-8//IGNORE', $s);
+                $s = self::iconv($fromEncoding, 'UTF-8', $s);
             }
 
             return preg_replace_callback('/[\x80-\xFF]+/', [__CLASS__, 'html_encoding_callback'], $s);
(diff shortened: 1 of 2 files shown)


GPHemsley:
I think this breaks the underlying assumption that `mb_substitute_character()` is always `none`. By allowing the use of IGNORE-less iconv, operations like `mb_convert_encoding()` and `mb_scrub()` may allow through invalid invalid characters.

I'm not sure it's a good idea to leak implementation details of the polyfill like `iconv(): Detected an illegal character in input string` rather than handle them directly?

nicolas-grekas:
That's a best effort / possibly good enough, until someone has a use case for investing more time on this.

PR #612, opened 2026-05-25
[IoPoll] Polyfill the Io\Poll API

nicolas-grekas:
Tracks the [Polling API RFC](https://wiki.php.net/rfc/poll_api) targeted at PHP 8.6.

Shipped as a dedicated `symfony/polyfill-io-poll` package (`src/Io/Poll/`), requiring PHP >= 8.1 (enums); the split repository and Packagist entry will need to be created before the next release. Surface:

- `Io\IoException`, `Io\Poll\PollException`, abstract `FailedPollOperationException` (with `ERROR_*` constants), and the seven specialized exceptions (`Failed{ContextInitialization,HandleAdd,WatcherModification,PollWait}Exception`, `BackendUnavailableException`, `InactiveWatcherException`, `HandleAlreadyWatchedException`, `InvalidHandleException`).
- `Io\Poll\Backend` enum with `Auto`, `Poll`, `Epoll`, `Kqueue`, `EventPorts`, `WSAPoll` cases and `isAvailable()` / `supportsEdgeTriggering()` / `getAvailableBackends()`.
- `Io\Poll\Event` enum with `Read`, `Write`, `Error`, `HangUp`, `ReadHangUp`, `OneShot`, `EdgeTriggered`.
- `Io\Poll\Handle` marker interface (annotated `@internal` to discourage userland implementations).
- Final `Io\Poll\Context`, `Io\Poll\Watcher`, and global `StreamPollHandle`, with private constructors, clone/double-construct/serialization guards throwing the native errors, and empty `__debugInfo()` like native; the watcher references its context through a `WeakReference` so that destroying the context deactivates its watchers like native refcounting does.

Backed by `stream_select()`, so only the `Poll` backend is available; `Auto` resolves to it and the other backends throw `BackendUnavailableException`. Behavior mirrors the native `poll` backend of `io_poll.c` / `poll_backend_poll.c`: exception messages and codes match, `Event::EdgeTriggered` is rejected with `ERROR_NOSUPPORT`, closed streams are dropped like `POLLNVAL`, a fired `OneShot` watcher stays active with `modify*()` throwing `ERROR_NOTFOUND`, `HangUp` (detected with `stream_socket_recvfrom(..., STREAM_PEEK)` and `feof()`) is reported alongside `Read` like `POLLIN|POLLHUP`, even when not watched, and never for regular files or datagram sockets, and a connection reset (peer closed with unread data, detected when peeking a readable socket fails) reports `Error|HangUp` like `POLLERR|POLLHUP`. `Event::ReadHangUp` is never emitted, matching the native `poll` backend. Known divergences are documented in the component README: TCP half-close reports `Read|HangUp` where native reports plain `Read`.

The test suite combines 64 unit tests with the 28 phpt tests of the native implementation, borrowed verbatim from php-src (`ext/standard/tests/poll`) and executed against the polyfill through a small PHPUnit runner; the phpt suite resolves its backend-specific expectations against the `Poll` backend, which is exactly what the polyfill emulates.

### DIFF
--- a/src/Io/Poll/Resources/stubs/Io/Poll/Context.php
+++ b/src/Io/Poll/Resources/stubs/Io/Poll/Context.php
@@ -1,0 +1,286 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Io\Poll;
+
+if (\PHP_VERSION_ID < 80600) {
+    final class Context
+    {
+        private Backend $backend;
+
+        /**
+         * The fd table of the emulated poll backend, keyed by stream resource id.
+         *
+         * @var array<int, Watcher>
+         */
+        private array $watchers = [];
+
+        public function __construct(Backend $backend = Backend::Auto)
+        {
+            if (isset($this->backend)) {
+                throw new \Error('Io\Poll\Context object is already constructed');
+            }
+
+            if (!$backend->isAvailable()) {
+                throw new BackendUnavailableException(\sprintf('Backend %s not available', $backend->name));
+            }
+
+            $this->backend = Backend::Auto === $backend ? Backend::Poll : $backend;
+        }
+
+        public function __destruct()
+        {
+            static $deactivate;
+            $deactivate ??= \Closure::bind(
+                static function (Watcher $w): void {
+                    $w->active = false;
+                    $w->context = null;
+                },
+                null,
+                Watcher::class,
+            );
+
+            foreach ($this->watchers as $watcher) {
+                $deactivate($watcher);
+            }
+        }
+
+        public function getBackend(): Backend
+        {
+            return $this->backend;
+        }
+
+        public function add(Handle $handle, array $events, mixed $data = null): Watcher
+        {
+            if (!$handle instanceof \StreamPollHandle) {
+                throw new InvalidHandleException(\sprintf('Handle of type "%s" is not supported by the polyfill; only %s is.', \get_class($handle), \StreamPollHandle::class));
+            }
+
+            $stream = $handle->getStream();
+            if (!\is_resource($stream) || 'MEMORY' === stream_get_meta_data($stream)['stream_type']) {
+                throw new InvalidHandleException('Invalid handle for polling');
+            }
+
+            static $normalize, $create;
+            $normalize ??= \Closure::bind(
+                static fn (array $events, string $method, int $argument): array => Watcher::normalizeEvents($events, $method, $argument),
+                null,
+                Watcher::class,
+            );
+            $create ??= \Closure::bind(
+                static fn (\WeakReference $ctx, Handle $h, array $e, mixed $d): Watcher => new Watcher($ctx, $h, $e, $d),
+                null,
+                Watcher::class,
+            );
+
+            $events = $normalize($events, __METHOD__, 2);
+
+            if (\in_array(Event::EdgeTriggered, $events, true)) {
+                throw new FailedHandleAddException('Failed to add handle', FailedPollOperationException::ERROR_NOSUPPORT);
+            }
+
+            $id = get_resource_id($stream);
+            if (null !== $existing = $this->watchers[$id] ?? null) {
+                if (\is_resource($existing->getHandle()->getStream())) {
+                    throw new HandleAlreadyWatchedException('Handle already added');
+                }
+                unset($this->watchers[$id]); // the resource id was recycled from a closed stream
+            }
+
+            return $this->watchers[$id] = $create(\WeakReference::create($this), $handle, $events, $data);
+        }
+
+        public function wait(?int $timeoutSeconds = null, int $timeoutMicroseconds = 0, ?int $maxEvents = null): array
+        {
+            if (null !== $timeoutSeconds) {
+                if ($timeoutSeconds < 0) {
+                    throw new \ValueError(\sprintf('%s(): Argument #1 ($timeoutSeconds) must be greater than or equal to 0', __METHOD__));
+                }
+                if ($timeoutMicroseconds < 0) {
+                    throw new \ValueError(\sprintf('%s(): Argument #2 ($timeoutMicroseconds) must be greater than or equal to 0', __METHOD__));
+                }
+            }
+
+            if (null !== $maxEvents && $maxEvents <= 0) {
+                throw new \ValueError(\sprintf('%s(): Argument #3 ($maxEvents) must be greater than 0', __METHOD__));
+            }
+
+            // Like poll() reporting POLLNVAL, drop watchers whose stream has been
+            // closed; poll() returns immediately in that case, without sleeping
+            $evicted = false;
+            foreach ($this->watchers as $id => $watcher) {
+                if (!\is_resource($watcher->getHandle()->getStream())) {
+                    unset($this->watchers[$id]);
+                    $evicted = true;
+                }
+            }
+
+            if (!$this->watchers) {
+                if (!$evicted && null !== $timeoutSeconds && 0 < $micros = $timeoutSeconds * 1_000_000 + $timeoutMicroseconds) {
+                    usleep($micros);
+                }
+
+                return [];
+            }
+
+            $read = $write = [];
+            foreach ($this->watchers as $id => $watcher) {
+                $stream = $watcher->getHandle()->getStream();
+
+                // poll() reports hang-ups and errors even when not requested;
+                // detecting them requires select() readability for every stream,
+                // at the cost of spurious wake-ups that the retry loop below absorbs
+                $read[$id] = $stream;
+
+                if (\in_array(Event::Write, $watcher->getWatchedEvents(), true)) {
+                    $write[$id] = $stream;
+                }
+            }
+
+            if ($evicted) {
+                $deadline = 0;
+            } elseif (null !== $timeoutSeconds) {
+                $deadline = hrtime(true) + ($timeoutSeconds * 1_000_000 + $timeoutMicroseconds) * 1000;
+            } else {
+                $deadline = null;
+            }
+
+            $triggered = [];
+            while (true) {
+                $r = $read;
+                $w = $write;
+                $e = null;
+
+                if (null === $deadline) {
+                    $result = @stream_select($r, $w, $e, null);
+                } else {
+                    $remaining = max(0, $deadline - hrtime(true));
+                    $result = @stream_select($r, $w, $e, \intdiv($remaining, 1_000_000_000), \intdiv($remaining % 1_000_000_000, 1000));
+                }
+
+                if (false === $result) {
+                    throw new FailedPollWaitException('Poll wait failed');
+                }
+
+                if (0 === $result) {
+                    return [];
+                }
+
+                foreach ($this->watchers as $id => $watcher) {
+                    $isReadable = isset($r[$id]);
+                    $isWritable = isset($w[$id]);
+
+                    if (!$isReadable && !$isWritable) {
+                        continue;
+                    }
+
+                    $stream = $watcher->getHandle()->getStream();
+                    $watched = $watcher->getWatchedEvents();
+                    $hangUp = $error = false;
+                    $readable = $isReadable;
+
+                    if ($isReadable) {
+                        $peek = @stream_socket_recvfrom($stream, 1, \STREAM_PEEK);
+                        $meta = stream_get_meta_data($stream);
+                        if (false === $peek) {
+                            if (str_contains($meta['stream_type'], 'socket')) {
+                                // peeking a readable socket only fails when the connection
+                                // was reset (peer closed with unread data): POLLERR|POLLHUP
+                                $error = $hangUp = true;
+                            } elseif (!$meta['seekable'] && feof($stream)) {
+                                // a pipe at EOF raises POLLHUP without POLLIN,
+                                // while a regular file always raises plain POLLIN
+                                $hangUp = true;
+                                $readable = false;
+                            }
+                        } elseif ('' === $peek && !\in_array($meta['stream_type'], ['udp_socket', 'udg_socket'], true)) {
+                            // a stream socket at EOF raises POLLIN|POLLHUP,
+                            // while an empty datagram is plain POLLIN
+                            $hangUp = true;
+                        }
+                    }
+
+                    $events = [];
+                    if ($readable && \in_array(Event::Read, $watched, true)) {
+                        $events[] = Event::Read;
+                    }
+                    if ($isWritable && \in_array(Event::Write, $watched, true)) {
+                        $events[] = Event::Write;
+                    }
+                    if ($error) {
+                        $events[] = Event::Error; // reported even when not watched, like POLLERR
+                    }
+                    if ($hangUp) {
+                        $events[] = Event::HangUp; // reported even when not watched, like POLLHUP
+                    }
+
+                    if ($events) {
+                        $triggered[] = [$watcher, $events];
+                        if (null !== $maxEvents && \count($triggered) === $maxEvents) {
+                            break;
+                        }
+                    }
+                }
+
+                if ($triggered) {
+                    break;
+                }
+
+                // Readiness that maps to no watched event; poll() would not have
+                // woken up, so keep waiting for the remaining timeout
+                if (null !== $deadline && hrtime(true) >= $deadline) {
+                    return [];
+                }
+                usleep(1000);
+            }
+
+            static $setTriggered;
+            $setTriggered ??= \Closure::bind(
+                static fn (Watcher $w, array $events) => $w->triggeredEvents = $events,
+                null,
+                Watcher::class,
+            );
+
+            $result = [];
+            foreach ($triggered as [$watcher, $events]) {
+                $setTriggered($watcher, $events);
+                $result[] = $watcher;
+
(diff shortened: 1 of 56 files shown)


nicolas-grekas:
@symfony/mergers this one is ready for review!

alexandre-daubois (commented):
Nothing to say on my side, looks good!

alexandre-daubois (approved):

src/Io/Poll/Resources/stubs/Io/Poll/Context.php
```diff
@@ -0,0 +1,232 @@
...
+                foreach ($watcher->getWatchedEvents() as $event) {
+                    if (Event::OneShot === $event) {
+                        unset($this->watchers[$watcher]);
+                        $clearContext($watcher);
+                        break;
+                    }
+                }
+            }
+
+            return $result;
```
GromNaN, line 263: The PHP stub marks `Context`, `Watcher`, and `StreamPollHandle` as `@not-serializable`. In native PHP that annotation generates `zend_throw_error()`, which surfaces in userland as `\Error` — not `\Exception`. Throwing `new \Exception()` here (and in the same methods of `Watcher` and `StreamPollHandle`) means that `catch (\Error $e)` blocks will not catch the polyfill's exception, creating a silent behavioural difference:

```php
try { serialize(new Context()); } catch (\Error $e) { /* never reached in polyfill */ }
```

Fix: throw `new \Error(...)` in all three `__serialize()` / `__unserialize()` methods, and update the test's `expectException(\Exception::class)` to `expectException(\Error::class)`.

src/Io/Poll/Resources/stubs/Io/Poll/Watcher.php
```diff
@@ -0,0 +1,107 @@
...
+                null,
+                Context::class,
+            );
+            $context = $this->context;
+            $this->context = null;
+            $detach($context, $this);
+        }
+
+        public function __serialize(): array
+        {
```
GromNaN, line 119: Same `\Exception` vs `\Error` issue as in `Context` (see the review comment there). Should throw `new \Error("Serialization of 'Io\\Poll\\Watcher' is not allowed")`.

src/Php86/Resources/stubs/StreamPollHandle.php
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public function isValid(): bool
+        {
+            return is_resource($this->stream);
+        }
+
+        public function __serialize(): array
+        {
+            throw new Exception("Serialization of 'StreamPollHandle' is not allowed");
```
GromNaN: Same `\Exception` vs `\Error` issue as in `Context` (see the review comment there). Should throw `new \Error("Serialization of 'StreamPollHandle' is not allowed")`.

src/Php86/Resources/stubs/Io/Poll/Watcher.php
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        public function isActive(): bool
+        {
+            return null !== $this->context;
+        }
+
+        public function modify(array $events, mixed $data = null): void
+        {
+            $this->ensureActive();
```
GromNaN: `Context::add()` validates that `$events` is non-empty and that every element is an `Event` instance, throwing `\TypeError` on violation. `modify()` and `modifyEvents()` skip all validation and overwrite `$this->events` directly. This means:

```php
$watcher->modifyEvents([]);            // silently accepted
$watcher->modifyEvents(['not-event']); // silently accepted
```

The corrupted watcher then silently watches no events (or wrong events) on the next `wait()` call. The same event-array validation from `add()` should be applied here.

src/Php86/Resources/stubs/Io/Poll/Context.php
```diff
@@ -0,0 +1,232 @@
...
+            foreach ($byId as $id => $watcher) {
+                $isReadable = isset($readCopy[$id]);
+                $isWritable = isset($writeCopy[$id]);
+                $hasOob = isset($exceptCopy[$id]);
+
+                if (!$isReadable && !$isWritable && !$hasOob) {
+                    continue;
+                }
+
+                $watched = $watcher->getWatchedEvents();
```
GromNaN: `stream_socket_recvfrom()` returns `false` on non-socket PHP streams (pipes from `proc_open()`, regular files opened with `fopen()`). When `$peek === false`, `'' === false` is `false`, so the `elseif` fires and `Event::Read` is reported instead of `Event::HangUp`.

Concretely: closing the write end of a `proc_open()` stdout pipe causes the read end to become readable with EOF. The polyfill reports `Event::Read` indefinitely; the caller reads zero bytes each time and spins in a busy loop, never detecting the hangup.

A possible fix: also check `feof($stream)` when `$peek === false` to detect EOF on non-socket streams.

src/Io/Poll/Resources/stubs/Io/Poll/Context.php
```diff
@@ -0,0 +1,232 @@
...
+            );
+            $watcher = $create($this, $handle, $events, $data);
+            $this->watchers[$watcher] = null;
+
+            return $watcher;
+        }
+
+        public function wait(?int $timeoutSeconds = null, int $timeoutMicroseconds = 0, ?int $maxEvents = null): array
+        {
+            if (null !== $timeoutSeconds) {
```
GromNaN, line 103: The `$timeoutMicroseconds < 0` guard is nested inside `if (null !== $timeoutSeconds)`, so calling `wait(null, -1)` silently bypasses validation. Native PHP raises `ValueError` for negative microseconds regardless of whether `$timeoutSeconds` is `null`.

The test `testWaitNegativeTimeoutMicrosecondsThrows` covers `wait(0, -1)` but not `wait(null, -1)`. The validation (and test) should be unconditional.

src/Php86/Resources/stubs/Io/Poll/Context.php
```diff
@@ -0,0 +1,232 @@
...
+                $byId[$id] = $watcher;
+
+                $read[$id] = $stream;
+                $except[$id] = $stream;
+
+                foreach ($watcher->getWatchedEvents() as $event) {
+                    if (Event::Write === $event) {
+                        $write[$id] = $stream;
+                        break;
+                    }
```
GromNaN: When no handles are registered and `$timeoutSeconds === null` (infinite wait), the polyfill returns `[]` immediately without blocking. Native PHP's `wait(null)` blocks indefinitely regardless of how many handles are registered.

This is probably an unavoidable limitation of a pure-PHP polyfill, but it should be documented — a note in the README and/or a `// Note: ...` comment here so callers are not surprised.

GromNaN (changes requested):
This polyfill is impressively thorough. Five conformity issues with the official stub or the RFC that should be addressed before merge.

src/Io/Poll/Resources/stubs/Io/Poll/Context.php
nicolas-grekas: Verified against the merged commit (php/php-src@6c6fb56): the `@not-serializable` flag surfaces through `ext/standard/var.c`, which calls `zend_throw_exception_ex(NULL, 0, "Serialization of '%s' is not allowed", ...)`. A `NULL` class entry resolves to `zend_ce_exception` in `zend_throw_exception_zstr()` — i.e. a plain `\Exception`, not `\Error` (`unserialize()` is the same via `var_unserializer.c`). So `catch (\Error $e)` would not catch it on native either; the polyfill's `\Exception` matches native, and switching to `\Error` would actually introduce the divergence. Keeping `\Exception` in all three classes.

src/Io/Poll/Resources/stubs/Io/Poll/Watcher.php
nicolas-grekas: Same as the `Context` thread: native throws `\Exception` here too. The `@not-serializable` path in `var.c` / `var_unserializer.c` uses `zend_throw_exception_ex(NULL, ...)`, and a `NULL` ce resolves to `zend_ce_exception`. So the polyfill's `\Exception` already matches. Keeping it.

src/Php86/Resources/stubs/StreamPollHandle.php
nicolas-grekas: Same as the `Context` thread — native throws `\Exception` (not `\Error`) for `@not-serializable`, so this matches. Keeping `\Exception`.

src/Php86/Resources/stubs/Io/Poll/Watcher.php
nicolas-grekas: Good catch — fixed. `modify()` and `modifyEvents()` now validate through a shared helper (rejecting an empty array and any non-`Event` element with `\TypeError("...Argument #1 ($events) must be array of Event enums")`), placed before the active check to match native ordering (`io_poll.c` validates the event array before `php_io_poll_watcher_modify_events()` performs the inactive check). Added data-provider tests for the empty and non-`Event` cases on both methods.

src/Php86/Resources/stubs/Io/Poll/Context.php
nicolas-grekas: Fixed: the read branch now also treats `false === $peek && feof($stream)` as `HangUp`, so a non-socket stream at EOF (a `proc_open()` pipe or a regular file) reports `HangUp` instead of spinning on `Read`. Added deterministic `tmpfile()` tests for both the EOF→HangUp and has-data→Read cases.

src/Io/Poll/Resources/stubs/Io/Poll/Context.php
nicolas-grekas: The nesting actually matches native here. In the merged `wait()` (and current master) the `timeout_microseconds < 0` check sits inside `if (timeout_seconds >= 0)`; when `timeout_seconds` is null the function takes neither validation branch and treats it as an infinite wait with microseconds ignored, so `wait(null, -1)` does not raise `ValueError`. I kept the guard nested for that reason.

That said, your comment did surface a real bug: the polyfill was forwarding `$timeoutMicroseconds` to `stream_select()` even when `$timeoutSeconds` was null, which PHP 8 rejects (`microseconds must be null when seconds is null`) — so `wait(null, 500)` would have thrown. Fixed (pass null for the microseconds argument when seconds is null), with a regression test for `wait(null, -1)`.

src/Php86/Resources/stubs/Io/Poll/Context.php
nicolas-grekas: This one is backend-dependent in native, and the polyfill matches the backend it emulates. The `poll`, `wsapoll` and `eventport` backends explicitly guard the empty set and return immediately on a null timeout — e.g. `poll_backend_poll.c`:

```c
int fd_count = php_poll_fd_table_count(backend_data->fd_table);
if (fd_count == 0) {
    if (timeout != NULL && (timeout->tv_sec > 0 || timeout->tv_nsec > 0)) {
        nanosleep(timeout, NULL);
    }
    return 0;
}
```

Only `epoll` and `kqueue` lack that guard and block indefinitely on an empty set. Since the polyfill is the `Poll` backend (`getBackend() === Backend::Poll`), its immediate return matches native-poll exactly, so there's no divergence for the emulated backend — I've dropped the README/comment note I'd briefly added.

The epoll/kqueue side looks like an upstream inconsistency (those two arguably want the same zero-fd guard the poll family has); we may raise it with the RFC author separately.

nicolas-grekas: ping @bukka about this ☝️

src/Php86/Resources/stubs/Io/Poll/Context.php
```diff
@@ -0,0 +1,232 @@
...
+            $writeCopy = $write;
+            $exceptCopy = $except;
+
+            $result = @stream_select($readCopy, $writeCopy, $exceptCopy, $timeoutSeconds, null === $timeoutSeconds ? null : $timeoutMicroseconds);
+
+            if (false === $result) {
+                $error = error_get_last();
+                if ($error && false !== stripos($error['message'], 'interrupt')) {
+                    return [];
+                }
```
alexandre-daubois: On EINTR signal, native `wait()` throws `FailedPollWaitException`: `poll()` returns -1, poll_backend_poll.c passes it up with no retry, and io_poll.c throws on num_events < 0.

Here the polyfill swallows the interrupt and returns [] instead. So code that catches `FailedPollWaitException` to handle signals behaves differently on PHP < 8.6. Could be worth throwing to match the native implem or documenting the difference?

tests/Php86/PollTest.php
```diff
@@ -0,0 +1,635 @@
...
+        $this->assertContains('Epoll', $names);
+        $this->assertContains('Kqueue', $names);
+        $this->assertContains('EventPorts', $names);
+        $this->assertContains('WSAPoll', $names);
+    }
+
+    public function testBackendAutoAndPollAreAvailable()
+    {
+        $this->assertTrue((Backend::Auto)->isAvailable());
+        $this->assertTrue((Backend::Poll)->isAvailable());
```
GromNaN, line 59: It is not necessary to enclose the enum cases in parentheses.
```suggestion
        $this->assertTrue(Backend::Auto->isAvailable());
        $this->assertTrue(Backend::Poll->isAvailable());
```

src/Php86/Resources/stubs/Io/Poll/Context.php
nicolas-grekas: Good catch, and confirmed against php-src: on EINTR `poll()` returns -1, `poll_backend_poll.c` passes it up with no retry, and `io_poll.c` throws `FailedPollWaitException` on `num_events < 0`. The polyfill was swallowing the interrupt and returning `[]` instead. Fixed: `wait()` now throws `FailedPollWaitException` when `stream_select()` is interrupted, matching native. Added a regression test that triggers a real EINTR via `pcntl_alarm()` with a no-`SA_RESTART` `SIGALRM` handler.

tests/Php86/PollTest.php
nicolas-grekas: Fixed, dropped the parentheses here and in the other enum-case assertions in this file.

nicolas-grekas: reverting: the code needs to be loadable under php 7

nicolas-grekas: Had to revert this one: PHPUnit loads every test file regardless of `@requires`, and `Backend::Auto->isAvailable()` (class-constant dereference without parentheses) is a parse error before PHP 8.0, so the file stopped parsing on the 7.2-7.4 jobs. The parentheses keep it parseable down to 7.2, so I restored them.

GromNaN (approved):

nicolas-grekas:
Status update on the changes pushed since the first review round:

**Review fixes** (amended into the first commit)
- `wait()` now throws `FailedPollWaitException` when interrupted by a signal instead of returning `[]`, matching `io_poll.c` (no EINTR retry in any backend). Covered by a test that triggers a real EINTR via `pcntl_alarm()`.
- The parentheses around enum cases in the tests had to stay: `Backend::Auto->isAvailable()` is a parse error before PHP 8.0 and PHPUnit parses every test file regardless of `@requires`, so dropping them broke the 7.2-7.4 jobs.

**Native-parity rework** (`c32e070`)
A close review against php-src (`ext/standard/io_poll.c`, `main/poll/poll_backend_poll.c`) surfaced ten behavioral divergences, all fixed:
- `StreamPollHandle::__construct()` rejects closed resources with the native `TypeError`; `isValid()` also checks EOF like `php_stream_poll_handle_is_valid()`.
- `Watcher::modify()` no longer erases the user data when the argument is omitted; watched events are normalized to declaration order like the native bitmask; `Event::EdgeTriggered` is rejected with `ERROR_NOSUPPORT`.
- After a `OneShot` watcher fires, only the fd entry is dropped: the watcher stays active, `remove()` succeeds and `modify*()` throws `FailedWatcherModificationException` with `ERROR_NOTFOUND`.
- Destroying the `Context` deactivates its watchers; the watcher references the context through a `WeakReference` so this happens as eagerly as with native refcounting.
- `wait()` drops watchers on closed streams like `POLLNVAL` and returns immediately instead of sleeping the whole timeout; wake-ups that map to no watched event keep waiting instead of returning `[]` early; the select except set is unused since it signals OOB data, which native never reports, so `Event::Error` is never emitted; `HangUp` is reported alongside `Read` like `POLLIN|POLLHUP`, never for regular files or datagram sockets, and even when not watched; `ReadHangUp` is never emitted since the native poll backend never requests `POLLRDHUP`.
- Exception messages and codes match `io_poll.c`, including `Poll wait failed`, `Cannot modify/remove inactive watcher`, `Invalid handle for polling` and `Failed to add handle`.

Remaining divergences, documented in the component README: a TCP half-close reports `Read|HangUp` where native reports plain `Read` (the distinction is kernel-only), and hang-ups on write-only watchers are not detectable (peeking without readability would block).

**Package split** (`c3756da`)
The polyfill moved from `Php86` to a dedicated `symfony/polyfill-io-poll` package (`src/Io/Poll/`) requiring PHP >= 8.1, since `polyfill-php86` supports PHP >= 7.2. The split repository and Packagist entry will need to be created before the next release.

Tests went from 35 to 64, CI is green on the whole 7.2-8.6 matrix.

nicolas-grekas:
Now with phpt borrowed from php-src

Note the proposal: split this into its own `symfony/polyfill-io-poll` package.

GromNaN (approved):
Given the size of the package and the specific constraints, splitting it into a separate package makes sense.
Thank you for the work you put into creating this polyfill; I can't wait to start using it.

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

Issue #613, opened 2026-05-25
Testing environment is not comprehensive


GPHemsley:
I've been taking a look at this repo over the last week or so, and I've noticed some things:
- The testing environment is not set up to test in environments where extensions are not enabled, thereby missing the testing of polyfills for those extensions.
- Testing the iconv and mbstring extension polyfills is difficult:
  - Composer cannot be installed if neither iconv nor mbstring is enabled.
  - PHPUnit actively prevents running tests if mbstring is not enabled.
  - The mbstring polyfill depends on iconv being enabled.
- The TestListener actively skips running tests in many environments and configurations that it shouldn't.

All of this amounts to many of the polyfills not being thoroughly tested in the environments they are most needed, subtle bugs going undetected even when there are tests for them, and the test suite not being able run at all in some cases.

The vast set of changes pushed today right before the release of 1.38.0 fixed some of these festering issues, but a handful do remain, and it's clear that PHP 7 support in particular needs some attention.

stof:
Our testsuite is using a TestListener to run tests for polyfills even when the extensions *are* installed (the testsuite runs twice, once using polyfill implementations, once using native implementations).

The set of changes pushed yesterday was actually about fixing things that our testsuite *was* catching, but we did not handle it earlier (the CI was red since months).

nicolas-grekas:
Let me close as this is not actionable as is. PR are always welcome of course, thanks for your PRs today.

GPHemsley:
> Our testsuite is using a TestListener to run tests for polyfills even when the extensions _are_ installed (the testsuite runs twice, once using polyfill implementations, once using native implementations).

I haven't figured out why yet, but if you disable the TestListener, many more tests are run and additional test failures fall out. That's where #615, #616, and #617 came from, and #619 originally wound up hitting another one that I haven't put a PR together for yet (see https://github.com/symfony/polyfill/actions/runs/26445271213/job/77849851523).

> The set of changes pushed yesterday was actually about fixing things that our testsuite _was_ catching, but we did not handle it earlier (the CI was red since months).

Agreed, that's actually what prompted me to look into this in the first place.

> Let me close as this is not actionable as is. PR are always welcome of course, thanks for your PRs today.

I think there's quite a bit that's still actionable here:
- Figuring out what TestListener is doing that is causing tests not to run.
- Figuring out and implementing a different testing strategy for iconv and mbstring that can run even when those extensions are enabled.
- Measuring and expanding code coverage overall.

PR #614, opened 2026-05-25
Improve testing environment

GPHemsley:
- Adds configuration to the testing matrix to test on both a minimal and maximal build of PHP.
- Fixes tests so that everything stays green.

Speaks to #613, but does _not_ address the issues with testing iconv/mbstring polyfills nor the issue with TestListener, which is still hiding failures even in this setup.


nicolas-grekas:
Thank you @GPHemsley.

PR #615, opened 2026-05-26
Fix newline handling in Php74 implementation of mb_str_split()

GPHemsley:
The `Php74` version of `mb_str_split()` is missing the `/s` modifier that the `Mbstring` version uses, resulting in a failure in `MbstringTest::testStrSplit()` in PHP < 7.4.


nicolas-grekas:
Thank you @GPHemsley.

PR #616, opened 2026-05-26
Fix grapheme_str_split() on PCRE 8

GPHemsley:
#598 bumped the PCRE2 version requirement from 10.39 to 10.44, in order to take advantage of the fix for PCRE2Project/pcre2#410, but left the PCRE1 version untouched at 8.32.

But the necessary fix was made 3 years after PCRE1 was EOL'd, so there is no 8.x version in which `\X` works the way grapheme_str_split() needs it to.

PHP 7.2.0 was bundled with PCRE1 8.41 before PHP 7.3.0 changed to PCRE2 10.32:
https://www.php.net/manual/en/pcre.installation.php

Falling back to the now-updated regex seems to work fine, though.


nicolas-grekas:
Thank you @GPHemsley.

PR #617, opened 2026-05-26
Polyfill more cURL constants

GPHemsley:
This PR fixes how the Php84 polyfill detects libcurl support for `CURL_HTTP_VERSION_3` and `CURL_HTTP_VERSION_3ONLY`, and adds polyfills for the related `CURLOPT_ISSUERCERT_BLOB` (Php81) and `CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256` (Php82) constants when libcurl supports them.

### Php84

- Fix a latent bug: `curl_version()['version']` is a string and was being compared to the integer `0x074200`; replaced with `['version_number']`.
- Guard the whole block behind `extension_loaded('curl')` so `curl_version()` is never called when the extension isn't present.
- Keep the `\PHP_VERSION_ID < 80200` short-circuit so opcache can fold the branch away on PHP 8.2+ (per review feedback in #617).
- For `CURL_HTTP_VERSION_3ONLY`, check libcurl `>= 7.88.0` explicitly instead of using `CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256` as a (loose, libcurl 7.80.0) proxy.
- Split the test into one per constant and skip each with the appropriate libcurl version requirement.

### Php81 / Php82

- Polyfill `CURLOPT_ISSUERCERT_BLOB` (added in PHP 8.1, libcurl 7.71.0) when libcurl supports it.
- Polyfill `CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256` (added in PHP 8.2, libcurl 7.80.0) when libcurl supports it.
- Add tests verifying both constants resolve to their canonical integer values.

Note: the polyfill packages are independent (Php84 does not require Php82), so the Php84 bootstrap still does its own explicit `curl_version()` checks rather than relying on the new constants being defined.

### DIFF
--- a/src/Php81/bootstrap.php
+++ b/src/Php81/bootstrap.php
@@ -19,6 +19,10 @@ if (defined('MYSQLI_REFRESH_SLAVE') && !defined('MYSQLI_REFRESH_REPLICA')) {
     define('MYSQLI_REFRESH_REPLICA', 64);
 }
 
+if (\extension_loaded('curl') && !defined('CURLOPT_ISSUERCERT_BLOB') && curl_version()['version_number'] >= 0x074700) {
+    define('CURLOPT_ISSUERCERT_BLOB', 40295);
+}
+
 if (!function_exists('array_is_list')) {
     function array_is_list(array $array): bool { return p\Php81::array_is_list($array); }
 }
--- a/src/Php82/bootstrap.php
+++ b/src/Php82/bootstrap.php
@@ -15,6 +15,10 @@ if (\PHP_VERSION_ID >= 80200) {
     return;
 }
 
+if (\extension_loaded('curl') && !defined('CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256') && curl_version()['version_number'] >= 0x075000) {
+    define('CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256', 10311);
+}
+
 if (extension_loaded('odbc')) {
     if (!function_exists('odbc_connection_string_is_quoted')) {
         function odbc_connection_string_is_quoted(string $str): bool { return p\Php82::odbc_connection_string_is_quoted($str); }
--- a/src/Php84/bootstrap.php
+++ b/src/Php84/bootstrap.php
@@ -15,13 +15,17 @@ if (\PHP_VERSION_ID >= 80400) {
     return;
 }
 
-if (defined('CURL_VERSION_HTTP3') || \PHP_VERSION_ID < 80200 && function_exists('curl_version') && curl_version()['version'] >= 0x074200) { // libcurl >= 7.66.0
-    if (!defined('CURL_HTTP_VERSION_3')) {
-        define('CURL_HTTP_VERSION_3', 30);
-    }
+if (\extension_loaded('curl')) {
+    // CURL_VERSION_HTTP3 is defined by PHP 8.2+ when libcurl >= 7.66.0
+    if (defined('CURL_VERSION_HTTP3') || \PHP_VERSION_ID < 80200 && curl_version()['version_number'] >= 0x074200) {
+        if (!defined('CURL_HTTP_VERSION_3')) {
+            define('CURL_HTTP_VERSION_3', 30);
+        }
 
-    if (!defined('CURL_HTTP_VERSION_3ONLY') && defined('CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256')) { // libcurl >= 7.80.0 (7.88 would be better but is slow to check)
-        define('CURL_HTTP_VERSION_3ONLY', 31);
+        // CURL_HTTP_VERSION_3ONLY requires libcurl >= 7.88.0 and is not gated by any PHP-defined constant before 8.4
+        if (!defined('CURL_HTTP_VERSION_3ONLY') && curl_version()['version_number'] >= 0x075800) {
+            define('CURL_HTTP_VERSION_3ONLY', 31);
+        }
     }
 }
 
--- a/tests/Php81/Php81Test.php
+++ b/tests/Php81/Php81Test.php
@@ -42,4 +42,17 @@ class Php81Test extends TestCase
         $this->assertTrue(\defined('MYSQLI_REFRESH_REPLICA'));
         $this->assertSame(MYSQLI_REFRESH_SLAVE, MYSQLI_REFRESH_REPLICA);
     }
+
+    /**
+     * @requires extension curl
+     */
+    public function testCurlOptIssuerCertBlobDefined()
+    {
+        if (curl_version()['version_number'] < 0x074700) {
+            $this->markTestSkipped('Requires libcurl >= 7.71.0');
+        }
+
+        $this->assertTrue(\defined('CURLOPT_ISSUERCERT_BLOB'));
+        $this->assertSame(40295, \CURLOPT_ISSUERCERT_BLOB);
+    }
 }
--- a/tests/Php82/Php82Test.php
+++ b/tests/Php82/Php82Test.php
@@ -49,6 +49,19 @@ class Php82Test extends TestCase
     }
 
     /**
+     * @requires extension curl
+     */
+    public function testCurlOptSshHostPublicKeySha256Defined()
+    {
+        if (curl_version()['version_number'] < 0x075000) {
+            $this->markTestSkipped('Requires libcurl >= 7.80.0');
+        }
+
+        $this->assertTrue(\defined('CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256'));
+        $this->assertSame(10311, \CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256);
+    }
+
+    /**
      * Test cases ported from upstream.
      *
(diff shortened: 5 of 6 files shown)


nicolas-grekas:
Thanks for this. The PR is draft because you want to tackle the todo, isn't it?

nicolas-grekas:
AFAIK, the PHP_VERSION_ID checks can be optimized by opcache, while define() calls can, but only when the const actually exists. When it doesn't exist, defined checks need to remain so that a dynamically defined const can still be tested.
This could be a reason to keep these checks, even if they look duplicates in your analysis.

nicolas-grekas:
Thank you @GPHemsley.

GPHemsley:
So wait, you added polyfills for the additional constants, but then removed them from the checks?

nicolas-grekas:
yes, PHP_VERSION_ID  does the job

nicolas-grekas:
Do you see any issue with the merged code?

GPHemsley:
Well, my main concern was that the choice of those constants was based on the versions they were introduced, rather than their inherent functionality; there are many more constants that can be polyfilled if we are motivated by functionality.

But also, I was adding those defined checks in order to defer the explicit version check. If that's not a performance concern (or the tradeoff isn't worth it), then that's fine by me. But without them, isn't the PHP version check redundant?

| **PHP >= 8.4.0** | curl < 7.66.0 | curl >= 7.66.0 |
|--------|--------|--------|
| `CURL_VERSION_HTTP3` defined | _shouldn't happen_ | `CURL_HTTP_VERSION_3` defined |
| `CURL_VERSION_HTTP3` undefined | `CURL_HTTP_VERSION_3` undefined | _shouldn't happen_ |

| **PHP >= 8.2.0, < 8.4.0** | curl < 7.66.0 | curl >= 7.66.0 |
|--------|--------|--------|
| `CURL_VERSION_HTTP3` defined | _shouldn't happen_ | `CURL_HTTP_VERSION_3` needs polyfill |
| `CURL_VERSION_HTTP3` undefined | no polyfill needed | _shouldn't happen_ |

| **PHP < 8.2.0** | curl < 7.66.0 | curl >= 7.66.0 |
|--------|--------|--------|
| `CURL_VERSION_HTTP3` defined | _shouldn't happen (polyfill mistake)_ | `CURL_VERSION_HTTP3` already polyfilled, `CURL_HTTP_VERSION_3` needs polyfill |
| `CURL_VERSION_HTTP3` undefined | no polyfull needed | both `CURL_VERSION_HTTP3` and `CURL_HTTP_VERSION_3` need polyfill |

So if there's any new polyfill we're adding with this, it should be `CURL_VERSION_HTTP3` in Php82.

nicolas-grekas:
If that can help reason about this: we're not looking for completeness of the polyfills at all costs. Usefulness should be the decider.

GPHemsley:
> If that can help reason about this: we're not looking for completeness of the polyfills at all costs. Usefulness should be the decider.

And that's just it: `CURLOPT_ISSUERCERT_BLOB` and `CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256` were only "useful" when the polyfill was using them. Thus, by contrast, `CURL_VERSION_HTTP3` is more "useful" because the polyfill _is_ using it.

nicolas-grekas:
I wanted to release 1.38.1 today so I merged quickly but I'm open to follow ups of course.

stof:
the decision regarding polyfilling constants in a `-php*` package is based on whether the functionality can be polyfilled when running on older PHP versions. If the answer is no, it is better to avoid polyfilling the constant, so that detecting the constant can be used to detect the feature it enables (that's why we rejected https://github.com/symfony/polyfill/pull/543 for instance).

Note that some of those `shouldn't happen` look suspicious to me. ext-curl can be compiled against libcurl from the system rather than using a version bundled in PHP (and it is how most distro do it), which means that you could totally have an uptodate libcurl with an older PHP version. The *only* constraint we can infer is that the curl version is higher than the *minimum supported version* of that PHP version (which is not the same than the bundled version btw)

GPHemsley:
> the decision regarding polyfilling constants in a `-php*` package is based on whether the functionality can be polyfilled when running on older PHP versions. If the answer is no, it is better to avoid polyfilling the constant, so that detecting the constant can be used to detect the feature it enables (that's why we rejected #543 for instance).

I'm not really sure what you're arguing for/against in this context. Many of the curl constants are just enum ints that get passed directly to the underlying libcurl. Without them, you could get the same behavior with magic numbers.

> Note that some of those `shouldn't happen` look suspicious to me. ext-curl can be compiled against libcurl from the system rather than using a version bundled in PHP (and it is how most distro do it), which means that you could totally have an uptodate libcurl with an older PHP version. The _only_ constraint we can infer is that the curl version is higher than the _minimum supported version_ of that PHP version (which is not the same than the bundled version btw)

Can you elaborate on which scenarios you think are suspicious? Constants can exist in the underlying libcurl library, but that doesn't mean they automatically exist in the PHP version built with it, as far as I understand. That's why the polyfills are needed in the first place (and why `CURL_VERSION_HTTP3` and `CURL_HTTP_VERSION_3` are defined two PHP versions apart, despite being added in the same libcurl version). On the flipside, the definitions in the PHP code are conditional by libcurl version.

I detailed this in my original description, but it was rewritten as part of the merge:

> These constants, `CURL_HTTP_VERSION_3` and `CURL_HTTP_VERSION_3ONLY`, were added in curl 7.66.0 and curl 7.88.0, respectively.
> 
> <details>
> <summary>How PHP adds curl constants</summary>
> 
> PHP syncs up its constants from this list:
> https://curl.se/libcurl/c/symbols-in-versions.html
> 
> Using this script:
> https://github.com/php/php-src/blob/master/ext/curl/sync-constants.php
> 
> In PHP 8.1 and prior, PHP's curl constants were defined in `interface.c`:
> https://github.com/php/php-src/blob/PHP-7.2/ext/curl/interface.c
> https://github.com/php/php-src/blob/PHP-7.3/ext/curl/interface.c
> https://github.com/php/php-src/blob/PHP-7.4/ext/curl/interface.c
> https://github.com/php/php-src/blob/PHP-8.0/ext/curl/interface.c
> https://github.com/php/php-src/blob/PHP-8.1/ext/curl/interface.c
> 
> Since PHP 8.2, they are defined in `curl_arginfo.h`:
> https://github.com/php/php-src/blob/PHP-8.2/ext/curl/curl_arginfo.h
> https://github.com/php/php-src/blob/PHP-8.3/ext/curl/curl_arginfo.h
> https://github.com/php/php-src/blob/PHP-8.4/ext/curl/curl_arginfo.h
> https://github.com/php/php-src/blob/PHP-8.5/ext/curl/curl_arginfo.h
> https://github.com/php/php-src/blob/master/ext/curl/curl_arginfo.h
> </details>
> 
> PHP first integrated these two particular constants in 8.4, although the related `CURL_VERSION_HTTP3` (also from curl 7.66.0) was added to PHP in 8.2.

Ayesh:
I'm the author of the PR that added the HTTP/3 constants to this polyfill, so I wanted to defend it first:

`CURL_HTTP_VERSION_3` and `CURL_HTTP_VERSION_3ONLY` are value constants. You pass them as the value for the `CURLOPT_HTTP_VERSION` option:

```php
curl_setopt($ch, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_3);
```

These two constants can be polyfilled because the [`CURLOPT_HTTP_VERSION` constant](https://php.watch/codex/CURLOPT_HTTP_VERSION) is always available.

Both ext-curl and libcurl expect an integer value for the `CURLOPT_HTTP_VERSION` option, so even without the polyfill, you can pass an integer value and it will work (or fail if HTTP/3 support is not built in):


```php
curl_setopt($ch, CURLOPT_HTTP_VERSION, 30); // Works
```

You can even pass a random integer value and it will also work. The call will return `false` if it is not a recognized _integer_ value. But you can still call it. 

That is why we polyfilled `CURL_HTTP_VERSION_3` and `CURL_HTTP_VERSION_3ONLY` constants. It is one of the very few cases where we can polyfill a curl _value_ constant and meaningfully enable a feature and still have that functionality working on older PHP versions. 

---

I align with @stof's comment regarding the addition of new constants.

This PR polyfills two new Curl _option_ constants. In ext-curl, `curl_setopt` (and `_array`) validates the option constants in a large internal `switch`. It rejects option constants that do not fall into any of the supported case. 

**Polyfilling Curl option constants does not make it possible to use that functionality**. In fact, it does not make sense polyfill any `CURLOPT_*` constants. The only constants worth polyfilling are `CURLE_*` constants (which ext-curl lacks coverage in comparison to libcurl), and the _value_ constants for Curl options that need to be decided case by case.

Unknown `CURLOPT_*` values are blocked by ext-curl so they never reach libcurl. This effectively makes it impossible to polyfill any of the `CURLOPT_*` constants regardless of the libcurl version PHP is built with.

GPHemsley:
I want to be clear about some things for the record:
- The current contents of the description are not my words. My original description covered completely different things, and provided what I think is important context for this discussion.
- I had started this in draft precisely because I wanted to investigate the consequences of polyfilling the constants I chose arbitrarily based on version numbers.
- I did not have a chance to document the changes I force-pushed after initially filing the draft PR.
- I was not involved in the decision to polyfill the `CURLOPT_ISSUERCERT_BLOB` or `CURLOPT_SSH_HOST_PUBLIC_KEY_SHA256` constants, nor the decision to move this out of draft and merge it.

PR #590, opened 2026-05-21
PHP can be built without PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT

alexpott:
It's possible to build PHP without `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` which results in errors on Drupal when using Drush - 
```
vendor/bin/drush status
Error: Undefined constant PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT in [constant expression]() (line 38 of vendor/symfony/polyfill-php84/Resources/stubs/Pdo/Mysql.php).
```

This occurs on cheap hosting.

PR #618, opened 2026-05-26
fix: Correctly define MySQL constants

mitelg:
hey :wave: 

the change made in https://github.com/symfony/polyfill/pull/590 was wrong and is now breaking our CI pipeline. 
The author assumed, the constants where defined under the same condition (see https://github.com/symfony/polyfill/pull/590#issuecomment-4507728932), but actually they are not. If you look closely you see that there is first used a `#ifndef` (mind the **n**) and the `ATTR_SSL_VERIFY_SERVER_CERT` constant is defined with `#ifdef` (no **n**). 

not sure if an additional condition is the correct solution, but checking for all four constants at the same time is definitely not correct. But having an additional check should also fix the original authors problem

### DIFF
--- a/src/Php84/Resources/stubs/Pdo/Mysql.php
+++ b/src/Php84/Resources/stubs/Pdo/Mysql.php
@@ -15,7 +15,7 @@ use PDO;
 
 if (\PHP_VERSION_ID < 80400 && \extension_loaded('pdo_mysql')) {
     // Feature detection for non-mysqlnd; see also https://www.php.net/manual/en/class.pdo-mysql.php#pdo-mysql.constants.attr-max-buffer-size
-    if (\defined('PDO::MYSQL_ATTR_MAX_BUFFER_SIZE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_FILE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_GROUP') && \defined('PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT')) {
+    if (\defined('PDO::MYSQL_ATTR_MAX_BUFFER_SIZE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_FILE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_GROUP')) {
         class Mysql extends \PDO
         {
             public const ATTR_COMPRESS = \PDO::MYSQL_ATTR_COMPRESS;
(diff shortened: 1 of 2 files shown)


tests/Php84/PdoTest.php
```diff
@@ -196,10 +196,12 @@ public function testMysqlConstants()
         $this->assertSame(\PHP_VERSION_ID < 80500 ? 1012 : 1011, \Pdo\Mysql::ATTR_SERVER_PUBLIC_KEY);
         $this->assertSame(\PHP_VERSION_ID < 80500 ? 1013 : 1012, \Pdo\Mysql::ATTR_MULTI_STATEMENTS);
         $this->assertSame(\PHP_VERSION_ID < 80500 ? 1015 : 1014, \Pdo\Mysql::ATTR_LOCAL_INFILE_DIRECTORY);
-        if (\defined('PDO::MYSQL_ATTR_MAX_BUFFER_SIZE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_FILE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_GROUP') && \defined('PDO::ATTR_SSL_VERIFY_SERVER_CERT')) {
+        if (\defined('PDO::MYSQL_ATTR_MAX_BUFFER_SIZE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_FILE') && \defined('PDO::MYSQL_ATTR_READ_DEFAULT_GROUP')) {
             $this->assertSame(1003, \Pdo\Mysql::ATTR_READ_DEFAULT_FILE);
             $this->assertSame(1004, \Pdo\Mysql::ATTR_READ_DEFAULT_GROUP);
             $this->assertSame(1005, \Pdo\Mysql::ATTR_MAX_BUFFER_SIZE);
+        }
+        if (\defined('PDO::ATTR_SSL_VERIFY_SERVER_CERT')) {
```
keulinho: in the original PR there was a typo here that you copy & pasted
```suggestion
        if (\defined('PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT')) {
```

src/Php84/Resources/stubs/Pdo/Mysql.php
```diff
@@ -35,6 +35,46 @@ class Mysql extends \PDO
...
+            public static function connect(string $dsn, ?string $username = null, ?string $password = null, ?array $options = null): self
+            {
+                try {
+                    return new self($dsn, $username, $password, $options);
+                } catch (\PDOException $e) {
+                    throw preg_match('/^Pdo\\\\Mysql::__construct\(\) cannot be used for connecting to the "([a-z]+)" driver/', $e->getMessage(), $matches) ? new \PDOException(\sprintf('Pdo\Mysql::connect() cannot be used for connecting to the "%s" driver', $matches[1])) : $e;
+                }
+            }
+        }
+    } elseif(\defined('PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT')) {
```
keulinho: Not sure if adding this case is enough, seems that the other three `PDO::MYSQL_ATTR` can be compiled separetly from `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` that means we likely need to maintain 4 branches here: 

* `PDO::MYSQL_ATTR_*` defined with `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT`
* `PDO::MYSQL_ATTR_*` defined without `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT`
* `PDO::MYSQL_ATTR_*` not defined, but `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` is defined
* Neither `PDO::MYSQL_ATTR_*` nor `PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` is defined

mitelg: if you look at the source code https://github.com/php/php-src/blob/PHP-8.3/ext/pdo_mysql/php_pdo_mysql_int.h#L160 your first case could not happen at all :thinking: 

all other three cases should be covered:
1: by the if
2: by the elseif
3: by the else

keulinho: Ok seems the case that both PDO::MYSQL_ATTR_* and PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT is defined cannot happen: http://github.com/php/php-src/blob/PHP-8.3/ext/pdo_mysql/php_pdo_mysql_int.h#L160

nicolas-grekas:
Thank you @mitelg.

Issue #570, opened 2026-04-13
mb_trim() breaks PHPStan since v1.34.0


eexit:
Hello,

I just upgraded from v1.33.0 to v1.34.0 and my PHPStan breaks on `mb_trim()` return value:

```php
\PHPStan\dumpType(mb_trim('foo'));
// Dumped type: mixed
```

It should be `string` and not `mixed`. This is related to https://github.com/symfony/polyfill/pull/501.

--- 

```
PHP 8.3.30 (cli) (built: Jan 30 2026 01:25:40) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.3.30, Copyright (c) Zend Technologies
    with Zend OPcache v8.3.30, Copyright (c), by Zend Technologies
    with Xdebug v3.5.1, Copyright (c) 2002-2026, by Derick Rethans
```

Thanks!

eexit:
Hello, any update on this?

Thanks!

mitelg:
hey @nicolas-grekas 

imo, this is not fixed correctly. In https://github.com/symfony/polyfill/pull/592 you added `string|false` as return type. 
But why returning `false`? I see that in the polyfill implementations in some cases `false` could be returned. But this contradicts to the original functions, which only return `string`.

nicolas-grekas:
Ah right, false can be returned but only on legacy version.

mitelg:
are you going to fix it or should I open a separate issue for this? 🙂

nicolas-grekas:
See #619 - please confirm before I merge if possible.

mitelg:
I will have a look 👍 thanks for the PR

eexit:
Thanks @nicolas-grekas & @mitelg!

PR #619, opened 2026-05-26
[Mbstring][Php83][Php84] Isolate PHP 7 mb_* declarations into bootstrap72.php

nicolas-grekas:
The new mb_* polyfills (`mb_str_pad`, `mb_ucfirst`, `mb_lcfirst`, `mb_trim`, `mb_ltrim`, `mb_rtrim`) trigger a warning and return `false` on PHP 7 in case of invalid encoding, to avoid requiring polyfill-php80 transitively for the `\ValueError` class (see #501). Their `@return string|false` annotation reflects this, which contradicts the native PHP 8.4 contract and confuses static analysis on PHP 8+.

This PR moves the PHP 7 declarations into a new `bootstrap72.php` file so the falsy return is physically scoped to PHP 7, while `bootstrap.php` declares the PHP 8+ variants with `: string` typed return matching the native contract.

For Php83 and Php84, `bootstrap80.php` is merged into `bootstrap.php` since those files only use PHP 7.2-compatible syntax. Mbstring keeps its `bootstrap80.php` because it uses PHP 8 union types like `array|string|null`.

Addresses https://github.com/symfony/polyfill/issues/570#issuecomment-4542262721


mitelg:
hey @nicolas-grekas 

I tested your changes and the PHPStan issues with the return type are gone for our supported PHP versions :+1:

Issue #512, opened 2024-12-21
mb_str_pad polyfill tests are failing for UTF-32 on PHP 8


stof:
See CI runs for those failures.

Apparently, the pad string used in `paddingEncodingProvider` for UTF-32 is now considered invalid by `iconv`. It triggers a notice `iconv_strlen(): Detected an illegal character in input string`, which then returns `false` for the length

PR #622, opened 2026-05-27
[Mbstring][Php3] Fix issues with mb_str_pad()

GPHemsley:
This PR does three things related to `mb_str_pad()`:
- It fixes another issue hidden by the TestListener (mentioned in #613 and seen during development of #619) where `ValueError` was being thrown on PHP 7 in the `Php83` version of `mb_str_pad()` instead of triggering the warning like is done in other mbstring situations, including the `Mbstring` implementation of `mb_str_pad()`.
- It rearranges the order of argument checks in both the `Php83` and `Mbstring` to match native PHP calls, and adds tests to back that up.
- It removes the skipping of UTF-32 checks in `MbstringTest`. #512 reported in 2024 that CI tests were failing on PHP 8 because of UTF-32, but I'm not seeing any evidence of that now (though #435 mentioned in 2023 that PHP 7.2 had issues with UTF-32 when mbstring was missing).

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -879,16 +879,6 @@ final class Mbstring
     /** @return string|false */
     public static function mb_str_pad(string $string, int $length, string $pad_string = ' ', int $pad_type = \STR_PAD_RIGHT, ?string $encoding = null)
     {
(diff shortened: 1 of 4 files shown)


nicolas-grekas:
Thank you @GPHemsley.

PR #607, opened 2026-05-25
[Mbstring] Make mb_convert_encoding($s, $y, 'HTML-ENTITIES') match native output

nicolas-grekas:
Fixes #344

Native `mb_convert_encoding($s, $y, 'HTML-ENTITIES')` decodes every numeric entity, including `&#39;` and the C0/C1 control characters (`&#0;`-`&#31;`, `&#127;`-`&#159;`). The polyfill was delegating to `html_entity_decode($s, ENT_COMPAT, 'UTF-8')` which leaves those untouched.

Switch to `ENT_QUOTES` so `&#39;` is decoded the same way native does, then post-process the result to decode the remaining numeric entities for control characters. All 1024 entities (0-1023) now match the native output.

PR #623, opened 2026-05-27
[Mbstring] Fix issues with mb_scrub()

GPHemsley:
- This moves the definition of `mb_scrub()` into the `Mbstring` class so that it can take advantage of other implementation details there, and can be checked by the TestListener for signature issues.
- This removes the iconv skip just added to `mb_convert_encoding()` in #607, which bypasses iconv when the input and output encodings are the same. Unlike the native PHP implementation (for UTF-8, anyway), it fails to account for invalid characters, which would be stripped by the call to iconv. This was causing `MbstringTest::testScrub()` to fail, because the invalid character was not being stripped.

### DIFF
--- a/src/Mbstring/Mbstring.php
+++ b/src/Mbstring/Mbstring.php
@@ -157,10 +157,6 @@ final class Mbstring
             $fromEncoding = 'UTF-8';
         }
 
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Thank you @GPHemsley.

Issue #624, opened 2026-05-31
Polyfill target design ambiguous


GPHemsley:
I think there is some ambiguity in the target design of the polyfills, which can lead to confusion where the possibilities interact with each other. It would be good to decide on a design principle and record that decision somewhere to guide future work.

So here are some questions that I have:
- When designing an extension polyfill, is the goal to replicate the expected functionality of the installation's PHP version, or to replicate the functionality of the latest version of PHP?
  - So if a later version of PHP drops deprecated functionality, would the polyfill implement and support that functionality for the earlier versions of PHP that supported it?
  - Or if a later version of PHP implements new functionality, would that functionality be backported to the earlier versions of PHP?
- When designing a PHP version polyfill, is the goal to replicate the PHP functionality at that version?
  - If the functionality is part of an extension, and that extension has its own polyfill, are the implementations meant to be the same?
- Does the fact that we can't influence `extension_loaded()` or conditionally deprecate constants (since `const` is only available at compile time and `define()` doesn't support attributes) change any of those decisions?
- Do the answers change if the polyfill is for an extension that was never bundled with PHP and is thus subject to its own version history?

Some concrete examples:
- `mb_decode_numericentity()` had an `is_hex` parameter that was dropped in PHP 8.0. Should the `Mbstring` polyfill support it? Should the bootstrap files accept it?
- Input validation in PHP 7 was done by triggering an `E_WARNING` and returning `false`; in PHP 8, this was changed to raising a `ValueError` exception (which was also a new exception).
  - What should existing functions in an extension polyfill do?
  - What should new functions in an extension polyfill do?
    - Does the answer differ for e.g. `mb_str_split()` (introduced in PHP 7.4) compared to `mb_str_pad()` (introduced in PHP 8.3)?
  - What should new functions in a PHP version polyfill do?
- The mbstring regex functions, which the `Mbstring` polyfill currently does not support, are going to be deprecated in PHP 8.6 because the underlying library is no longer maintained. Does that mean no effort should be made to add them to the polyfill? Should they be added just to throw a deprecation notice?
- The `Uuid` extension polyfill (which is polyfilling a PECL extension that was never bundled with PHP) has to contend with the deprecation of the `UUID_TYPE_DCE` and `UUID_TYPE_NAME` constants (which I think are pass-through deprecations from the underlying `libuuid`).
  - How are the version changes in that extension represented in the polyfill?
[...]

PR #639, opened 2026-07-01
Document the polyfill design principles

nicolas-grekas:
Refs #624.

That issue asks for the polyfill target design to be written down. This adds a "Design principles" section to the README capturing the decisions that are already applied across the codebase:

- Backport the API of the latest PHP version to older ones; do not re-add removed/legacy behavior (e.g. the `is_hex` argument dropped from `mb_decode_numericentity()` in 8.0).
- A polyfill reproduces the native contract of the function it shims, so `ValueError`-on-invalid-input functions throw on every version while `false`-returning ones keep returning `false`.
- Bootstrap signatures match the native signature of the running PHP version; new arguments that cannot be implemented are accepted and ignored (e.g. the grapheme `$locale` argument in 8.5).
- Never-bundled extensions (`uuid`, `deepclone`) follow the extension's own history, mirroring the newest native state including deprecations.
- A function provided by both an extension polyfill and a version polyfill must behave identically.

The wording can of course be adjusted to match your intended policy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,43 @@ It is strongly recommended to upgrade your PHP version and/or install the missin
 extensions whenever possible. This polyfill should be used only when there is no
 better choice or when portability is a requirement.
 
+Design principles
+=================
+
+Polyfills target the API of the latest PHP version and backport it to older ones.
+The goal is that code written for a recent PHP runs unchanged on the oldest
+supported version; it is not to reproduce the historical behavior of every
+intermediate version. In practice:
+
+- New functions, classes, constants and arguments are backported; removed or
+  legacy behavior is not. For example, the `is_hex` argument that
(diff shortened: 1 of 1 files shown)


stof:
I would also add the mention of constants being polyfilled only when they can produce a meaningful effect. When the constant is meant to enable a new feature in an existing function, it is better to skip polyfilling it as the effect cannot be polyfilled (and the absence of the constant can be used as feature detection).
See for instance https://github.com/symfony/polyfill/pull/543 which was rejected because of that.

nicolas-grekas:
Good point, added a bullet: a constant whose only role is to toggle behavior in a function we can't polyfill (like `FILTER_THROW_ON_FAILURE` for `filter_var()`, ref #543) is left undefined so its absence stays usable for feature detection. I also noted the contrast with attributes, which are polyfilled because reflection still conveys their semantics without engine support.

PR #643, opened 2026-08-05
[DeepClone] Hydrate mangled keys through the grouped path even when they resolve to the object's own scope

nikophil:
`deepclone_hydrate()` fails on mangled keys when they all resolve to the object's own class scope: 

  ```php
class Foo
{
    protected $prot;
    private $priv;
}

// 💥 Error: Cannot access property starting with "\0"
deepclone_hydrate(Foo::class, (array) new Foo());
```
(errors in the CI seems unrelated)


nicolas-grekas:
Thank you @nikophil.

PR #645, opened 2026-08-07
[IoPoll] Take Time\Duration in Context::wait()

nicolas-grekas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Depends on #642, on top of which this branch sits — only the last commit belongs here.

php/php-src#23092 is merged and ships in 8.6 Beta 1, alongside `Time\Duration`. It replaces the `$timeoutSeconds`/`$timeoutMicroseconds` arguments of `Io\Poll\Context::wait()` by a single `?Time\Duration $timeout`:

```php
-public function wait(?int $timeoutSeconds = null, int $timeoutMicroseconds = 0, ?int $maxEvents = null): array
+public function wait(?\Time\Duration $timeout = null, ?int $maxEvents = null): array
```

The vendored phpt files get exactly the edits php-src made in that commit: `wait(0)` becomes `wait(Time\Duration::fromSeconds(0))` and `wait(0, 100000)` becomes `wait(Time\Duration::fromMicroseconds(100000))`. Every file that was a byte-identical copy of `ext/standard/tests/poll/` still is; the three that already differed (`poll.inc`, `poll_stream_sock_rw_close.phpt`, `poll_stream_sock_rw_multi_level.phpt`, all stale rather than adapted) keep exactly the differences they had.

`symfony/polyfill-time` is deliberately not a dependency: the polyfill never builds a `Duration`, it only reads `->negative`, `->seconds` and `->nanoseconds` off the instance it is handed, and the type declaration is only resolved when a non-null value is passed. Calling `wait()` or `wait(null)` therefore works without it, and a caller that has a `Duration` to pass necessarily has the class already. The component README says where to get it.

One divergence worth a decision: native reports the negative-timeout error as argument #2,

```c
if (timeout->duration.negative) {
    zend_argument_value_error(2, "must not be negative");
```

which looks like a leftover from the old signature, since `$timeout` is argument #1 and #2 is now `$maxEvents`. No phpt covers it. The polyfill reports #1; if upstream keeps #2 I will align.

### DIFF
--- a/src/Io/Poll/README.md
+++ b/src/Io/Poll/README.md
@@ -16,6 +16,10 @@ never reported, and a TCP half-close reports `Read|HangUp` where native
 reports plain `Read`. The phpt tests of the native implementation, borrowed
 from php-src, run against the polyfill as part of the test suite.
 
+`Context::wait()` takes its timeout as a `Time\Duration`. That class is not
+required to call `wait()` without a timeout; install `symfony/polyfill-time`
+to build one on PHP < 8.6.
+
 More information can be found in the
 [main Polyfill README](https://github.com/symfony/polyfill/blob/main/README.md).
 
--- a/src/Io/Poll/Resources/stubs/Io/Poll/Context.php
+++ b/src/Io/Poll/Resources/stubs/Io/Poll/Context.php
@@ -98,20 +98,17 @@ if (\PHP_VERSION_ID < 80600) {
             return $this->watchers[$id] = $create(\WeakReference::create($this), $handle, $events, $data);
         }
 
-        public function wait(?int $timeoutSeconds = null, int $timeoutMicroseconds = 0, ?int $maxEvents = null): array
+        public function wait(?\Time\Duration $timeout = null, ?int $maxEvents = null): array
         {
-            if (null !== $timeoutSeconds) {
-                if ($timeoutSeconds < 0) {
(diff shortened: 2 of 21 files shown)


stof:
> which looks like a leftover from the old signature, since `$timeout` is argument #1 and #2 is now `$maxEvents`. No phpt covers it. The polyfill reports #1; if upstream keeps #2 I will align.

I suggest reporting an issue upstream about that.

stof:
> the three that already differed (`poll.inc`, `poll_stream_sock_rw_close.phpt`, `poll_stream_sock_rw_multi_level.phpt`, all stale rather than adapted) keep exactly the differences they had.

do they differ intentionally or because php-src updated them after we vendored them ? In the later case, maybe we should sync them.

TimWolla:
> I suggest reporting an issue upstream about that.

php/php-src#23099

nicolas-grekas:
php/php-src#23099 settles the argument number question: the negative timeout is reported as argument #1, which is what this polyfill already did, so there is nothing to align. Its new `poll_ctx_wait.phpt` is now vendored here byte-identical and passes as is, both messages included.

Tinywan:
When will it be released? @nicolas-grekas

Issue #46, opened 2016-03-22
"hex2bin()"-polyfill break tests from swiftmailer (php 5.3)


voku:
In the test-class (https://github.com/voku/swiftmailer/blob/5.x/tests/unit/Swift/Transport/Esmtp/Auth/NTLMAuthenticatorTest.php) from swiftmailer, there is already a simple polyfill for php 5.3. But when I use the "symfony-polyfill"-package, some of this tests will break... Here the test results via Travis-CI: https://travis-ci.org/voku/swiftmailer/jobs/117393133

stof:
The logic in the polyfill is the same than in the Swiftmailer test, except for the additional validation rules missing in Swiftmailer. Are you sure that your test failure is not an actual bug in your code detected by the tests ?

binwiederhier:
This is a duplicate of https://github.com/symfony/polyfill/issues/47

nicolas-grekas:
Close in #49

PR #646, opened 2026-08-22
[Idn] Bound the Punycode decoder input length

iliaal:
| Q             | A
| ------------- | ---
| Branch?       | 1.x
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| License       | MIT

`punycodeDecode()` inserts every decoded character into an array with `array_splice()`, making the decode **quadratic** in the payload length. The label size is only checked *after* decoding, so a crafted multi-kilobyte `xn--` label reaching `idn_to_utf8()` / `idn_to_ascii()` spends minutes of CPU in a single call before being rejected:

| payload | time (before) |
|---|---|
| 5 KB | 36 ms |
| 80 KB | 12.4 s |
| 200 KB | **177 s** |

This is reachable from any application that validates attacker-controlled domains while ext-intl is not installed — the deployment this polyfill targets. `egulias/emailvalidator`, for one, depends on this package for exactly that.

Valid ACE labels are limited to 63 bytes, so the patch rejects payloads beyond a generous 1024-byte bound up front with `ERROR_PUNYCODE`, capping worst-case work at roughly half a million elementary operations (~ms). This mirrors observable native behavior: native ICU never decodes absurdly long labels — it rejects a 5 KB domain instantly.

Labels above the DNS limit but below the bound still decode exactly as before, so ToUnicode keeps reporting their decoded form even when invalid, and all UTS #46 spec fixtures pass unchanged (37602 tests green, run both with and without ext-intl).

After the patch: same 200 KB input completes in 125 ms (the residual is the linear mapping pass), and legitimate domains (`xn--bcher-kva.com`, `bücher.example`) behave identically.

### DIFF
--- a/src/Intl/Idn/Idn.php
+++ b/src/Intl/Idn/Idn.php
@@ -62,6 +62,13 @@ final class Idn
     public const MAX_INT = 2147483647;
 
     /**
+     * Punycode decoding does work quadratic in the payload length. Valid ACE
+     * labels are limited to 63 bytes, so payloads beyond this (generous) bound
+     * are rejected without being decoded to keep the work bounded.
+     */
+    private const MAX_DECODE_PAYLOAD_SIZE = 1024;
+
+    /**
      * Contains the numeric value of a basic code point (for use in representing integers) in the
      * range 0 to BASE-1, or -1 if b is does not represent a value.
      *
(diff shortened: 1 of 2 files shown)


iliaal:
The single failing job (PHP 8.6 + APC) is pre-existing CI breakage on `1.x` itself: APC 3.1.9 no longer compiles against PHP 8.6 (`apc.h:69: TSRMLS_DC`). Recent runs on the base branch fail the same way (e.g. runs 31153435446, 30343456194). All Idn tests pass in every matrix cell that gets that far, including the no-intl cells this change targets.

nicolas-grekas:
Opened #648 as a companion for the encoder, which has the mirror problem: `punycodeEncode()` rescans the code point array once per distinct code point, so a label of distinct code points costs 25 s at 64 KB. Same root cause as here, in that the polyfill converts input ext-intl refuses outright, and the two changes do not overlap. This one keeps the decoder bound as you wrote it.

nicolas-grekas:
Thank you @iliaal.

PR #654, opened 2026-09-20
[Uuid] Fix duplicate v1 UUIDs generated within the same microsecond

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 1.x
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

`UuidTest::testCreateNoOverlap` has been red on 1.x since #650, and it is not luck. `microtime()` gives microseconds where the timestamp field of a v1 UUID holds hundreds of nanoseconds, so two UUIDs generated in the same microsecond differ only by the 14 random bits of their clock sequence. On a runner fast enough to emit two or three of them per microsecond, that is a birthday problem, which is where the one or three duplicates per 100000 come from.

Frozen clock, to see it without the statistics: 10000 UUIDs generated in a single microsecond give 7507 unique ones today.

The clock sequence is now drawn at random when the clock moved, and counted up from there when it did not, which is unique up to 16384 UUIDs per microsecond. The new test covers that through a `microtime()` shadowed in the polyfill namespace, so it fails or passes for the reason above rather than by chance.


tests/Uuid/frozen_clock.php
```diff
@@ -0,0 +1,20 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Polyfill\Uuid;
+
+/**
+ * Lets UuidTest stop the clock, so that generating within one microsecond can be tested.
+ */
+function microtime($asFloat = null)
```
stof, line 17: Why not using the ClockMock feature of the phpunit bridge instead of reimplementing the same manually ?

PR #656, opened 2026-09-20
[IoPoll] Make wait() independent of the number of watchers

nicolas-grekas:
| Q             | A
| ------------- | ---
| Branch?       | 1.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

`wait()` walked every watcher three times per call: once to drop closed streams, once to build the `select()` sets, once to map readiness back to watchers. The sets are now kept in sync with the fd table, the result loop only visits what `select()` reported, and closed streams are dropped when `select()` rejects them instead of being looked for on every call.

Overhead per `wait()` over a raw `stream_select()` doing the same job:

| watchers | before | after |
| --- | --- | --- |
| 1 | +4.03 us | +2.67 us |
| 50 | +21.1 us | +2.92 us |
| 200 | +68.9 us | +5.72 us |
| 400 | +153.2 us | +7.35 us |

Measured from the Revolt event loop driver I am writing on this API: 20k socket round trips with N idle watched sockets alongside, on 8.5, went from 4.7k/s to 13.8k/s at 400 idle sockets, where `stream_select()` itself does 15.1k/s. It was 3x slower than `stream_select()`, it is now within 10%.

Two smaller ones in the same pass: the per-call `set_error_handler()` pair is replaced by `error_clear_last()`, and the stream traits that hang-up detection needs are read once in `add()` instead of on every wake-up.


src/Io/Poll/Resources/stubs/Io/Poll/Context.php
```diff
@@ -23,6 +27,33 @@ final class Context
...
+
+        /** @var array<int, true> */
+        private array $wantsRead = [];
+
+        /**
+         * Traits of each watched stream that hang-up detection needs, as a bitmask of
+         * self::IS_SOCKET, self::IS_SEEKABLE and self::IS_DATAGRAM. They never change
+         * for a given stream, so they are read once instead of on every wake-up.
+         *
+         * @var array<int, int>
```
stof, line 53:
```suggestion
         * @var array<int, int-mask-of<self::IS_*>>
```

